PR #359, opened 2014-12-28
Add integration test for configuration loading.

troessner:
After we had merged #356 I realized that while the new way of configuration loading was tested with unit tests, there certainly was an integration spec missing. This PR introduces a proper integration which I believe covers most if not all use cases.

### DIFF
--- a/features/configuration_loading.feature
+++ b/features/configuration_loading.feature
@@ -1,0 +1,62 @@
+Feature: Offer different ways how to load configuration
+
+  There are 3 ways of passing reek a configuration file:
+  - Using the cli "-c" switch
+  - Having a file ending with .reek either in your current working directory or in a parent directory
+  - Having a file ending with .reek in your HOME directory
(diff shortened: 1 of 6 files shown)


troessner:
Soooo, what's our status here?

While I'm also not the biggest fan of creating files outside of the project directory this is a test that actually really tests the functionality. I mean, everything else just.....doesn't. So we might have super-fancy tests that still work while the actual functionality doesn't.

And in any case, having an integration spec for this functionality is better than no integration spec at all.

I'd suggest that we merge this little rascal and just see how it plays out. 
If we constantly run into trouble when running tests locally or on travis we can revise this approach.
But for now I'd say let's go with it.

What do you guys think?

mvz:
I agree. I would like a different approach in the end, but it will be a fair bit of work I can't find the the time at the moment. To be revisited later, I guess.

troessner:
Can you please merge then @mvz ?
I really don't like merging my own PR's.;)

mvz:
Done :-).

PR #362, opened 2015-01-09
Revise rake task.

### LINKED ISSUE #326: Re-visit reek task
As mentioned in #325, there are some odd things about the reek task provided:

* Why is the library path altered? The user is supposed to have reek installed.
* Why is ruby called, and not directly reek? This only makes sense if you're working on reek itself and don't have it installed.
* Why provide a way to add options to ruby?
* The task changes the normal default config file behavior.

I'd like to revisit this task and simplify it.

troessner:
This fixes #326 - I basically rewrote the whole task more or less.
See comments in the diff for further explanation.

### DIFF
--- a/features/rake_task/rake_task.feature
+++ b/features/rake_task/rake_task.feature
@@ -82,15 +82,3 @@ Feature: Reek can be driven through its Task
     And it reports:
       """
       """
-  Scenario: can be configured with config_files
-    When I run rake reek with:
-      """
-      Reek::Rake::Task.new do |t|
-        t.config_files = 'spec/samples/minimal_smelly_and_masked/config.reek'
-        t.source_files = 'spec/samples/minimal_smelly_and_masked/minimal_dirty.rb'
-      end
-      """
-    Then it succeeds
-    And it reports:
-      """
-      """
--- a/lib/reek/rake/task.rb
+++ b/lib/reek/rake/task.rb
@@ -3,6 +3,7 @@
 require 'rake'
 require 'rake/tasklib'
 require 'pathname'
+require 'English'
 
 module Reek
   #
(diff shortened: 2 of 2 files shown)


features/rake_task/rake_task.feature
```diff
@@ -93,4 +94,7 @@ Feature: Reek can be driven through its Task
     Then it succeeds
     And it reports:
       """
+      [DEPRECATED]
+      Starting with reek 1.6 there is only one configuration file possible.
+      Please use "config_file" instead of "config_files".
```
troessner: At one point in time we should get rid of "config_files" so let's start by deprecating it and remove it in a year or so.

lib/reek/rake/task.rb
```diff
@@ -3,6 +3,7 @@
 require 'rake'
 require 'rake/tasklib'
 require 'pathname'
+require 'English'
```
troessner: I'm afraid I'm just not 1337 enough for remembering the difference between $? and $!.;)

lib/reek/rake/task.rb
```diff
@@ -32,118 +33,87 @@ module Rake
     #   rake reek REEK_OPTS=-s                   # sorts the report by smell
     #
     class Task < ::Rake::TaskLib
-      # Name of reek task.
-      # Defaults to :reek.
-      attr_accessor :name
+      # Name of reek task. Defaults to :reek.
+      attr_writer :name
```
troessner: I basically changed all attr_accessors to attr_writers because that is all we need.

lib/reek/rake/task.rb
```diff
@@ -32,118 +33,87 @@ module Rake
...
 
-      # Array of directories to be added to $LOAD_PATH before running reek.
-      # Defaults to ['<the absolute path to reek's lib directory>']
-      attr_accessor :libs
```
troessner: Was obsolete.

lib/reek/rake/task.rb
```diff
@@ -32,118 +33,87 @@ module Rake
...
-      attr_accessor :source_files
+      attr_writer :source_files
 
       # String containing commandline options to be passed to Reek.
       # Setting the REEK_OPTS environment variable overrides this value.
       # Defaults to ''.
-      attr_accessor :reek_opts
-
-      # Array of commandline options to pass to ruby. Defaults to [].
-      attr_accessor :ruby_opts
```
troessner: And that was just...weird..;)

lib/reek/rake/task.rb
```diff
@@ -32,118 +33,87 @@ module Rake
...
-      attr_accessor :config_file
-
-      def config_files=(value)
-        @config_file = value
-      end
-
-      def config_files
-        @config_file
-      end
+      attr_writer :config_file
```
troessner: BTW we could easily get rid of this as well since there's absolutely nothing keeping you from putting this into REEK_OPTS. I left it in to not f**k up everybodies rake tasks.

features/rake_task/rake_task.feature
chastell: I’d say let’s drop it in 2.0 already, that’s exactly what major version bumps are for.

troessner: Good idea!

troessner:
@chastell have you reviewed this PR? If yes, I'd appreciate either constructive criticism or you hitting the "merge pull request" button.:)

lib/reek/rake/task.rb
mvz: RuboCop agrees with you :-).

mvz:
Would you like to release this on a final 1.x version? If this is straight for 2.0, we can remove the `config_files` option altogether.

mvz:
Oh, I see this also includes a breaking change in the default config file location, so I guess 2.0 it is. I think the deprecation warning is probable good for now; perhaps we can add something like 'this option will be removed in 2.1' (or some other version).

troessner:
>>  I think the deprecation warning is probable good for now; perhaps we can add something like 'this option will be removed in 2.1' (or some other version).

You had an excellent point in 

>> If this is straight for 2.0, we can remove the config_files option altogether.

Then let's just remove it.:)

troessner:
@mvz removed the deprecation warning and the option itself.
So I guess this bad boy is mergeable then?

troessner:
I'm afraid reeks rake task in 1.6.* is still not working properly. 

If you don't specifiy a config file at all then [this](https://github.com/troessner/reek/blob/master/lib/reek/rake/task.rb#L89)
will set it to nil. 

Which will then [here] (https://github.com/troessner/reek/blob/master/lib/reek/rake/task.rb#L118)
lead to an "-c" flag that is passed with an empty string, resulting in the first "source_file" to be specified as config file.

So if you have the source files lib/foo.rb and lib/bar.rb then this will result in this reek call:

>> reek -c lib/foo.rb lib/bar.rb

which will rightfully fail.:)

Now before 1.6.0 this was set to a glob which still didn't work if there was no file, but it kind of did the right thing still.:)

This pull request fixes this so we should get this merged and released as 1.6.2 a.s.a.p. - I also re-introduced "config_files" (see the diff) so it fully backwards-compatible.

@mvz or @chastell can you please re-review and then merge this right away if it looks good to you?

lib/reek/rake/task.rb
```diff
@@ -32,118 +33,83 @@ module Rake
...
-        @reek_opts = ''
+        @name          = name
+        @reek_opts     = ''
         @fail_on_error = true
-        @sort = nil
+        @source_files  = 'lib/**/*.rb'
+        @verbose       = false
 
         yield self if block_given?
-        @config_file ||= Pathname.glob('config/**/*.reek').detect(&:file?)
```
mvz: You will have to somehow restore this globbing as well if you want to merge this for 1.6.2.

mvz:
Hm. Maybe it'll be quicker to just fix the 1.6.x bug and merge this after?

troessner:
>> Hm. Maybe it'll be quicker to just fix the 1.6.x bug and merge this after?

Yes, good point. I'll remove config_files again and will open up a separate PR in a couple of minutes.

mvz:
I've just implemented this too :-).

troessner:
>> I've just implemented this too :-).

You did? Great, I haven't started yet.::D

So I'll just wait for your PR?

troessner:
Ah, it's there already.:D

troessner:
@mvz regarding your comment:

>> I think using an array as argument to system is better because if I remember correctly it avoids the shell re-parsing the command line, so you won't run into any problems if the config file is called my config.reek, or if any of the other file names contain spaces.

Yes,

>> system 'reek', '-c', '/tmp/my config.reek', 'dirty.rb'

will work while

>> system 'reek -c /tmp/my config.reek dirty.rb'

won't. But handling those edge cases will make things significantly more complictated.
Please have a look at the second commit in this PR (i kept that separate on purpose).
This handles the case I describe above but it doesn't handle a couple of more complicated cases.

E.g.:

>> system ...., '--ultra-verbose --quit'

won't work so we need to split REEK_OPTS as well. And then it gets even more complicated because you can also do "-c /tmp/my config.reek" and there is no way we will no how to split this.

Long story short: I believe we shouldn't cater to this. Having "/tmp/my config" is an absolute edge case. I doubt that there are more than 3 reek users out there who have something like this. If any at all. So in my mind we should not handle this edge case and rather keep things simple.

mvz:
Hm, I see, so we're either stuck with having to split REEK_OPTS, or not handling spaces in file names. :cry:

I agree we should probably go for the simple case then and handle the edge cases if and when problems arise.

BTW, since we're basically duplicating a lot of the stuff the RSpec rake task implements, I have a feeling there's room for a gem here. I might start one.

troessner:
Hmmm, I mean we could just ignore that special case with "-c my config.reek" from above (there is nothing we can do anyway) and just add a 

```Ruby
*REEK_OPTS.split("\+s")
```

Then this parameter handling would be done more or less and we can merge this PR.
What do you think?

troessner:
@mvz I think I found a solution that we can all live with:

I updated my second commit so that it uses "system" the way you proposed.
I'd suggest that we merge this PR as those 2 commits and if we run into trouble we just revert the second commit.

What do you think?

mvz:
@troessner that sounds like a good plan. I have made one more comment on the second commit, once that's fixed this can be merged.

troessner:
>> I have made one more comment on the second commit, once that's fixed this can be merged.

Done, we're now using 

>>  ['reek', ...., *source_files].

as well.

But before we can merge this we need to agree on how we treat the 1.6. branch from then on which we are discussing in #365 :)

troessner:
Now that we have that stable-1.6 branch up let's merge this @mvz :)

PR #364, opened 2015-01-11
Fix unparser issue with ruby 1.9.3. by pinning unparser to 0.1.16.

troessner:
The current unparser version 0.1.17 doesn't work with ruby 1.9.3 anymore:

Gem::InstallError: unparser requires Ruby version >= 2.0.0.
An error occurred while installing unparser (0.1.17), and Bundler cannot continue.
Make sure that `gem install unparser -v '0.1.17'` succeeds before bundling.

### DIFF
--- a/reek.gemspec
+++ b/reek.gemspec
@@ -31,7 +31,7 @@ Gem::Specification.new do |s|
   s.summary = 'Code smell detector for Ruby'
 
   s.add_runtime_dependency('parser', ['~> 2.2.0.pre.7'])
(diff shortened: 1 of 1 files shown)


mvz:
Yes, this is a serious problem. Unparser just dropped support for 1.9.3 in the version that added support for 2.2.0. Would be nice if RubyGems could resolve that, but sadly no. I'm trying to convince the author to at least allow installation on 1.9.3 even if he doesn't test on it.

PR #366, opened 2015-01-11
Prefer pry-byebug in specs and features.

troessner:


troessner:
pry-byebug gives us the best of both worlds (pry and byebug): https://github.com/deivid-rodriguez/pry-byebug so i don't see a reason to require 'pry' and 'byebug' separetely.

PR #368, opened 2015-01-13
Do not provide -c option if no config file is found

mvz:
Avoids problem if rake task cannot find any config files.


troessner:
Nice work! Merged. I'll release 1.6.2 right away.

PR #369, opened 2015-01-16
Refactor SmellRepository.

troessner:
Fixes https://github.com/troessner/reek/compare/refactor-smell-repository?expand=1#diff-df382c1a7be57c0d63dd0813c69367e9L12

### DIFF
--- a/lib/reek/core/smell_repository.rb
+++ b/lib/reek/core/smell_repository.rb
@@ -1,4 +1,5 @@
 require 'reek/smells'
+require 'reek/smells/smell_detector'
 
 module Reek
   module Core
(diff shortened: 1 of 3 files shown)


mvz:
Have you considered `SmellDetector.descendents`?

troessner:
Quite frankly, I did this more for the fun of it without giving it much thought (in my current job I never code) but now that I think of it, I believe I could basically just replace the whole enchilada from above with:

```Ruby
class SmellDetector
  def self.descendants
    ObjectSpace.each_object(Class).select { |klass| klass < self }
  end
end
```

or am I missing something here?.:)

mvz:
That should work, but will probably give problems on JRuby where ObjectSpace is disabled by default IIRC. I hadn't realized descentents was a Rails thing, but maybe something like https://github.com/dkubb/descendants_tracker might be an idea? [Here](https://www.ruby-forum.com/topic/193281) is a longish discussion of options, pointing to [this code](http://pastie.org/604847) that seems fairly simple.

Another option I have used recently is having each descendant registering itself explicitely.

troessner:
I went ahead with the "inherited" callback, ready for re-review.

mvz:
Very nice and simple.

Issue #319, opened 2014-11-15
Clean up reek_of/smell_of matchers


mvz:
We have both reek_of and smell_of matchers. This seems silly. Perhaps we can make reek_of smarter so it doesn't check all smells but just the one being tested. This should be possible now that SmellRepository can be initialized with the smells to test.

mvz:
I think we should pick this up for Reek 2.0 as well, since it might involve a breaking change to `#reek_of`.

troessner:
Yes, makes total sense to me.

troessner:
Started to work on this in #371 (which is highly wip).

troessner:
Regarding:

>> We have both reek_of and smell_of matchers. This seems silly. 

Let's compare "smell_of" with "reek_of":

**smell_of:**

spec/reek/smells/long_yield_list_spec.rb:      

```Ruby
expect(src).to smell_of(LongYieldList, count: 4)
```

spec/reek/smells/control_parameter_spec.rb:

```Ruby
expect(src).to smell_of(ControlParameter)
expect(src).to smell_of(ControlParameter, name: 'arg')
```

spec/reek/smells/data_clump_spec.rb:    

```Ruby
expect(src).to smell_of(DataClump,
                            parameters: %w(p1 p2)

expect(src).to smell_of(DataClump,
                            count: 3,
                            parameters: ['pa', 'pb', 'pc'])
```
**reek_of:**

spec/reek/core/module_context_spec.rb:

```Ruby
    expect('
      module Fred
        def simple(x) x + 1; end
      end
    ').to reek_of(:UncommunicativeParameterName, /x/, /simple/)
```

spec/reek/smells/irresponsible_module_spec.rb:


```Ruby
expect(src).to reek_of(:IrresponsibleModule)
```

spec/reek/smells/uncommunicative_variable_name_spec.rb:

```Ruby
expect(src).to reek_of(:UncommunicativeVariableName, /@x/, /Thing/, /variable name/)
```

**Things that I notice when looking at "smell_of":**

- it's not very caller-friendly - it expects a class, and if we pass a wrong constant here the method doesn't catch and handle this in a friendly way but rather the method call itself will raise.
- smell_of is a very sophisticated matcher. It allows for exact matching 
- Nobody would use this matcher as a client, at least not for precise matching. Nobody. This one makes only sense for us, the `reek` contributors, to test `reek` itself. 


**Things that I notice when looking at "reek_of":**

[...]

PR #371, opened 2015-01-18
Replace "smell_of" with "reek_of".

troessner:


troessner:
Ready for review @mvz ! Following the boyscout rule I also fixed the formatting of the spec files I touched in the course of this PR.

troessner:
This is the last thing that's keeping us from releasing reek-2  - anybody up for reviewing this bad boy?.:)

mvz:
I'll take a look this weekend.

lib/reek/spec/should_reek_only_of.rb
```diff
@@ -15,25 +15,53 @@ def matches?(actual)
...
+    #   reek_only_of(UncommunicativeParameterName, name: 'x2')
+    #   reek_only_of(DataClump, count: 3)
+    #
+    # Examples from a real spec
+    #
+    #   expect(src).to reek_only_of(:TooManyStatements,  count: 6)
     #
-    def reek_only_of(smell_category, *patterns)
-      ShouldReekOnlyOf.new(smell_category, patterns)
+    def reek_only_of(smell_category, *smell_details_list)
```
mvz: I don't think you should use splat here and in the other matcher, since we expect exactly zero or one hash argument.

spec/reek/smells/boolean_parameter_spec.rb
```diff
@@ -8,35 +8,35 @@
...
+        expect(src).to reek_of(BooleanParameter, name: 'arga')
       end
       it 'reports two parameters defaulted to booleans' do
         src = 'def cc(nowt, arga = true, argb = false, &blk) end'
-        expect(src).to smell_of(BooleanParameter,
-                                { name: 'arga' },
-                                { name: 'argb' })
+        expect(src).to reek_of(BooleanParameter,
+                               { name: 'arga' },
+                               { name: 'argb' })
```
mvz: Here, `reek_of` will combine the two hashes into just one. I'm not sure which value of `name` it will pick :-).

lib/reek/spec/should_reek_of.rb
```diff
@@ -6,34 +6,76 @@ module Spec
...
+      def initialize(smell_category, *smell_details_list)
+        @smell_category     = normalize smell_category
+        @smell_details_list = smell_details_list
       end
 
       def matches?(actual)
         @examiner = Examiner.new(actual)
         @all_smells = @examiner.smells
-        @all_smells.any? { |warning| warning.matches?(@klass, @patterns) }
+        no_smell_details_given? ? match_without_smell_details? : match_with_smell_details?
```
mvz: Maybe implement `smell_details_given?` and swap this logic around?

lib/reek/smell_warning.rb
```diff
@@ -59,13 +59,21 @@ def yaml_hash
...
+    def common_parameters_equal?(other_parameters)
+      # Why not check for strict parameter equality instead of just the common ones?
+      #
+      # In `self`, `parameters` might look like this:  {:name=>"@other.thing", :count=>2}
+      # Coming from specs, 'other_parameters' might look like this, e.g.:
+      # {:name=>"@other.thing"}
+      # So in this spec we are just specifying the "name" parameter but not the "count".
+      # In order to allow for this kind of leniency we just test for common parameter equality,
+      # not for a strict one.
+      common_keys = parameters.keys &  other_parameters.keys
```
mvz: What about if `other_parameters` has keys that are not in `parameters`?

troessner: Good point. They will be ignored. What do you think about raising ArgumentError?

lib/reek/spec/should_reek_only_of.rb
troessner: What do you mean? We expect an array of hashes here, so 0..n items,  see "match_with_smell_details" above.

lib/reek/spec/should_reek_of.rb
troessner: Good idea, will do.

lib/reek/spec/should_reek_only_of.rb
troessner: Ah yes, of course, good catch, will change that!

spec/reek/smells/boolean_parameter_spec.rb
troessner: Yes, good catch, will fix that.

troessner: >> Here, reek_of will combine the two hashes into just one.

Actually that's not true. Inside of "reek_of" it looks like that:

>> [1] pry(#<Reek::Spec::ShouldReekOf>)> p @smell_details_list
[{:name=>"arga"}, {:name=>"argb"}]
=> [{:name=>"arga"}, {:name=>"argb"}]

I still agree though that the constructor above shouldn't a splat as well.:)

lib/reek/spec/should_reek_only_of.rb
troessner: Actually your comment still confuses me. Shouldn't this comment be rather one line below and say:

>> since we expect exactly one array argument.

?

I think we're misunderstanding each other here. 
It **was** intended that you can do:

```Ruby
expect(src).to reek_of(:UncommunicativeVariableName, { name: x1 }, { name: x2 }, { name: x3 }, ....
```

This is how we use it in the specs and that's why the tomdoc from above says:

>> smells_details_list - An AoH with each hash containing smell_warning parameters

I do agree though that the second splat in the line below is unnecessary and thus confusing.

I improved that in https://github.com/troessner/reek/commit/56057e1202ed8a1792dcb8bd856493a6ebfe95b9
Was this what you had in mind?

mvz: > It was intended that you can do:
> 
> ````
> expect(src).to reek_of(:UncommunicativeVariableName, { name: x1 }, { name: x2 }, { name: x3 }, ....
> ```

Ah, I thought you were getting rid of that possibility, I was confused by the examples in the tomdoc. I'll take another look.

lib/reek/spec/should_reek_of.rb
```diff
@@ -6,34 +6,76 @@ module Spec
...
+        !@smell_details_list.empty?
+      end
+
+      def match_without_smell_details?
+        @all_smells.any? { |warning| warning.matches?(@smell_category) }
+      end
+
+      def match_with_smell_details?
+        @smell_details_list.map do |smell_detail|
+          @all_smells.any? { |warning| warning.matches?(@smell_category, smell_detail) }
```
mvz: I think we shoult have ReekOf with smell details require that it finds the exact number of smells of that category, just like  SmellOf did. What we should allow is reeking of other smell categories.

lib/reek/spec/should_reek_only_of.rb
```diff
@@ -15,25 +15,53 @@ def matches?(actual)
...
+
+      private
+
+      def match_without_smell_details?
+        @warning.matches?(@smell_category)
+      end
+
+      def match_with_smell_details?
+        @smell_details_list.map do |smell_detail|
+          @warning.matches?(@smell_category, smell_detail)
```
mvz: You're matching every smell in the smell details list to the same warning.

lib/reek/spec/should_reek_only_of.rb
mvz: I'm still confused. Your doc says we expect exactly one warning. If that's the case, we don't need an AoH for the smell details, we just need a Hash.

lib/reek/spec/should_reek_only_of.rb
troessner: Good point, doesn't make any sense.

lib/reek/spec/should_reek_only_of.rb
troessner: Now I realize that my tomdocs above were confusing, should have been:

>>
# Examples
#
#   reek_only_of(UncommunicativeParameterName, { name: 'x2' }, { name: 'x3'} )

instead of:

>>
# Examples
#
#   reek_only_of(UncommunicativeParameterName, name: 'x2')

Now that I think of it this is more confusing than I had initially imagined.

So let's go with just one Hash and not an AoH - what do you think about that?

troessner:
I had thought that we had discussed this enough in issue #319  but now I realize that we probably haven't.

So let's get this sorted out!

I'd say we go for:

```Ruby
reek_of(DuplicateMethodCall, name: 'thingie', count: 4)
```

With the second argument being one hash that is optional. 

But if this hash is given, it 

1.) must match exactly so if "count" is 5 this match would fail.
2.) if this hash contains non-sensical keys like "mame" (mind the typo) we raise ArgumentError to make it clear to the user (which would be mostly us)

The same would apply for reek_only_of.

WDYT?

mvz:
@troessner this seems the way to go. Some edge cases:

* `reek_of` doesn't mind if there are other smells of a different category. `reek_only_of` will fail in that case.
* If there are two smells of the given category, `reek_of` with hash is happy if one of them matches the hash.
* If there are two smells of the given category, `reek_only_of` with hash will fail; `reek_only_of` without hash will not fail.
* `should_not reek_only_of` should fail if there are no smells that match the given category and hash, e.g.,  if there are no smells at all, or if there are only smells of other categories.

troessner:
>> reek_of doesn't mind if there are other smells of a different category. reek_only_of will fail in that case.

Ack.

>> If there are two smells of the given category, reek_of with hash is happy if one of them matches the hash.

Ack.

>> If there are two smells of the given category, reek_only_of with hash will fail; reek_only_of without hash will not fail.

I don't get it, why?

Consider this code:

```Ruby
src = <<-EOS
  def(x, y)
    puts "hossa"
  end
EOS
```

`reek` will report 2  UnusedParameters warnings.

So 

```Ruby
reek_only_of(UnusedParameters, name: 'x')
```

would fail but

```Ruby
reek_only_of(UnusedParameters)
```

wouldn't? What's the rationale behind this?
In my mind, this is not intuitive and too complicated.
How about having both examples succeed?

>>should_not reek_only_of should fail if there are no smells that match the given category and hash, e.g., if there are no smells at all, or if there are only smells of other categories.

This sounds to complicated to me. But let's rather discuss this after the points above.

troessner:
P.S.: I revamped the docs for reek_of significantly to make sure that we are on the same  page.

troessner:
P.P.S:

I vote for keeping things simple:

[1] "reek_only_of" succeeds if source reeks only of one and the same category / type.
[2]  This category / type can occur multiple times
[3] If we pass a hash [1] and [2] still apply. Only this time we're looking for a smell that matches the smell_detail.

To me this seems to be the most intuitive and simple solution and the way to go.

What do you think?

mvz:
Does your step [3] mean, that all found smells must match both the category and the hash? That would be most intuitive to me.

As an aside, perhaps `reek_only_of` with a hash is too complicated, because it is unclear what should be meant. We only use it in a couple of places, and I think those would work just fine with `reek_only_of` with just the smell category (and even where we use it, the 'only' part is sometimes not relevant, or the spec's clarity would probably be improved by splitting it into `not reek_of(X)` plus `reek_of(Y)`, because other smells are not relevant).

troessner:
>> Does your step [3] mean, that all found smells must match both the category and the hash? That would be most intuitive to me.

No, just one would have to match.:)

>> As an aside, perhaps reek_only_of with a hash is too complicated, because it is unclear what should be meant.

Sounds reasonable. Let's drop the smell_details for "reek_only_of".

troessner:
@mvz I believe I implemented everything we talked about. Final review please.:)

lib/reek/spec/should_reek_only_of.rb
```diff
@@ -15,25 +15,30 @@ def matches?(actual)
       def matches_examiner?(examiner)
         @examiner = examiner
         @warnings = @examiner.smells
-        @warnings.length == 1 && @warnings[0].matches?(@klass, @patterns)
+        return false unless @warnings.length == 1
+        @warning = @warnings[0]
+        @warning.matches?(@smell_category, @smell_details)
```
mvz: I think these three lines should be replaced by `@warnings.all? { |w| w.matches?(@smell_category) }`.

spec/factories/factories.rb
```diff
@@ -20,7 +20,7 @@
     context 'self'
     lines [42]
     message 'smell warning message'
-    parameters     {}
+    parameters Hash.new
```
mvz: RuboCop 0.29 doesn't like this (and neither do I :smile:). Perhaps `parameters { {} }` ?

lib/reek/spec/should_reek_only_of.rb
troessner: Good catch!
However we still need a variation of 

>> return false unless @warnings.length == 1

namely:

>> return false if @warnings.empty?

otherwise this matcher will return true on an empty smell list.

The other suggestion I incorporated though.

troessner:
@mvz incorporated all your feedback - merge?..:)

mvz:
Awesome! Merging....

troessner:
P.S.: I also updated our wiki: https://github.com/troessner/reek/wiki/RSpec-matchers
I think this information should be exhaustive to get everybody started.

Issue #254, opened 2014-03-20
Clean up/organize command line options


mvz:
For reek 1.4 (or 2.0), we should take a good look at the set off command line options. Right now, they're a little bit of a mess due to backwards-compatibility considerations and organic growth. Let's see if we can restructure them and perhaps even get rid of some of them.

mvz:
So summarize, what I'll be doing is re-organizing and simplifying the set of switches, but keeping the total set of options intact.

troessner:
Closing now that #372 is merged.

PR #372, opened 2015-01-18
Re-organize command line options

mvz:
Work-in-progress implementation of #254.

Done:
* Verbosity boiled down to showing headings for fragrant files or not
* Line number options simplified
* Line number options are fully separated from `--ultra-verbose`: We can now have links showing, but line numbers hidden.

To do:
* [x] Is it sensible to have `-o` as short option for `--color`, or even have a short option at all, since it is the default?
* [x] Is `--ultra-verbose` a good name? How about `--show-links` or `--wiki-links`?
* [ ] Should `--single-line` be a fully separate kind of report
* [x] Perhaps we can replace all these format options with `--format`, with values `text`, `single-line`, `yaml`, `html`?
* [x] Replace `--sort-by-issue-count` with `--sort=..` option


troessner:
>> Is it sensible to have -o as short option for --color, or even have a short option at all, since it is the default?

Excellent point, let's throw it out.

>> Is --ultra-verbose a good name? How about --show-links or --wiki-links?

I think I like "--wiki-links" better than "--ultra-verbose", let's go with that.:)

>> Should --single-line be a fully separate kind of report? Perhaps we can replace all these format options with --format, with values text, single-line, yaml, html?

Good idea! I also think that "--format" is the way to go.

mvz:
I've sort of decided to keep `--single-line` as a suboption of `--format=text`.

mvz:
I ended up doing a bit of refactoriong as well: There is now only one 'Command', since help and version are trivial to inline inside `Options`. There is more to be done (ReekCommand should absorb Command and be renamed; Options should be renamend).

I decided not to make `--single-line` into a separate format, since then it becomes very unclear which options apply to what.

And with that, I invite comments to this pull request.

troessner:
I'll review this bad boy today..;)

features/command_line_interface/options.feature
```diff
@@ -25,29 +25,33 @@ Feature: Reek can be controlled using command-line options
...
+          -U, --[no-]wiki-links            Show link to related Reek wiki page for each smell
+          -n, --[no-]line-numbers          Show line numbers in the output (this is the default)
+          -s, --single-line                Show location in editor-compatible single-line-per-smell format
+              --sort-by SORTING            Sort reported files by the given criterium:
+                                             smelliness ("smelliest" files first)
+                                             none (default - output in processing order)
+
+      Utility options:
+          -h, --help                       Show this message
+          -v, --version                    Show version
```
troessner: :+1:

features/reports/reports.feature
```diff
@@ -51,6 +51,12 @@ Feature: Correctly formatted reports
       9 total warnings
       """
 
+    Examples:
+      | option         |
+      |                |
+      | --sort-by none |
+      | --sort-by n    |
```
troessner: "--sort-by n" seems weird. Why not get rid of this?

troessner: Actually why not get rid of the complete "none" option? What do we gain from it?

features/reports/reports.feature
```diff
@@ -204,4 +228,20 @@ Feature: Correctly formatted reports
     Examples:
       | option           |
       | -U               |
-      | --ultra-verbose  |
+      | --wiki-links  |
+
+  Scenario Outline: --wiki-links is independent of --line-numbers
```
troessner: A little confused here - why is this scenario in here? Seems arbitrary to test a random coupling of options.

lib/reek/cli/help_command.rb
```diff
@@ -1,15 +0,0 @@
...
-    # A command to display usage information for this application.
-    #
-    class HelpCommand < Command
-      def execute(view)
-        view.output(@options.help_text)
-        view.report_success
-      end
-    end
-  end
-end
```
troessner: :1

lib/reek/cli/input.rb
```diff
@@ -1,7 +1,7 @@
 require 'reek/source'
 
 module Reek
-  module CLI
+  module Cli
```
troessner: But "CLI" is an acronym, not a word. I have no hard feelings towards this but it seems a little strange to me...

lib/reek/cli/option_interpreter.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      def report_class
+        case @options.report_format
+        when :yaml
+          Report::YamlReport
+        when :html
+          Report::HtmlReport
+        else # :text
+          Report::TextReport
+        end
```
troessner: Very nice!

lib/reek/cli/option_interpreter.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      def warning_formatter
+        klass = if @options.show_links
+                  Report::UltraVerboseWarningFormatter
```
troessner: Then we probably should rename the Formatter as well to "ShowLinksWarningFormatter" - what do you think?

lib/reek/cli/option_interpreter.rb
```diff
@@ -0,0 +1,82 @@
...
+        end
+      end
+
+      def sort_by_issue_count
+        case @options.sorting
+        when :smelliness
+          true
+        else
+          false
+        end
```
troessner: Why not just 

```Ruby
def sort_by_issue_count
  @options.sorting == :smelliness
end
```

lib/reek/cli/options.rb
```diff
@@ -14,110 +9,122 @@ module Cli
...
+        @options.argv = @argv
+        Rainbow.enabled = @options.colored
+        @options
       end
 
+      private
+
       def banner
-        progname = @parser.program_name
+        program_name = @parser.program_name
```
troessner: Ah, special thanks for that.:)

lib/reek/cli/options.rb
```diff
@@ -14,110 +9,122 @@ module Cli
...
+        set_configuration_options
+        set_alternative_formatter_options
+        set_report_formatting_options
+        set_utility_options
+      end
+
+      def set_utility_options
+        @parser.separator "\nUtility options:"
+        @parser.on_tail('-h', '--help', 'Show this message') do
+          puts @parser
```
troessner: I like this a lot. Using separate commands for this always seemed weird.

lib/reek/cli/options.rb
```diff
@@ -14,110 +9,122 @@ module Cli
...
-        @parser.separator 'Common options:'
-        @parser.on('-h', '--help', 'Show this message') do
-          @command_class = HelpCommand
+        set_configuration_options
+        set_alternative_formatter_options
+        set_report_formatting_options
+        set_utility_options
+      end
+
+      def set_utility_options
```
troessner: Can we order the methods below the way they are used in set_up_parser? Just makes it a little easier to follow the flow.

troessner:
Reviewed, looks great overall.

features/reports/reports.feature
mvz: At some point we may want to support default CLI arguments in the reek configuration (I think there's an issue for that). Then, it becomes necessary to be able to undo the `--sort-by smelliness`. The `n` bit is to show that OptionParser will accept unique prefixes for these parameters and these will be handled correctly. That bit is not really an important feature so I can take it out of the scenario.

lib/reek/cli/input.rb
mvz: It's `Cli` everywhere else in the codebase. I can change it all to `CLI` if you prefer. I agree that would make more sense.

lib/reek/cli/option_interpreter.rb
mvz: D'oh! I'll change it.

lib/reek/cli/option_interpreter.rb
mvz: I agree.

features/reports/reports.feature
mvz: In 1.6, you cannot have wiki links with no line numbers or with the single-line format.

features/reports/reports.feature
troessner: Ok, understood - then let's leave it like it is.

features/reports/reports.feature
troessner: Ah..:)

lib/reek/cli/input.rb
troessner: Alright, then let's leave it like it is and fix the spelling in a separate PR some time later.

troessner:
Looking good! Please squash so I can merge.

mvz:
Done!

guilhermesimoes:
Don't forget to update the README regarding the format options when you release a new version.

troessner:
Great catch @guilhermesimoes ! I just created https://github.com/troessner/reek/issues/383 so we don't forget to update this.

PR #374, opened 2015-01-21
Fix for running reek on a build server

maser:
We use reek on Jenkins like this: `reek .`

We upgraded to the latest version and noticed that reek was suddenly not reporting anything anymore. After a lot of debugging we figured out that reek ignores the `.` we pass because it assumes we are piping Ruby code in.

I gave files given as arguments higher priority so reek only reads from `$stdin` when it is given no files.

### DIFF
--- a/lib/reek/cli/input.rb
+++ b/lib/reek/cli/input.rb
@@ -7,14 +7,14 @@ module Reek
     #
     module Input
       def sources
-        if input_was_piped?
-          source_from_pipe
-        else
-          if no_source_files_given?
-            working_directory_as_source
+        if no_source_files_given?
+          if input_was_piped?
+            source_from_pipe
           else
-            sources_from_argv
+            working_directory_as_source
           end
+        else
+          sources_from_argv
         end
       end
 


troessner:
>> After a lot of debugging we figured out that reek ignores the . we pass because it assumes we are piping Ruby code in.

I'm not sure I follow - how can we reproduce this?

If use the latest reek version and do a 

>> cd my_project
reek .

this works out exactly as I would expect: It uses the current directory.

maser:
You can do this to reproduce the behavior we see on Jenkins:

`reek -y ~/foo.rb </dev/null`

Take this sample Jenkins job:

```
cat > foo.rb <<EOT
class Dirty
  # This method smells of :reek:NestedIterators but ignores them
  def awful(x, y, offset = 0, log = false)
    puts @screen.title
    @screen = widgets.map {|w| w.each {|key| key += 3}}
    puts @screen.contents
  end
end
EOT

reek -y foo.rb
```

With reek 1.6.3 it does this:

```
+ reek -y foo.rb
--- []
```

And with 1.6.1 it does this:

```
+ reek -y foo.rb
---
- smell_category: IrresponsibleModule
  smell_type: IrresponsibleModule
  source: foo.rb
  context: Dirty
  lines:
  - 1
[…]
```

The reason is that `$stdin.tty?` returns false.

troessner:
>> reek -y ~/foo.rb </dev/null

Ah, ok!

Looks good to me, merged!

Thanks a lot for this, I'll release 1.6.4 right away.

troessner:
And 1.6.4 is released with your bugfix, thanks!

maser:
Great! We updated to 1.6.4 on our Jenkins server and it works fine.

PR #378, opened 2015-01-31
Fix YAML export on JRuby

mvz:
So, Psych doesn't actually install on JRuby, so my previous fix didn't work for that.

Solution: Decide we're not writing a two-way converter between yaml and SmellWarning, but that instead we're writing an exporter. The result: This PR lets each SmellWarning convert to a hash with `#yaml_hash` and then calls `#to_yaml` on the resulting array.


troessner:
Nice!

PR #382, opened 2015-02-08
Fix generating HTML report

guilhermesimoes:
Fix needed due to #346


troessner:
Great! Thanks for that, merged.;)

guilhermesimoes:
Yey, I got a commit into Reek! :dancer:

mvz:
@troessner perhaps we need to patch 1.6 with this too?

troessner:
Yes! I just released 1.6.6 with the fix.:)

PR #386, opened 2015-02-15
Remove deprecated API for Examiner.

troessner:


troessner:
P.S.: I know that the code I removed says "deprecated" and "API" but we really shouldn't care about that.

1.) It's a reek internal class (yes, one of the core ones, but still) 
2.) I have a hard time imagening that anybody is using this. From my experience I'd say 99% of all reek users just use the rake task and the spec matchers.
3.) And finally: This has been deprecated for long enough now, time to remove it.

And if someone does in this world he'll have to upgrade. Backwards-compatibility is important but not to the point where it basically kills all refactoring for no tangible bonus. Let's be more valiant with things like that.

mvz:
Agreed.

PR #388, opened 2015-02-15
Refactor smells module.

troessner:


lib/reek/smells.rb
```diff
@@ -1,28 +1,8 @@
...
-require 'reek/smells/uncommunicative_module_name'
-require 'reek/smells/uncommunicative_parameter_name'
-require 'reek/smells/uncommunicative_variable_name'
-require 'reek/smells/unused_parameters'
-require 'reek/smells/utility_function'
-# SMELL: Duplication -- all these should be found automagically
+smell_glob = Pathname.new('..').
+             join('smells').
+             expand_path(__FILE__).
+             join('*.rb')
```
mvz: It seems odd to start with `'..'`...

mvz: How about this:

```
smell_glob = Pathname.new(__FILE__).dirname.realpath.join('smells', '*.rb')
```

mvz:
Even nicer: https://github.com/jarmo/require_all

mvz:
If I read the docs correctly, that would just be:

```
require_rel 'smells'
```

for our case.

troessner:
>> Even nicer: https://github.com/jarmo/require_all

I know require_all and I think its awesome -  I just didn't want add more dependencies to reek. 
But maybe we shouldn't be so strict with that if the gem itself is small and self-contained.

Unless anybody objects I'll integrate it.

mvz:
:+1:

troessner:
Ok, this looks a lot more awesome with require_all. Please re-review.:)

PR #389, opened 2015-02-16
Support more Rubies

mvz:
- Do not run rubocop task on Rubinius by default
- Add CRuby head to Travis CI config
- No longer allow build failures on Rubinius and JRuby


troessner:
Very nice!

Issue #373, opened 2015-01-19
Is anyone using the Cucumber annotations?


mvz:
Our cucumber feature files are full of annotations like `@yaml` that indicate which general feature of Reek is being tested there. I never use this, and always just run either the entire suite or a pass the files (optionally with a line number) I want to run to cucumber.

If no-one else is using this, I'd like to remove the annotations altogether.

troessner:
I also never use them, lets throw 'em out.

PR #391, opened 2015-02-16
Remove cucumber annotations

mvz:
#373.

### DIFF
--- a/features/command_line_interface/options.feature
+++ b/features/command_line_interface/options.feature
@@ -1,4 +1,3 @@
-@options
 Feature: Reek can be controlled using command-line options
   In order to change reek's default behaviour
   As a developer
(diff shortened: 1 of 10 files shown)


troessner:
PR looks good but travis is failing big time.;)

chastell:
Sorry for the thread necromancy (hi!), but weren’t `@ruby20` and `@ruby21` useful for skipping them on Ruby 1.9 and 2.0? Ah, maybe the parser gem parses new Ruby syntax just fine even when running under Ruby 1.9…

mvz:
Yes, we explicitely use parser's ruby 2.1 parser to allow reek to parse the new syntax even on older rubies. We could move to 2.2 even.

chastell:
Cool – and thanks for the explanation, that makes perfect sense.

PR #393, opened 2015-02-16
Refactor smell specs.

troessner:
**What I did:**

1.) Properly scoped things. No more 

```Ruby
include Reek
include Reek::Smells
```

And the same for Reek::Core::CodeContext and so on.

2.) Fixed the formatting for heredocs, newlines, indentation and so on.
3.) Used factories were possible
4.) Fixed all calls to reek_of and friends to use symbols like this:

```Ruby
expect(src).to reek_of(:BooleanParameter)
```

**What I left to do:**

1.) I left the "@source_name" in here:
https://github.com/troessner/reek/pull/393/files#diff-ef046f7765020f5fa43c96d8361fb3bfR8
because this is something we need to fix conceptually in the shared_example
2.) I didn't fix all the requires in the smells. There's still some room left for refactoring.

### DIFF
--- a/spec/reek/smells/attribute_spec.rb
+++ b/spec/reek/smells/attribute_spec.rb
@@ -3,13 +3,10 @@ require 'reek/smells/attribute'
 require 'reek/core/module_context'
 require 'reek/smells/smell_detector_shared'
 
-include Reek::Core
-include Reek::Smells
-
-describe Attribute do
+describe Reek::Smells::Attribute do
   before :each do
-    @source_name = 'ticker'
-    @detector = Attribute.new(@source_name)
+    @source_name = 'dummy_source'
+    @detector = build(:smell_detector, smell_type: :Attribute, source: @source_name)
   end
 
   it_should_behave_like 'SmellDetector'
@@ -17,7 +14,7 @@ describe Attribute do
   context 'with no attributes' do
     it 'records nothing in the module' do
       src = 'module Fred; end'
-      ctx = CodeContext.new(nil, src.to_reek_source.syntax_tree)
+      ctx = Reek::Core::CodeContext.new(nil, src.to_reek_source.syntax_tree)
       expect(@detector.examine_context(ctx)).to be_empty
     end
   end
@@ -29,7 +26,7 @@ describe Attribute do
 
     shared_examples_for 'one attribute found' do
       before :each do
-        ctx = CodeContext.new(nil, @src.to_reek_source.syntax_tree)
+        ctx = Reek::Core::CodeContext.new(nil, @src.to_reek_source.syntax_tree)
         @smells = @detector.examine_context(ctx)
       end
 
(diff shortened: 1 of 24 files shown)


spec/reek/smells/too_many_statements_spec.rb
```diff
@@ -31,34 +26,12 @@ def process_singleton_method(src)
...
-          @libruby         = "/lib/ruby"
-          @librubyver      = "/lib/ruby/"
-          @librubyverarch  = "/lib/ruby/"
-          @siteruby        = "lib/ruby/version/site_ruby"
-          @siterubyver     = siteruby
-          @siterubyverarch = "$siterubyver/['arch']}"
-        end
-      end
     EOS
-    expect(src).to reek_only_of(:TooManyStatements)
```
troessner: This whole spec doesnt make any sense anymore since we refactored reek_of and friends. This spec was implicit by failing if we had had multiple TooManyStatements which we now explicitly allow.

troessner:
@mvz any chance you get to review this PR soon? Just asking cause this PR is blocking me from further spec refactorings...

spec/reek/smells/too_many_methods_spec.rb
```diff
@@ -1,95 +1,78 @@
...
-include Reek::Smells
-
-describe TooManyMethods do
+describe Reek::Smells::TooManyMethods do
   before(:each) do
-    @source_name = 'elephant'
-    @detector = TooManyMethods.new(@source_name)
+    @source_name = 'dummy_source'
+    @detector = described_class.new(@source_name)
+    @detector.configure_with 'max_methods' => 2
```
mvz: Nice simplification!

spec/reek/smells/too_many_statements_spec.rb
mvz: Yeah I can't figure out where this test came from in the first place. There must have been some issue that triggered this, perhaps some other smell also being reported? Seems reasonable to get rid of it.

spec/reek/smells/attribute_spec.rb
```diff
@@ -3,13 +3,10 @@
...
-include Reek::Core
-include Reek::Smells
-
-describe Attribute do
+describe Reek::Smells::Attribute do
   before :each do
-    @source_name = 'ticker'
-    @detector = Attribute.new(@source_name)
+    @source_name = 'dummy_source'
+    @detector = build(:smell_detector, smell_type: :Attribute, source: @source_name)
```
mvz: I'm not sure this is simpler/clearer than `Reek::Smells::Attribute.new(source: @source_name)`.

mvz:
I'm a little unsure about the factories in this case. It seems to me the call to the factory only makes it more complex. What do you think?

troessner:
>> I'm a little unsure about the factories in this case. It seems to me the call to the factory only makes it more complex. 

I had the same thought but after giving it some time I kind of decided to stick with factories - it might be more to write (which kind of every ruby'ist is against) **but** they:

* provide a cleaner abstraction since you don't even have to be aware of the namespace "Reek::Smells"
* unify the way we create smells in our specs which will decrease confusion and increase readability
* make it easier to refactor our specs

Hence I'd vote for leaving them - what's your take on that?

mvz:
I'm not fully convinced about the factories, but let's give it a try.

PR #397, opened 2015-02-26
Escape Regexp-like Strings on CodeContext matches

chastell:
I recently noticed a bug where a `|`-ending entry in an `exclude` config setting for a given smell would effectively exclude everything from the given smell; the below excludes _all_ methods (not just `Foo#|`) from being checked aginst `UncommunicativeVariableName`:

```yaml
UncommunicativeVariableName:
  exclude:
    - Foo#|
```

Today I noticed a bug where a `[]`-ending entry blows up the whole tool; the below makes Reek not work at all:

```yaml
UncommunicativeParameterName:
  exclude:
    - Bar#self.[]
```

I tracked down the culprit of both of these cases – the excludes were treated as regular expressions rather than straight strings.

### DIFF
--- a/lib/reek/core/code_context.rb
+++ b/lib/reek/core/code_context.rb
@@ -29,7 +29,10 @@ module Reek
 
       def matches?(candidates)
         my_fq_name = full_name
-        candidates.any? { |str| /#{str}/ =~ my_fq_name }
+        candidates.any? do |candidate|
+          candidate = Regexp.quote(candidate) if candidate.is_a?(String)
+          /#{candidate}/ =~ my_fq_name
+        end
       end
 
       #
--- a/spec/reek/core/code_context_spec.rb
+++ b/spec/reek/core/code_context_spec.rb
@@ -25,11 +25,17 @@ describe CodeContext do
     it 'does not match when its own short name is not given' do
       expect(@ctx.matches?(['banana'])).to eq(false)
     end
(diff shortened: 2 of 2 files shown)


mvz:
There may be backwards-compatibility issues for people taking advantage of the fact that strings are interpreted as regexes. :cry:

chastell:
Hm, I think this borders on the classic ‘if someone potentially depending on a buggy behaviour means a bugfix is a breaking change then _every_ bugfix would require a major version bump’ problem.

I looked at Reek docs and it seems [Basic Smell Options](https://github.com/troessner/reek/wiki/Basic-Smell-Options) explicitly suggests that if you want to use a `Regexp` then you should mark it as one:

```yaml
ControlCouple: 
  exclude:
  - !ruby/regexp /write/
```

The problem with the current version is that – contrary to what the other examples suggest – you _can’t_ use a string, because you have to explicitly escape it by hand. Also, excludes ending with `|` silently hide other excludes and excludes ending with `[]` blow up Reek.

I think all of the above outweight the potential breakage (especially that if someone depends on regular expressions in the config they can still use regular expressions).

mvz:
I agree. If the documentation explicitely mentions regexps like that we can safely make this change.

troessner:
@chastell and 2.0.1 is released.:)

chastell:
:sparkles:

PR #398, opened 2015-02-26
Only install byebug where it compiles

mvz:
This will allow switching to, e.g., JRuby locally and not have to remove `:local_development` from the bundle.


troessner:
Good catch!

PR #402, opened 2015-03-02
Adds JSON output format

leonelgalan:
Includes both rspec and cucumber tests as seen on the yaml report.
_____

I needed JSON output while setting up a CI with Reek. I was using the YamlReport and `YAML.parse(output).to_ruby` to get a Hash, but I thought this could be useful for others as well.

I'm getting some 1 failure on rspec and 1 on cucumber, but they are not related to the files changed on this PR. I'm getting the same failures when pulling from master. Maybe is something on my local setup, so I thought I'll give this a try.

### DIFF
--- a/lib/reek/cli/report/report.rb
+++ b/lib/reek/cli/report/report.rb
@@ -1,4 +1,5 @@
 require 'rainbow'
+require 'json'
 
 module Reek
   module Cli
@@ -98,6 +99,15 @@ module Reek
       class YamlReport < Base
         def show
           print smells.map(&:yaml_hash).to_yaml
+        end
+      end
+
+      #
+      # Displays a list of smells in JSON format
+      # JSON with empty array for 0 smells
+      class JsonReport < Base
+        def show
+          print ::JSON.generate(smells.map(&:yaml_hash))
         end
       end
 
--- a/spec/reek/cli/json_report_spec.rb
+++ b/spec/reek/cli/json_report_spec.rb
@@ -1,0 +1,20 @@
+require 'spec_helper'
+require 'reek/examiner'
+require 'reek/cli/report/report'
+require 'reek/cli/report/formatter'
+
+describe Reek::Cli::Report::JsonReport do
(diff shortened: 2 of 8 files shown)


mvz:
Very useful. Looks good to me.

spec/reek/cli/json_report_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+require 'reek/examiner'
+require 'reek/cli/report/report'
+require 'reek/cli/report/formatter'
+
+include Reek
+include Reek::Cli
```
troessner: I know we are doing this a lot in our codebase but at least for new specs, lets not do the "include Reek" dance and use fully qualified constants.

troessner:
Excellent addition!
Lets fix this one thing i commented on and re-squash - then i'll happily merge..;)

lib/reek/cli/report/report.rb
```diff
@@ -102,6 +103,15 @@ def show
       end
 
       #
+      # Displays a list of smells in JSON format
+      # JSON with empty array for 0 smells
+      class JsonReport < Base
```
chastell: Does it have to be `JsonReport` rather than `JSONReport`?

(The only reason for these capitalisation abominations I heard about was related to Rails’ code autoloading, but I’m working on a Rails 2.3 app at the moment and even it would find `JSONRerport` in `json_report.rb` just fine.)

leonelgalan:
Fixed and squashed! Thanks again for building Reek!

@chastell, I feel you but we the same applies to Yaml, Html, Cli, etc. It might require a major refactor to change all this and add consistency across the project.

troessner:
Great job, merged, thanks!

chastell:
> @chastell, I feel you but we the same applies to Yaml, Html, Cli, etc. It might require a major refactor to change all this and add consistency across the project.

Challenge accepted. :bowtie:

PR #403, opened 2015-03-03
Rake task: drop the no longer supported --quiet option

chastell:
Running `rake reek` on the current master blows up with the `invalid option: --quiet` error.


troessner:
Nice catch!

PR #409, opened 2015-03-16
A variable can also be a shadowarg

awwaiid:
There is some little-known syntax which allows you to create a shadow variable, like:

    def foo(x)
        x.each do |thing ; x, y |
          #... x is local, not the param; y is also a local shadowarg
        end
    end

I'm not sure where a good place to put a test for this is -- let me know and I'll add one.

### DIFF
--- a/lib/reek/smells/uncommunicative_variable_name.rb
+++ b/lib/reek/smells/uncommunicative_variable_name.rb
@@ -107,6 +107,8 @@ module Reek
           case subexp.type
           when :mlhs
             recursively_record_variable_names(accumulator, subexp)
+          when :shadowarg
+            record_variable_name(exp, subexp.shadowarg, accumulator)
           else
             record_variable_name(exp, subexp.name, accumulator)
           end


troessner:
>> There is some little-known syntax which allows you to create a shadow variable

Not sure this is correct. Let me create a slightly better readable example:

```Ruby
 a = [[:f1, :f2, :f3], [:f4, :f5, :f6]]

def no_semicolon(a); a.each { |c,b,a| p "a is #{a}" }; end
def with_semicolon(a); a.each { |c;b,a| p "a is #{a}" }; end
```

Would give

```Bash
[10] pry(main)> no_semicolon a
"a is f3"
"a is f6"
[11] pry(main)> with_semicolon a
"a is "
"a is "
```
 
The reason for this is that the ";" in the block parameters causes all parameters after the ";" to be set to nil. But in both cases the local block parameter "a" is shadowing the outer "a", so I don't understand how the ";" is relevant here?

Both cases from above are not handled by reek right now:

>> ( $ ) echo 'def with_semicolon(a); a.each { |c;b,a| p "a is #{a}" }; end' | reek
/home/timo/.rvm/gems/ruby-2.1.1/gems/reek-2.0.1/lib/reek/smells/uncommunicative_variable_name.rb:111:in `block in recursively_record_variable_names': undefined method `name' for (shadowarg :b):#<Class:0x00000002bc3f48> (NoMethodError)
	from /home/timo/.rvm/gems/ruby-2.1.1/gems/reek-2.0.1/lib/reek/smells/uncommunicative_variable_name.rb:106:in `each'
	from /home/timo/.rvm/gems/ruby-2.1.1/gems/reek-2.0.1/lib/reek/smells/uncommunicative_variable_name.rb:106:in `recursively_record_variable_names'
	from /home/timo/.rvm/gems/ruby-2.1.1/gems/reek-2.0.1/lib/reek/smells/uncommunicative_variable_name.rb:101:in `block in find_block_argument_variable_names'
	from /home/timo/.rvm/gems/ruby-2.1.1/gems/reek-2.0.1/lib/reek/smells/uncommunicative_variable_name.rb:100:in `each'

And:

>>   ( $ ) echo 'def no_semicolon(a); a.each { |c,b,a| p "a is #{a}" }; end' | reek
$stdin -- 4 warnings:
  [1]:no_semicolon has the parameter name 'a' (UncommunicativeParameterName)
  [1]:no_semicolon has the variable name 'a' (UncommunicativeVariableName)
  [1]:no_semicolon has the variable name 'b' (UncommunicativeVariableName)
  [1]:no_semicolon has the variable name 'c' (UncommunicativeVariableName)

What from above is your pull request addressing?

awwaiid:
This is for the first case where a stack error is thrown. When I get some time I'll make a test to demonstrate. Here are some other variations if it helps, where I am forcing b to already exist and there are more than 2 (destructured) params to the each block.

    a = [[:f1, :f2, :f3], [:f4, :f5, :f6]]
    def no_semicolon1(a); b = "hello" ; a.each { |a| p "a is #{a}, b is #{b}" }; end
    def no_semicolon2(a); b = "hello" ; a.each { |a, b| p "a is #{a}, b is #{b}" }; end
    def with_semicolon(a); b = "hello" ; a.each { |a; b| p "a is #{a}, b is #{b}" }; end
    no_semicolon1(a)
    no_semicolon2(a)
    with_semicolon(a)

    output:
    "a is [:f1, :f2, :f3], b is hello"
    "a is [:f4, :f5, :f6], b is hello"
    "a is f1, b is f2"
    "a is f4, b is f5"
    "a is [:f1, :f2, :f3], b is "
    "a is [:f4, :f5, :f6], b is "

mvz:
The better solution is to define `#name` on the shadowarg node. This would solve the recursive case as well. You can do this by adding a module to `lib/reek/source/sexp_extensions.rb`.

mvz:
By the way, since the errors occur in UncommunicateVariableName, tests for this are best placed in `spec/reek/smells/uncommunicative_variable_name_spec.rb`.

mvz:
@awwaiid do you have time to look into this in the near future? Otherwise, I'm happy to pick up this bug.

mvz:
I've implemented the suggested spec and fix in #421.

awwaiid:
Awesome, thanks @mvz! Sorry I didn't get around to it myself... next time!

PR #410, opened 2015-03-20
Fix punctuation (reeks -> reek's)

andyw8:


chastell:
Thanks! Good catches.

Issue #53, opened 2009-10-02
Feature Envy reported even when it is reporting Utility Function


ashmoran:
For example:

    class Level1Generator
      def classes(row, col, value)
        (row == 1 || col == 1) ? %w[ header ] : [ ]
      end
    end

Gets both smells.  Surely it's not informative to report Feature Envy in this case?

mvz:
It seems that any method that triggers UtilityFunction also triggers FeatureEnvy. Therefore, my proposal is to remove UtilityFunction entirely. We can then concentrate on making FeatureEnvy behave in the case of class or module functions (#27, #112).

mvz:
@troessner, @chastell what do you think?

troessner:
I can't think of a meaningful distinction between UtilityFunction and FeatureEnvy. 
So I agree, let's get rid of UtilityFunction and merge it with FeatureEnvy.

chastell:
I think these are different with regards to suggesting ‘what should be done’ – UtilityFunction suggests ‘this does not need instance state, so maybe turn it into a class method’ while FeatureEnvy suggests ‘this might belong somewhere else entirely’. That said, the ‘make it into a class method’ suggestion is not very object-oriented (and a special case of ‘move it somewhere else’, really), so I’m for merging these two.

I wonder whether this won’t have the impact of ‘some of my methods were first and foremost UtilityFunctions (to be fixed by extracting them to the state-less, class level) and some had only FeatureEnvy; now all of them have FeatureEnvy, and a lot of them clearly belong close to where they’re used now, so I’ll better blacklist that smell entirely’, but I doubt that’s a good reason to keep UtilityFunction. :)

troessner:
>> I think these are different with regards to suggesting ‘what should be done’ – UtilityFunction suggests ‘this does not need instance state, so maybe turn it into a class method’ while FeatureEnvy suggests ‘this might belong somewhere else entirely’.

Yes, very well put..:)

mvz:
Fixed in #418.

PR #416, opened 2015-04-01
[WIP] Remove utility function

mvz:
Possible solution for #53.

Some considerations:
* Perhaps the UtilityFunction specs need to be moved to FeatureEnvy to ensure all UtilityFunction cases are caught by FeatureEnvy (I think they are, but better make sure).
* Is FeatureEnvy a good name? In general, both names suggest the function should go elsewhere. Perhaps FeatureEnvy can be renamed to the generic smell type LowCohesion?


chastell:
Re: `UtilityFunction` specs move to `FeatureEnvy` specs – I did a cursory check, potentially worthwhile to save:

* the ‘with a singleton method’ context,
* the ‘does not report literal’ case,
* the ‘recognises an ivar reference within a block’ case,
* the ‘does not report a call to a constant’ case,
* the ‘should count self reference within a dstr’ and ‘…calls to self…’ cases,
* the ‘does not report a method that calls super’ case,
* the ‘should recognise a deep call’ case.

Re: naming – I think this shows the distinction between UtilityFunction and FeatureEnvy: UtilityFunction is stronger and says ‘you should be able to move this without touching its implementation’, while FeatureEnvy says ‘this should be moved elsewhere but you’ll have to edit it to cut the ties to its current location’. FeatureEnvy is LowCohesion, UtilityFunction is NoCohesion. (I’m not saying UtilityFuntion/NoCohesion should stay as a separate smell, just pointing out the difference; I think I’m still for dropping it and making the code simpler.)

As for FeatureEnvy vs LowCohesion – this is totally irrational and subjective, but for me, personally, calling it ‘FeatureEnvy’ makes me more motivated to address it: I read it as ‘this method _has_ feature envy’ vs ‘this method has _low_ cohesion’ – the former seems more categorical, the latter has a fuzzy part which seems to weaken the potential offence.

mvz:
I'm closing this for now as I prefer #418.

PR #417, opened 2015-04-01
Fix programmatic access

mvz:
* Use Aruba for cukes
* Fix programmatic use of reek
* Improve config file finder scenarios

PR #418, opened 2015-04-04
Separate feature envy from utility function

mvz:
This is alternative to #416.

### DIFF
--- a/lib/reek/smells/feature_envy.rb
+++ b/lib/reek/smells/feature_envy.rb
@@ -30,6 +30,9 @@ module Reek
     # Currently +FeatureEnvy+ reports any method that refers to self less
     # often than it refers to (ie. send messages to) some other object.
     #
+    # If the method doesn't reference self at all, +UtilityFunction+ is
+    # reported instead.
+    #
     class FeatureEnvy < SmellDetector
       def self.smell_category
         'LowCohesion'
@@ -42,6 +45,7 @@ module Reek
       # @return [Array<SmellWarning>]
       #
       def examine_context(method_ctx)
+        return [] unless method_ctx.references_self?
         method_ctx.envious_receivers.map do |ref, occurs|
           target = ref.to_s
           SmellWarning.new self,
--- a/config/defaults.reek
+++ b/config/defaults.reek
@@ -109,4 +109,4 @@ UnusedParameters:
 UtilityFunction:
   enabled: true
   exclude: []
-  max_helper_calls: 1
+  max_helper_calls: 0
--- a/features/samples.feature
+++ b/features/samples.feature
@@ -8,7 +8,7 @@ Feature: Basic smell detection
     Then the exit status indicates smells
     And it reports:
     """
-    spec/samples/inline.rb -- 44 warnings:
+    spec/samples/inline.rb -- 43 warnings:
       CompilationError has no descriptive comment (IrresponsibleModule)
       File has no descriptive comment (IrresponsibleModule)
       File#self.write_with_backup has approx 6 statements (TooManyStatements)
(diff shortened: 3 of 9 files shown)


mvz:
I found one case of UtilityFunction in the current reek source (in UtilityFunction, actually) that was not caught by FeatureEnvy:

```ruby
      def depends_on_instance?(exp)
        Reek::Source::ReferenceCollector.new(exp).num_refs_to_self > 0
      end
```

 Hence this alternative: FeatureEnvy is only triggered if there are some references to self, and UtilityFunction is triggered if there are no references to self.

chastell:
> FeatureEnvy is only triggered if there are some references to self, and UtilityFunction is triggered if there are no references to self.

:+1: I think I like this a bit more than ‘UtilityFunction is just a corner case of FeatureEnvy, so let’s drop UtilityFunction altogether’ – it’s true that it is a case of FeatureEnvy, but it might be worthwhile to keep it marked explicitly (e.g., I imagine automatic refactoring tools special-case UtilityFunctions). But I’m still ok with conflating them if that’s the collective preference. :)

mvz:
I prefer this solution as well.

mvz:
Ideally, I would remove the `max_helper_calls` config for UtilityFunction.

mvz:
This one is ready for review.

chastell:
I looked around and it seems `max_helper_calls` and `HELPER_CALLS_LIMIT_KEY` are not tested anywhere. Dropping them would probably be a backward-incompatible change, though. :( (Actually, is lowering `max_helper_calls`’ default from 1 to 0 backward-compatible?)

Otherwise this looks good, thanks!

mvz:
Yeah, we need to make a decision or policy on what we consider compatible and incompatible changes. I'll open an issue about that.

troessner:
I'd love to review this as well, please gimme a day.:)

troessner:
>> Hence this alternative: FeatureEnvy is only triggered if there are some references to self, and UtilityFunction is triggered if there are no references to self.

Good idea!

lib/reek/smells/feature_envy.rb
```diff
@@ -42,6 +42,7 @@ def self.smell_category
       # @return [Array<SmellWarning>]
       #
       def examine_context(method_ctx)
+        return [] unless method_ctx.exp.depends_on_instance?
```
troessner: Can we abstract the guard away into a speaking method so that we make this idea:

>> Hence this alternative: FeatureEnvy is only triggered if there are some references to self, and UtilityFunction is triggered if there are no references to self.

clear to everybody including ourselves in a couple of months?
How about

```Ruby
return [] if no_references_to_self?(method_ctx.exp)
```

Additionally I'd appreciate if we could put 

>> FeatureEnvy is only triggered if there are some references to self, and UtilityFunction is triggered if there are no references to self.

as a introductory comment at the top of the file.

troessner:
Reviewed, looks great to me as well expect for that one tiny comment.:)

troessner:
I'd take over this PR if you don't have the time right now and fix up the last comments - ok for you?

mvz:
I'll handle it tonight.

mvz:
Updated, rebased and squashed.

troessner:
Fantastic, merged!
Could you please update the wiki entries for UtilityFunction and FeatureEnvy as well?

troessner:
I'd release 2.1 with those changes. 
That would also include #417 so that we can close #444 as well.

troessner:
I know we have two other issues marked for 2.1 : https://github.com/troessner/reek/issues?q=milestone%3A2.1+is%3Aissue+is%3Aopen - but let's re-schedule them for 2.2.

mvz:
:+1: on the re-scheduling.

mvz:
This branch actually fixes #53, so that doesn't need re-scheduling. I moved the other one.

PR #419, opened 2015-04-04
Unify recorded references to lvar and lvasgn

mvz:
Reek would record `lvar.foo` and `lvar += 4` as calls to different entities. This PR fixes this.


troessner:
Very nice, merged!

Issue #406, opened 2015-03-08
Make development calls to reek behave more sane


troessner:
Not sure how to phrase this properly.
We have a rather weird  "Use the local reek version when calling reek from the reek working directory. But still require reek to be installed properly" - logic in place.

One example would be #405 
Another example would be this:

In the reek working directory:
>>
  ( $ (master)) pwd
/home/timo/dev/reek

Nothing works when reek is not installed via rubygems:
>>
 ( $ (master)) reek -v
No command 'reek' found, did you mean:
 Command 'ree' from package 'ree' (universe)
reek: command not found

Ok, let's install it:
>>
  ( $ (master)) gem install reek
Fetching: reek-2.0.1.gem (100%)
Successfully installed reek-2.0.1
Parsing documentation for reek-2.0.1
Installing ri documentation for reek-2.0.1
Done installing documentation for reek after 1 seconds
1 gem installed

reek version now works:
>>
  ( $ (master)) reek -v
reek 2.0.1

which reek lists the rubygems path:
>>
  ( $ (master)) which reek
/home/timo/.rvm/gems/ruby-2.1.1/bin/reek

So let's add a raise at the beginning of bin/reek:
>>
  ( $ (master)) sed -i '1s/^/raise /' bin/reek 
{-timo-|-columbia-} => [~/dev/reek]

And there we go:
>>
  ( $ (master)) reek -v
/home/timo/dev/reek/bin/reek:1:in `<top (required)>': unhandled exception
	from /home/timo/.rvm/gems/ruby-2.1.1/bin/reek:23:in `load'
	from /home/timo/.rvm/gems/ruby-2.1.1/bin/reek:23:in `<main>'
	from /home/timo/.rvm/gems/ruby-2.1.1/bin/ruby_executable_hooks:15:in `eval'
	from /home/timo/.rvm/gems/ruby-2.1.1/bin/ruby_executable_hooks:15:in `<main>'

It's great for development that a call to reek uses the local reek version if we're in that directory, but the overall behaviour is pretty weird as you can see above.
Let's discuss how can make this less insane.:)

chastell:
Isn’t this the way RubyGems handles this nowadays? I use ruby-install + chruby + rubygems-bundler, so I have:

```
$ which reek
/home/chastell/.gem/ruby/2.2.1/bin/reek
```

…but it looks like this:

```ruby
#!/usr/bin/env ruby_executable_hooks
#
# This file was generated by RubyGems.
#
[...]

PR #427, opened 2015-04-07
Switch all local requires to require_relative

chastell:
At least a stepping stone towards fixing https://github.com/troessner/reek/issues/406.

### DIFF
--- a/bin/reek
+++ b/bin/reek
@@ -6,6 +6,6 @@
 # Author: Kevin Rutherford
 #
 
-require 'reek/cli/application'
+require_relative '../lib/reek/cli/application'
 
 exit Reek::Cli::Application.new(ARGV).execute
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -2,7 +2,7 @@ require 'English'
 require 'tempfile'
 require 'fileutils'
 require 'open3'
-require 'reek/cli/application'
+require_relative '../../lib/reek/cli/application'
 
 begin
   require 'pry-byebug'
@@ -31,8 +31,7 @@ class ReekWorld
 
   def rake(name, task_def)
     header = <<EOS
-$:.unshift('lib')
-require 'reek/rake/task'
+require_relative 'lib/reek/rake/task'
 
 EOS
     rakefile = Tempfile.new('rake_task', '.')
--- a/lib/reek.rb
+++ b/lib/reek.rb
@@ -1,6 +1,6 @@
 #
 # Reek's core functionality
 #
-require 'reek/version'
-require 'reek/examiner'
-require 'reek/smell_warning'
+require_relative 'reek/version'
+require_relative 'reek/examiner'
+require_relative 'reek/smell_warning'
--- a/lib/reek/cli/application.rb
+++ b/lib/reek/cli/application.rb
@@ -1,6 +1,6 @@
-require 'reek/cli/options'
-require 'reek/cli/reek_command'
-require 'reek/configuration/app_configuration'
+require_relative 'options'
+require_relative 'reek_command'
+require_relative '../configuration/app_configuration'
 
 module Reek
   module Cli
--- a/lib/reek/cli/option_interpreter.rb
+++ b/lib/reek/cli/option_interpreter.rb
@@ -1,8 +1,8 @@
 require 'forwardable'
-require 'reek/cli/input'
-require 'reek/cli/report/report'
-require 'reek/cli/report/formatter'
-require 'reek/cli/report/heading_formatter'
+require_relative 'input'
+require_relative 'report/report'
+require_relative 'report/formatter'
+require_relative 'report/heading_formatter'
 
 module Reek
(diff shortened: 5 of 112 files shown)


troessner:
Reviewed and in favor of this. While this is certainly a lot more verbose it also is a lot clearer, at least in my mind.
@mvz what do you think?

chastell:
_puts his thought leadership cap on_

I think most of the verbosity stems from two things: high coupling (many `require_relative` lines in classes that need a lot of other classes to work) and class hierarchy not matching usage (the longish `../..` parts of the relative paths). Both are things to be gradually fixed, but this diff is their symptom, not a problem with `require_relative` itself.

troessner:
>> I think most of the verbosity stems from two things: high coupling (many require_relative lines in classes that need a lot of other classes to work) and class hierarchy not matching usage/dependencies (the longish ../.. parts of the relative paths). Both are things to be gradually fixed, but this diff is their symptom, not a problem with require_relative itself.

Yes, couldn't agree more!
We already removed a lot of unnecessary requires in the last months but I think there are still quite many left that could be removed or where the class hierarchy should be changed - but this would be a different PR like you already noted as well.

mvz:
I'll have a look at this later in the week or this weekend.

mvz:
Sorry, I didn't manage to take a look at this yet. Hope you can hold on a little longer ...

Just FYI, my current thinking is:
* The Ruby ecosystem should handle which version of a gem gets loaded, by using `require_relative` we're circumventing that.
* This is a workaround for a problem that only occurs if you're developing reek **and** you have rubygems-bundler installed.
* I don't see *any* guarantee in the code of rubygems that going to `../lib/` from inside `bin/` will get you to the `lib` part of the same gem (I think this is the most dangerous part of this change).

chastell:
> The Ruby ecosystem should handle which version of a gem gets loaded, by using `require_relative` we're circumventing that.

I believe that the given `bin/reek` executable should always load its related reek files. RubyGems take care of _which_ `bin/reek` executable gets run (`$GEM_HOME/bin/reek` is a thin wrapper that does exactly this dispatch), but once the given `bin/reek` executable is run it should load _its_ related files, and that’s exactly what `require_relative` does.

The same goes for `require 'reek'` done by a third party (i.e., using reek as a library) – RubyGems take care of which `lib/reek.rb` gets required, but _that_ file should always load its related implementation.

> This is a workaround for a problem that only occurs if you're developing reek **and** you have rubygems-bundler installed.

I think this is unrelated to rubygems-bundler (and I don’t think it’s a workaround) – and, anyway, rubygems-bundler’s behaviour should be exactly the same as pure RubyGems’ with `RUBYGEMS_GEMDEPS` set to `-` (currently RubyGems’ `Gemfile` parsing is lacking, so it blows up in reek’s case, but it’s the same in principle and will work the same as soon as RubyGems’ `Gemfile` parsing gets better).

> I don't see _any_ guarantee in the code of rubygems that going to `../lib/` from inside `bin/` will get you to the `lib` part of the same gem

I don’t understand. RubyGems’ executable wrapper (at `$GEM_HOME/bin/reek`) runs the right `bin/reek` by doing `load Gem.bin_path('reek', 'reek', version)` (once the right `version` is computed) – and it’s pure Ruby that makes sure that a `require_relative` inside _that_ `load`ed `bin/reek` finds _its_ `../lib`. If that wasn’t guaranteed then there would be a Ruby bug (`require_relative` is always relative to `__FILE__`, it doesn’t care about the current directory or the original/top-level `load`/`require` call).

troessner:
I agree with everything @chastell said. 
Especially:

>> I believe that the given bin/reek executable should always load its related reek files. RubyGems take care of which bin/reek executable gets run ($GEM_HOME/bin/reek is a thin wrapper that does exactly this dispatch), but once the given bin/reek executable is run it should load its related files, and that’s exactly what require_relative does.
The same goes for require 'reek' done by a third party (i.e., using reek as a library) – RubyGems take care of which lib/reek.rb gets required, but that file should always load its related implementation.

troessner:
And I think things like that were part of the reason why "require_relative" was invented in the first place. I think using "require" for all the things is as common as it is misplaced.

troessner:
Sooooo, what do we do here? Seems like we're in a dead end here.
Should we take a democratic vote?.;)

mvz:
I'm convinced now that this is in general a good thing, especially because of this:

> RubyGems take care of which `lib/reek.rb` gets required, but that file should always load its related implementation.

So, :+1: on loading related files from `lib/reek.rb` using `require_relative`.

I have one remaining worry, which is that `bin` and `lib` may not be guaranteed by ruygems to be sister directories (at least, looking at the implementation). However, this seems to be generally assumed, so let's assume this as well.

One nice thing about this change is that checking out reek and running straight `bin/reek` will use the correct libraries.

Conclusion: :shipit:

troessner:
Great, merged!

PR #428, opened 2015-04-07
Refactor ConfigurationFileFinder and its spec

chastell:
Just an indea, let me know what you think. Technical changes: `Options#config_file` is now a `Pathname`, `ConfigurationFileFinder` now also parses from `$HOME` up (IMHO harmless – especially when we already go up to the root from the current dir – and makes the code simpler).

### DIFF
--- a/lib/reek/configuration/configuration_file_finder.rb
+++ b/lib/reek/configuration/configuration_file_finder.rb
@@ -14,41 +14,26 @@ module Reek
     # The order in which ConfigurationFileFinder tries to find such a
     # configuration file is exactly like above.
     module ConfigurationFileFinder
-      class << self
-        def find(options)
-          configuration_by_cli(options) ||
-            configuration_in_file_system    ||
-            configuration_in_home_directory
-        end
+      module_function
 
-        private
+      # FIXME: switch to kwargs on upgrade to Ruby 2 and drop `params.fetch` calls:
+      # def find(options: nil, current: Pathname.pwd, home: Pathname.new(Dir.home))
+      def find(params = {})
+        options = params.fetch(:options) { nil                    }
+        current = params.fetch(:current) { Pathname.pwd           }
+        home    = params.fetch(:home)    { Pathname.new(Dir.home) }
+        find_by_cli(options) || find_by_dir(current) || find_by_dir(home)
+      end
 
-        def configuration_by_cli(options)
-          return unless options # Return gracefully allowing calls without app context
-          config_file_option = options.config_file
-          return unless config_file_option
-          path_name = Pathname.new config_file_option
-          raise ArgumentError, "Config file #{path_name} doesn't exist" unless path_name.exist?
-          path_name
-        end
+      def find_by_cli(options)
+        options && options.config_file
+      end
 
-        def configuration_in_file_system
-          detect_or_traverse_up Pathname.pwd
-        end
-
-        def configuration_in_home_directory
-          detect_configuration_in_directory Pathname.new(Dir.home)
-        end
-
-        def detect_or_traverse_up(directory)
-          file = detect_configuration_in_directory(directory)
-          return file unless file.nil?
-          return if directory.root?
-          detect_or_traverse_up directory.parent
-        end
-
-        def detect_configuration_in_directory(directory)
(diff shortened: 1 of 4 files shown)


spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -1,34 +1,58 @@
...
+      found = ConfigurationFileFinder.find(options, current: current)
+      expect(found).to eq(Pathname.new('spec/samples/exceptions.reek'))
     end
 
-    it 'should return the first configuration file it can find in the current directory' do
-      allow(ConfigurationFileFinder).to receive(:detect_or_traverse_up).
-        and_return config_path_same_level
+    it 'returns the file in current dir if options is nil' do
+      current = Pathname.new('spec/samples')
+      found = ConfigurationFileFinder.find(nil, current: current)
```
troessner: Can we move the "options" hash into the second hash? It's really confusing when a methods first parameter is nil. It always makes me think "if I need to pass nil in here just to make it work, maybe this parameter shouldn't be so prominent?".

lib/reek/configuration/configuration_file_finder.rb
```diff
@@ -14,41 +14,25 @@ module Configuration
...
-          detect_or_traverse_up directory.parent
-        end
+      def find_by_cli(options)
+        options && options.config_file
+      end
 
-        def detect_configuration_in_directory(directory)
-          directory.children.select(&:file?).find { |path| path.to_s.end_with?('.reek') }
+      def find_by_dir(start)
+        start.ascend do |dir|
```
troessner: Nice! I didn't know Pathname#ascend until now!

troessner:
Very nice, love it. Just one comment and then let's get this merged.:)

lib/reek/configuration/configuration_file_finder.rb
chastell: Me neither – half of this PR was reading pathname’s docs. :)

Fun fact: `Pathname#ascend` is so old it requires a block; if it returned an `Enumerator` we could do this instead:

```ruby
      def find_by_dir(start)
        start.ascend.find do |dir|
          dir.children.select(&:file?).find { |file| file.to_s.end_with?('.reek') }
        end
      end
```

spec/reek/configuration/configuration_file_finder_spec.rb
chastell: I did that, but I don’t really like it very much – `options` is different from `current` and `home`, as `current` and `home` are dependency injections for testing, while `options` is actually used when calling by working code (the `ConfigurationFileFinder.find(nil)` calls are only in the tests, they wouldn’t happen in working code).

What I think is more problematic is that `options` can be `nil` in the first place (this is so that we can call [`AppConfiguration#initialize_with(nil)`](https://github.com/troessner/reek/blob/7ad005afdbc99129b67b414f8e6a85cd6013f8ec/lib/reek/configuration/app_configuration.rb#L16-L21)). I thought about this a bit and I think eventually we should try to simplify the configuration management – currently there’s `Cli::Options`, `Configuration::AppConfiguration`, `Cli::Input`, the Rake task… We could try coming up with a single place to store the settings and that place could then be either told by the CLI that the file was given or look for it via `ConfigurationFileFinder`.

But I need to think about this a bit more first and that’s for a quite separate PR – let me know whether this one is ok for now. :)

troessner: >>  let me know whether this one is ok for now. :)

Yes, using kwargs fixed all-the-things..:)

>> I thought about this a bit and I think eventually we should try to simplify the configuration management – currently there’s Cli::Options, Configuration::AppConfiguration, Cli::Input, the Rake task… We could try coming up with a single place to store the settings and that place could then be either told by the CLI that the file was given or, if not, look for it via ConfigurationFileFinder (which wouldn’t then have to do the backwards check on the passed CLI options).

Makes sense to me.

troessner:
Looks excellent to me now! Please squash, then I'll merge this little rascal.:)

chastell:
Ready to be merged! [Meanwhile, in Ruby…](https://bugs.ruby-lang.org/issues/11052) ;)

troessner:
>> Meanwhile, in Ruby… ;)

Wow, impressive my friend!

PR #429, opened 2015-04-09
Simplify ConfigurationFileFinder spec a bit

chastell:
A minor simplification of the new `ConfigurationFileFinder` spec from #428. Apologies for not catching that yesterday.


troessner:
Apology not accepted, that was an inexusable mistake!

PR #430, opened 2015-04-09
Clean-up gemspec and bump rubocop and cucumber

chastell:
This bumps rubocop and cucumber and cleans-up the gemspec:
* `executables` are now ‘everything that’s in `bin`’;
* `files` were (mistakenly, I think) computed as an intersection rather than the sum of the `Dir[…]` and `git ls-files` calls – as a result, things like `.rubocop.yml`, `Gemfile`, `spec/samples/masked_by_dotfile/.reek`, etc. were not in the gem;
* `homepage` is now the canonical URL for the wiki;
* `rdoc_options` are copy-pasteable to the commandline;
* `require_paths` are `['lib']` by default, so dropped;
* `rubyforge_project` dropped as RubyForge no longer exists;
* `rubygems_version` [is not supposed to be set](http://guides.rubygems.org/specification-reference/#rubygems_version);
* dependencies are cleaned-up and sorted by name.

### DIFF
--- a/reek.gemspec
+++ b/reek.gemspec
@@ -15,31 +15,24 @@ Gem::Specification.new do |s|
 
   s.license = 'MIT'
   s.email = ['timo.roessner@googlemail.com']
-  s.executables = ['reek']
   s.extra_rdoc_files = ['CHANGELOG', 'License.txt']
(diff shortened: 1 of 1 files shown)


troessner:
Very nice cleanup, merged!

PR #435, opened 2015-04-13
Remove core_extras.

### LINKED ISSUE #433: Remove core_extras monkeypatches
https://github.com/troessner/reek/blob/master/lib/reek/source/core_extras.rb

We're monkeypatching File, IO and String.
We're just extending them, so it' certainly not the worst monkeypatch I have ever seen, but we still shouldn't not monkepatch core classes, unless it's really necessary. Which it isn't in this case.
Furthermore, with us defining the reek API for reek 3 and thus promoting the API usage this is certainly an even bigger no-no to modify core classes.

I'll come up with a pull request that fixes this the next couple of days.

troessner:
Fixes #433 

4 specs are still failing with

>>  [{"smell_category"=>"IrresponsibleModule",
   "smell_type"=>"IrresponsibleModule",
-  "source"=>"$stdin",
+  "source"=>"io",
   "context"=>"Turn",
   "lines"=>[1],
   "message"=>"has no descriptive comment",

Now, I could fix that but before re-adding complexity let's quickly discuss it:

I don't think we should add this behaviour where we differentiate between "io" and "$stdin".
Just "io" is fine.
And even that is probably unnecessary. I have a hard time to think about when this would actually be of ANY value to anybody.
What do you think?


chastell:
I agree that distinguishing between ‘io’ and ‘$stdin’ feels artificial from reek’s point of view – and I’m also not sure why the type of the source (when it’s _not_ a path to a file) was there in the beginning (i.e., ‘io’ vs ‘string’).

I can see the benefits for the user to know whether the smells are for code at a given path or at standard input; I’m not sure what ‘io’/‘string’ would mean to the user, though – so I’d either leave in (from UI perspective) the distinction between a path and ‘stdin’ or just betwen a path and nothing.

troessner:
I'd suggest going with "path" or "IO", nothing else. WDYT?

chastell:
> I'd suggest going with "path" or "IO", nothing else.

:+1:

troessner:
Great, i'll update the PR - @mvz whats your take on this?

troessner:
@chastell I updated the PR - it's fully reviewable now.
I'd appreciate a quick review.:)

chastell:
Looks good, thanks! Merging…

mvz:
:+1:

PR #436, opened 2015-04-13
Remove source rb

troessner:
....because it is unnecessary. And files should be required where they are used, we shouldn't just require all-the-things.:)

Branched off of #435 - let's get this one merged first.


troessner:
Ready for review @mvz @chastell !.:)

mvz:
I agree with this change in principle, but the builds are all failing :-).

chastell:
Thanks for fixing, merged!

troessner:
>> I agree with this change in principle, but the builds are all failing :-).

Whoopsie.:)

PR #439, opened 2015-04-15
Remove the unnecessary "SourceFile" abstraction.

troessner:


troessner:
Ready for review!

chastell:
:+1: – I forgot to grumble how `SourceFile.new` breaks the Liskov Substitution Principle when the core extensions were removed. ;) Will merge as soon as Travis feels green enough.

troessner:
>> I forgot to grumble how SourceFile.new breaks the Liskov Substitution Principle

Good catch!
Also this class was guilty of the smell:

>> Lazy class / Freeloader: a class that does too little.

(from http://en.wikipedia.org/wiki/Code_smell#Common_code_smells)

PR #440, opened 2015-04-15
Properly capitalise acronyms in class names

chastell:
Ok, here’s a potentially controversial one: my brain freezes for a moment (and my eyes bleed a bit) every time I read one of those class names that contain mostly-lowercased acronyms.

The only reason for spelling `ASTNode` as `AstNode` that I can think of would be if there was some class autoloading that couldn’t handle acronyms, but (a) there isn’t one and (b) even Rails 2.3 handles such cases properly (_I don’t want to speak about why I know this in 2015_).

Let me know what you think. I’ll be happy to add backward compatibility (via `AstNode = ASTNode`, etc.) if you think we need to keep it until 3.0.

### DIFF
--- a/bin/reek
+++ b/bin/reek
@@ -8,4 +8,4 @@
 
 require_relative '../lib/reek/cli/application'
 
-exit Reek::Cli::Application.new(ARGV).execute
+exit Reek::CLI::Application.new(ARGV).execute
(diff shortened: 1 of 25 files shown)


mvz:
:shipit:

troessner:
Good idea!
No need for adding complexity by backwards compatibility, lets go for it..;)

troessner:
Merged!

PR #442, opened 2015-04-16
Pick config file that comes first alphabetically

mvz:
This fixes a small problem I had where removing `exceptions.reek` and then adding it again broke the ConfigurationFileFinder specs.

### DIFF
--- a/lib/reek/configuration/configuration_file_finder.rb
+++ b/lib/reek/configuration/configuration_file_finder.rb
@@ -31,7 +31,7 @@ module Reek
 
       def find_by_dir(start)
         start.ascend do |dir|
-          files = dir.children.select(&:file?)
+          files = dir.children.select(&:file?).sort
           found = files.find { |file| file.to_s.end_with?('.reek') }
           return found if found
         end
--- a/spec/reek/configuration/configuration_file_finder_spec.rb
+++ b/spec/reek/configuration/configuration_file_finder_spec.rb
@@ -4,6 +4,8 @@ require 'pathname'
 require 'tmpdir'
 require_relative '../../spec_helper'
 
+include Reek::Configuration
+
 describe ConfigurationFileFinder do
   describe '.find' do
     it 'returns the config_file if it’s set' do


spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -4,6 +4,8 @@
 require 'tmpdir'
 require_relative '../../spec_helper'
 
+include Reek::Configuration
```
troessner: Soooooooooo, we're doing this one last here, right?.:)

lib/reek/configuration/configuration_file_finder.rb
```diff
@@ -31,7 +31,7 @@ def find_by_cli(options)
 
       def find_by_dir(start)
         start.ascend do |dir|
-          files = dir.children.select(&:file?)
+          files = dir.children.select(&:file?).sort
```
troessner: I don't understand how this change relates to:

>> This fixes a small problem I had where removing exceptions.reek and then adding it again broke the ConfigurationFileFinder specs.

Could you please explain that a little more detailed?

spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -4,6 +4,8 @@
...
+
```
chastell: I guess we need this so we can run the spec file individually – can we instead `describe Reek::Configuration::ConfigurationFileFinder` and use `described_class` where `ConfigurationFileFinder` is used in the specs?

(I’m ok with the `include` if that’s your preference, but in general I prefer not to mess with namespacing more than necessary.)

troessner: @chastell -> https://github.com/troessner/reek/issues/443

I believe we are all in favor of that.
That's why I wrote "one last time"..:)

lib/reek/configuration/configuration_file_finder.rb
chastell: I think `Pathname#children` returns the files in filesystem order, which is probably inode-based, so removing `exceptions.reek` and adding it back pushed it to the end of the returned children.

mvz: @chastell that is indeed what I think the problem was.

spec/reek/configuration/configuration_file_finder_spec.rb
mvz: Yep :-). Didn't want to fix #443 in the same PR.

chastell:
Ok, with #443 on the horizon I’m ok with this fix. Merging!

PR #445, opened 2015-04-16
Drop self-includes to lessen namespace pollution

### LINKED ISSUE #443: Test classes by full name
Our specs call `include Reek`, etc., in several places, to avoid having to specify the full names of the classes. This seems silly and hides any problems there might be with the hierarchy (see #438). Also, it means we have to be careful that each spec file includes the modules it needs, or risk being unable to run it on its own. I propose removing all these calls to include, and just use the full name for each class or module.

chastell:
Fixes #443.


mvz:
Looks good to me. :shipit:

troessner:
Awesome!

Issue #438, opened 2015-04-15
Restructure modules


troessner:
I believe our current module / class organisation has some potential for improvements:

Right now the classes and modules structure i'd like to discuss in this issue look like this:

>>
lib/reek
├── core
│   ├── code_context.rb
│   ├── method_context.rb
│   ├── module_context.rb
│   ├── object_refs.rb
│   ├── singleton_method_context.rb
│   ├── smell_configuration.rb
│   ├── smell_repository.rb
│   ├── stop_context.rb
│   ├── tree_walker.rb
│   └── warning_collector.rb
├── source
│   ├── ast_node_class_map.rb
│   ├── ast_node.rb
│   ├── code_comment.rb
│   ├── reference_collector.rb
│   ├── sexp_extensions.rb
│   ├── sexp_formatter.rb
│   ├── sexp_node.rb
│   ├── source_code.rb
│   ├── source_file.rb
│   ├── source_locator.rb
│   ├── source_repository.rb
│   └── tree_dresser.rb
├── spec.rb
└── version.rb
├── examiner.rb
├── smells.rb
├── smell_warning.rb

I suggest the following changes:

- We move "smells.rb" and "smell_warning" to "./smells", "spec.rb" to "./specs", "examiner.rb" to "./core"
- We move everything in "./source" that does not start with "source_*" to "./core". Right now I feel it's highly confusing since the distinction between "./source" and "./core" seems arbitrary, at least partially.
- We move everything in "./source" that starts with "sexp_*" to "lib/reek/sexp"

I believe this would give a clearer picture of what's happening where, it would improve our name system and it would actually put responsibilities together that belong together.

What do you think?

chastell:
I like the new grouping!

I’m not sure about the naming; I (minimally) prefer more functional ones, so maybe ‘analysis` instead of ‘core’? (I tried to come up with something better than ‘source’ and ‘sexp’, but the more I tried the more I realised they are probably good ones.)

I assume `SourceCode` will become `Source::Code`, `SexpNode` will become `Sexp::Node`, etc.?

mvz:
Perhaps we can also simplify the require_relative calls in the specs by making sure `require 'reek'` or `require_relative '../lib/reek'` loads everything?

mvz:
[...]

PR #451, opened 2015-04-18
Reorganize "core" and "source" modules.

troessner:
Implements #438


lib/reek/core/examiner.rb
```diff
@@ -0,0 +1,81 @@
+require_relative '../source/source_repository'
+require_relative 'warning_collector'
+require_relative 'smell_repository'
+require_relative 'tree_walker'
+
+module Reek
+  module Core
```
troessner: Now Examiner is scoped under Core so everything is indented two to the right, hence the diff. Same goes for SmellWarning below.

chastell:
:+1:, merging!

lib/reek/smells/smells.rb
```diff
@@ -1,6 +1,6 @@
 require 'require_all'
 
-require_rel 'smells'
+require_rel '.'
```
mvz: I'm rather surprised that this works.

mvz: Apparently, it doesn't: #453.

lib/reek/spec/spec.rb
```diff
@@ -1,6 +1,6 @@
-require_relative 'spec/should_reek'
-require_relative 'spec/should_reek_of'
-require_relative 'spec/should_reek_only_of'
+require_relative 'should_reek'
+require_relative 'should_reek_of'
+require_relative 'should_reek_only_of'
```
mvz: I would prefer to keep this file at `reek/spec.rb`, for two reasons:

* When a namespace *foo* is introduced, I prefer the idiom of having `foo.rb` and `foo/some_class.rb` over `foo/foo.rb` or `foo/base.rb`.
* Users will now have to `require 'reek/spec/spec'`, which is a breaking change and looks odd.

troessner: Yes, makes sense.

troessner: @mvz please check #454

chastell:
Argh, apologies for the trigger-happy merge. Thanks for fixing the warts in separate PRs.

PR #453, opened 2015-04-18
Drop require_all and side-step circular require

chastell:
We seem to be using the require_all gem in a single place _and_ it’s throwing a `loading in progress, circular require considered harmful` warning.

This replaces the require_all dependency with three lines that do the same and moves `smells.rb` a level higher.


mvz:
This problem is due to the recent moving of `smells.rb` into `smells`. We could just move that back.

chastell:
That’s a good point, sorry for missing that one. I do think it should live in `lib/reek/smells.rb`, so I moved it back there.

I still don’t think it makes sense to depend on require_all for just this one use (and, in general, I much prefer explicit requires), especially when it can be replaced by three lines, but I’m open to arguments for its use. :)

mvz:
require_all was introduced in #388. I think @troessner agreed that it just looks nicer that way.

lib/reek/smells.rb
```diff
@@ -0,0 +1,11 @@
+Dir["#{File.dirname(__FILE__)}/smells/*.rb"].each do |file|
```
mvz: Oh god, now I'm seeing globbing issues everywhere!

mvz:
I couldn't resist: https://github.com/jarmo/require_all/issues/9.

lib/reek/smells.rb
chastell: :D

chastell:
I switched this to a non-globbing approach – but if you’re both sold on require_all then I’ll bring it back, of course. :)

troessner:
I like it,  and i believe @mvz is not die-hard set on require_all, merged!

mvz:
:+1:

PR #455, opened 2015-04-18
ConfigurationFileFinder: test magic paths

chastell:
<del>The fact that we couldn’t match the `.reek` config file via `*.reek` globbing was growing on me, so I dove into docs and found a way.</del> It turns out the globbing doesn’t work 100% of the time in Rubinius, so this PR only adds the below:

I also fondly remember the wise words of Michael Grünewald:

> Hobby: creating bug tickets, waiting for people to find out about my home directory – which happens to contain spaces, stars and newlines.

– so I added a relevant spec.

### DIFF
--- a/spec/reek/configuration/configuration_file_finder_spec.rb
+++ b/spec/reek/configuration/configuration_file_finder_spec.rb
@@ -1,5 +1,6 @@
 # encoding: UTF-8
 
+require 'fileutils'
 require 'pathname'
 require 'tmpdir'
 require_relative '../../spec_helper'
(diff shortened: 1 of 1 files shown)


chastell:
Well did I just find a Rubinius b̶u̶g̶ i̶n̶c̶o̶m̶p̶a̶t̶i̶b̶i̶l̶i̶t̶y̶ improvement over MRI…

chastell:
Ok, so Rubinius 2.5.2 _does_ pass the magic paths spec when we do the old way around, so

```ruby
          files = dir.children.select(&:file?).sort
          found = files.find { |file| file.to_s.end_with?('.reek') }
```

rather than

```ruby
          found = Pathname.glob("#{dir}/{.reek,*.reek}").sort.find(&:file?)
```

– so for now I reverted my globbing simplification and will revisit this if/when Rubinius acts the same as MRI and JRuby.

mvz:
Interesting. Globbing in general seems to be a problematic thing. See for example https://github.com/rubinius/rubinius/issues/3149 and https://bugs.ruby-lang.org/issues/10700.

Also, interpolating `"ma\ngic-d*r"` into a globbing pattern yields a pattern that will match `"ma\ngic-d-how-did-we-get-here-r". I don't think there's any ready made escape function we can apply to fix this.

My conclusion is that as soon as we make `dir` part of the globbing pattern, we're not guaranteed to get only files in `dir`, and we're better off with the current implementation.

chastell:
> Also, interpolating `"ma\ngic-d*r"` into a globbing pattern yields a pattern that will match `"ma\ngic-d-how-did-we-get-here-r"`

Oh, good point. I removed the `FIXME` with the globbing approach – I still think it’s worth to have this test, though.

mvz:
> I still think it’s worth to have this test, though.

Definitely.

PR #457, opened 2015-04-19
Add link to overcommit under "Tools"

lencioni:
overcommit is a fully configurable and extendable Git commit hook
manager that includes out-of-the-box support for running reek as a Git
pre-commit hook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,8 @@ If you don't feel like getting your hands dirty with code there are still other 
 * [Colorful output for `reek`](https://github.com/joenas/preek)
   (also with [Guard::Preek](https://github.com/joenas/guard-preek))
 * [Atom plugin for `reek`](https://atom.io/packages/linter-reek)
+* [overcommit, a Git commit hook manager with support for
+  `reek`](https://github.com/brigade/overcommit)
 
 ### Find out more:
 


mvz:
:shipit:

troessner:
I'm hesitant to merge this. I did check out https://github.com/brigade/overcommit and reek is nowhere even mentioned there. So this feels highly arbitrary. If we'd merge this, we could just list all somehow CI or code quality related tools under "Tools".
Can you please at least mention in your docs?

mvz:
It looks like the Reek support is in https://github.com/brigade/overcommit/blob/2bef9f19772d9d95928a03d023d6f43451c19077/lib/overcommit/hook/pre_commit/reek.rb

I do agree that the way it is now, the lack of mentions of Reek in overcommit would be quite confusing for Reek users checking it out from this link.

In fact, I think any overcommit user would be happy to have a nice list of existing hooks in the README rather than having to dig through the hooks directory.

lencioni:
These are good suggestions, thanks! I opened up a dialog about this at https://github.com/brigade/overcommit/issues/189

troessner:
`reek` is now explicitly listed in the README, happy to merge this one, thanks @lencioni !

troessner:
Oh, and I didnt know overcommit which looks pretty awesome..:)

lencioni:
Thanks!

PR #459, opened 2015-04-19
Clean up RuboCop config

chastell:
This cleans up the RuboCop config as per current develop state: lowers acceptable offence parameters closer to RuboCop’s defaults and removes offences the codebase no longer commits.

Let me know if you believe we should allow any of the removed offences in the future.

(Apologies for the slew of clean-up PRs, but I’m trying to get the codebase as friendly as possible before ROSSConf Vienna next weekend – and RuboCop config is one of the things I want to talk about.)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,7 +6,7 @@ AllCops:
 
 # FIXME: Lower the ABC size by fixing the biggest offenders
 Metrics/AbcSize:
-  Max: 25
+  Max: 18
 
 # FIXME: Make the class shorter
 Metrics/ClassLength:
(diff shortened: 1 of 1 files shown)


mvz:
Seems reasonable. There are a couple of settings I tend to keep in my projects, like the GuardClause one, but we can always put them back if there's occasion to do so. It's nice to see that we've managed to lower things like the maximum method length.

chastell:
> There are a couple of settings I tend to keep in my projects, like the GuardClause one, but we can always put them back if there's occasion to do so.

I agree – I like to keep my `.rubocop.yml` (and `.reek`…) files as short as the current codebase allows, even if I suspect some exceptions may come back in the future.

.rubocop.yml
```diff
@@ -15,7 +15,7 @@ Metrics/ClassLength:
 
 # FIXME: Lower the method length by fixing the biggest offenders
 Metrics/MethodLength:
-  Max: 46
```
troessner: Still too much, but epic progress!

chastell: I’d like to show this as a potential low-hanging fruit for ROSSConf Vienna. :)

PR #462, opened 2015-04-20
Refactor SmellWarning#yaml_hash and lower max ABC size by 2

chastell:
This refactors `SmellWarning#yaml_hash`, our biggest ABC metric offender.

This is an example refactoring I’d like to show for tackling RuboCop offences (and, down the road, reek smells) in a codebase.


troessner:
Nice!

PR #464, opened 2015-04-21
NilCheck: drop the dot from the message

chastell:
For some strange reason NilCheck’s message ends with a dot (I checked and no other message does this). This PR drop the dot and makes the message match the others.


troessner:
You are on an unparalleled merge spree..:D

chastell:
I’m preparing my slides for ROSSConf Vienna and it just stuck out in the example output. :smile: 

(But I also want to use this PR as an example of ‘we welcome even the smallest fixes if they’re valid’.)

PR #466, opened 2015-04-24
Refactor sources

troessner:


troessner:
Commits are not squashed since all commits are independent refactoring commits.

lib/reek/cli/utilities.rb
```diff
@@ -0,0 +1,12 @@
+module Reek
+  module CLI
+    # CLI utilities
+    module Utilities
+      def self.require_silently
+        old_verbose, $VERBOSE = $VERBOSE, nil
+        yield if block_given?
+        $VERBOSE = old_verbose
+      end
```
chastell: I’m probably way over-zealous in this case, but can we put the reset line in an `ensure` block?

```ruby
      def self.require_silently
        old_verbose, $VERBOSE = $VERBOSE, nil
        yield if block_given?
      ensure
        $VERBOSE = old_verbose
      end
```

(We _could_ try making this thread-safe, but that would be overdoing it even for my tastes.) ;)

lib/reek/cli/utilities.rb
```diff
@@ -0,0 +1,12 @@
+module Reek
+  module CLI
+    # CLI utilities
+    module Utilities
```
mvz: Looks my per-commit notes don't show up here, so I'll quote myself:

> This name suggests this will be a grab bag of vaguely related methods. How about calling it SilentRequirer or something like that and give it only the one purpose?

lib/reek/cli/utilities.rb
```diff
@@ -0,0 +1,12 @@
...
+      def self.require_silently
+        old_verbose, $VERBOSE = $VERBOSE, nil
+        yield if block_given?
```
mvz: I would just require a block and drop the `block_given?` part: If there's no block this method is pretty useless :laughing:.

lib/reek/cli/utilities.rb
mvz: Oh, actually the `require` part is incidental. How about `Reek::Silencer.silently { require 'x' }`?

lib/reek/cli/utilities.rb
troessner: I agree, but AFAIS using block_given? even in such cases is pretty idiomatic.
I can remove it if you insist but from personal point of view I'd leave it in...

troessner:
@mvz @chastell addressed both your comments!

lib/reek/cli/utilities.rb
mvz: I won't insist :smile:.

mvz:
:shipit:

troessner:
@mvz then can you please go ahead and merge?..;)

mvz:
Sure, was just giving @chastell a chance to chip in ... too late now :-D.

chastell:
:+1: – sorry I wasn’t quciker, but I was actually SPEAKING ABOUT `reek` AT A CONFERENCE at the time. ;)

mvz:
Awesome!

Best. Excuse. Ever.

How did it go?

chastell:
We’ll see in the afternoon how many people join, but it seems to have gone quite well. :)

troessner:
Haha, nice!

PR #467, opened 2015-04-25
Extract method to reduce ABC size

andyw8:
RuboCop identified that `PrimaDonnaMethod#examine_constent` had a high
ABC size.

### DIFF
--- a/lib/reek/smells/prima_donna_method.rb
+++ b/lib/reek/smells/prima_donna_method.rb
@@ -28,18 +28,25 @@ module Reek
 
       def examine_context(ctx)
         ctx.node_instance_methods.map do |method_sexp|
-          next unless method_sexp.ends_with_bang?
-
-          version_without_bang = ctx.node_instance_methods.find do |sexp_item|
-            sexp_item.name.to_s == method_sexp.name_without_bang
-          end
-          next if version_without_bang
-
-          SmellWarning.new self,
-                           context: ctx.full_name,
-                           lines: [ctx.exp.line],
-                           message: "has prima donna method `#{method_sexp.name}`"
+          check_for_smells(method_sexp, ctx)
         end.compact
+      end
+
+      private
+
+      def check_for_smells(method_sexp, ctx)
+        return unless method_sexp.ends_with_bang?
+
+        version_without_bang = ctx.node_instance_methods.find do |sexp_item|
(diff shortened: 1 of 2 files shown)


lib/reek/smells/prima_donna_method.rb
```diff
@@ -28,18 +28,25 @@ def self.contexts # :nodoc:
...
-          next if version_without_bang
+      private
 
-          SmellWarning.new self,
-                           context: ctx.full_name,
-                           lines: [ctx.exp.line],
-                           message: "has prima donna method `#{method_sexp.name}`"
-        end.compact
+      def check_for_smells(method_sexp, ctx)
+        return unless method_sexp.ends_with_bang?
```
chastell: I’m on the fence whether this check should be a part of the method (in general, I’d say yes) or – in this particular case – remove it from this method and move the check to the caller:

```ruby
check_for_smells(method_sexp, ctx) unless method_sexp.ends_with_bang?
```

Thoughts?

chastell: Or, better yet:

```ruby
ctx.node_instance_methods.select(&:ends_with_bang?).map do |method_sexp|
```

chastell:
Looks good to me other than the minor discussion on the guard call.

Thanks so much for this PR – this is exactly the kind of contribution we hoped for! :)

lib/reek/smells/prima_donna_method.rb
mvz: If the method name is `check_for_smells`, I would keep this check as part of it. There are alternative ways to organize the code of course; Probably a method `corresponding_non_bang_method_exists?` could be extracted. Not sure if that gains much, though.

chastell: Agreed! Although it’s a private method (rightly so!), not to be callable from outside – and in the case of private methods it makes sense to make them useful / simple for the local callers. But it would indeed be best to rename this method in such a case.

But I’m ok with leaving the guard check (and the name). If there are no more nitpicks I’d love to merge this.

mvz: Yes, please merge.

chastell:
Our first ROSSConf-triggered PR is merged, thanks so much!

PR #468, opened 2015-04-25
fix Gem::ConflictError when executing bin/reek directly

beanieboi:
after cloning and executing `bin/reek` we faced the following error.

```
$ bin/reek .
/Users/ben/.rbenv/versions/2.2.1/lib/ruby/2.2.0/rubygems/specification.rb:2112:in `raise_if_conflicts': Unable to activate unparser-0.2.2, because parser-2.2.2.1 conflicts with parser (~> 2.2.0.2) (Gem::ConflictError)
```

require `bundler/setup` to load the gem versions in Gemfile.lock

worked on this with @phillipp

#rossconf :dancer:

### DIFF
--- a/bin/reek
+++ b/bin/reek
@@ -6,6 +6,8 @@
 # Author: Kevin Rutherford
 #
 
+require 'bundler/setup'
+
 require_relative '../lib/reek/cli/application'
 
 exit Reek::CLI::Application.new(ARGV).execute


mvz:
Interesting. Since `bin/reek` is what gets installed when the user does `gem install reek`, I'm not too sure about this solution. Perhaps our dependency on unparser needs to be updated?

chastell:
This is a dependency-resolving issue (we are ok with parser 2.2.2, but [unparser isn’t yet](https://github.com/mbj/unparser/pull/43)) – using `bundle exec` fixes it:

`bundle exec bin/reek --version`

as well as telling RubyGems to do the `Gemfile.lock`-based resoltuion:

`RUBYGEMS_GEMDEPS=- bin/reek --version`

I’d rather not couple Reek to Bundler this much – but if we do, we should make Bundler a runtime dependency as well.

chastell:
(Another solution would be to have `RUBYGEMS_GEMDEPS=-` exported in your environment all the time, of course.)

mvz:
If you're going to that, you might as well instruct people to run `bundle exec reek` :laughing:.

chastell:
Oh definitely, we’re quite set on `bundle exec` all the things in the contributions guidelines (do propose fixes if we’re not obvious enough).

RubyGems’ `Gemfile.lock` support is still quite experimental and lacking. :)

troessner:
I just freshly installed 2.2.1 via rvm, did a bundle install followed by a fresh clone and 

>> bin/reek .

works for me like a charm. Does this have to do something with rbenv / rvm?

phillipp:
To reproduce you would probably have to install another, older version of one of the gems from the Gemfile‎.

Von: Timo Rößner
Gesendet: Samstag, 25. April 2015 21:30
An: troessner/reek
Antwort an: troessner/reek
Cc: Phillipp Röll
Betreff: Re: [reek] fix Gem::ConflictError when executing bin/reek directly (#468)



I just freshly installed 2.2.1 via rvm, did a bundle install followed by a fresh clone and

bin/reek .

works for me like a charm. Does this have to do something with rbenv / rvm?

troessner:
Ah, I clearly misunderstood the problem..;)
In that case what @mvz said, bundler all the way.

chastell:
Ok, I’m closing this one then. Thanks for the debugging and the discussion!

mvz:
I've found another way to fix this: Rubygems will correctly handle these dependencies as long as unparsers is required first.

mvz:
Fixed in #486.

chastell:
Ah, this makes perfect sense, as unparser has more conservative dependencies.

(This is a bug in RubyGems’ dependency resolution that will get fixed eventually, and this particular instance will solve itself once parser 2.2.2.2 is released with the https://github.com/whitequark/parser/issues/192 fix and unparser 0.2.3 is released with https://github.com/mbj/unparser/pull/43 included. But if there’s a simple fix on our side that can side-step this then sure, let’s do it.) :)

mvz:
I'm not sure this will ever get fixed in RubyGems, since I think it lacks the information to fix it: How would it know we're going to need unparser later?

chastell:
By parsing `Gemfile.lock` if it exists – this is what RubyGems wants to do eventually (and tries to do now), but still adorably fails now-and-then.

mvz:
Hm ... we'll see where that goes.

PR #469, opened 2015-04-25
Remove ruby management related files

bruno-:
Some people might prefer rbenv so `.ruby-gemset` file only creates noise for them.

The person actually using the `.ruby-gemset` file should probably gitignore it globally from `~/.gitignore`.


chastell:
I switched from RVM to ruby-install + chruby, so I’m for this change, but I’d love to have @troessner’s / @mvz’s input before merging.

mvz:
I'm all for removing this file, since I don't use it.

andyw8:
+1

It should be up to the RVM user to decide which gemset to use when working on a project.

troessner:
Thx!

PR #471, opened 2015-04-25
Do not enable colorization if stdout is not a TTY

sometimesfood:
If the standard output is not a TTY -- for example when piping output to a pager such as less or redirecting it into a file -- colorization should not be enabled by default.

### DIFF
--- a/lib/reek/cli/options.rb
+++ b/lib/reek/cli/options.rb
@@ -10,10 +10,11 @@ module Reek
     # Parses the command line
     #
     class Options
-      def initialize(argv)
+      def initialize(argv = [])
         @argv    = argv
         @parser  = OptionParser.new
-        @options = OpenStruct.new(colored: true, smells_to_detect: [])
+        @options = OpenStruct.new(colored: color_support?,
+                                  smells_to_detect: [])
         set_up_parser
       end
 
@@ -25,6 +26,10 @@ module Reek
       end
 
       private
+
+      def color_support?
+        $stdout.tty?
+      end
 
       def set_up_parser
         set_banner
--- a/spec/reek/cli/options_spec.rb
+++ b/spec/reek/cli/options_spec.rb
@@ -1,0 +1,19 @@
+require_relative '../../spec_helper'
+
+require_relative '../../../lib/reek/cli/options'
+
+describe Reek::CLI::Options do
+  describe '#initialize' do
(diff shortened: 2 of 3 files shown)


chastell:
I thought we [already do this](https://github.com/troessner/reek/blob/ff168d5867546e1f539429d17274cd583b6bdf09/lib/reek/cli/input.rb#L21-L23) – is there a case where it doesn’t work as expected?

mvz:
@chastell that's stdin, not stdout.

This will probably be a tricky feature to get exactly right :-).

chastell:
Argh, right, sorry. I do think we should turn this off for non-TTYs at least, so @sometimesfood is working on a fix and a test for it (and a temporary RuboCop whitelist for the `CLI::Options` class length…). ;)

sometimesfood:
Added a simple spec and the temporary whitelist. Let me know if this is too much brute force.

lib/reek/cli/options.rb
```diff
@@ -9,11 +9,13 @@ module CLI
     #
     # Parses the command line
     #
+    # rubocop:disable ClassLength
```
chastell: Can you move it to the [`FIXME` exclusion of `.rubocop.yml`](https://github.com/troessner/reek/blob/17c5fbce761727768c3d35c1bc12aa7abfa66c32/.rubocop.yml#L7-L10), please? We do want to fix it eventually. :)

spec/reek/cli/options_spec.rb
```diff
@@ -0,0 +1,21 @@
+require_relative '../../spec_helper'
+
+require_relative '../../../lib/reek/cli/options'
+
+describe Reek::CLI::Options do
+  describe '#initialize' do
+    let(:argv) { [] }
```
chastell: Can you either inline this `let` or make `CLI::Options.new`’s parameter optional? I understand the need for DRY-ness, but I think this makes the individual tests a bit less readable (you have to jump to the `let` just to see that it’s an empty `Array`).

sometimesfood:
@chastell Sure thing, just changed my branch according to your comments.

chastell:
:+1: – I checked the `io/console` stdlib and it seems there’s no easy way to check the colour support of the terminal (and I doubt we want to shell out just to do `tput colors`). We might find a way with one of the gems for colouring console output, but for now I’ll merge this once Travis gets greener.

PR #478, opened 2015-04-25
Improve sexp formatter output and remove DuplicateMethodCalls.

clemenshelm:


chastell:
Thanks! Looks good – can you squash your commits into one?

lib/reek/core/smell_repository.rb
```diff
@@ -24,7 +24,8 @@ def initialize(source_description = nil, smell_types = self.class.smell_types)
       end
 
       def configure(klass, config)
-        @detectors[klass].configure_with(config) if @detectors[klass]
+        detector = @detectors[klass]
+        detector.configure_with(config) if detector
```
troessner: Actually I'm not so sure if this guard makes sense (I know that you didnt introduce it in this PR, but it makes still sense to discuss it now).
I mean, the expectation here is that we have a valid detector. And if we don't get a valid one, we should rather raise an ArgumentError here.
WDYT? cc @chastell @mvz

lib/reek/core/ast_node_class_map.rb
```diff
@@ -12,12 +12,11 @@ def initialize
       end
 
       def klass_for(type)
-        @klass_map[type] ||=
-          begin
-            klass = Class.new(ASTNode)
-            klass.send :include, extension_map[type] if extension_map[type]
-            klass.send :include, Sexp::SexpNode
-          end
+        @klass_map[type] ||= Class.new(ASTNode).tap do |klass|
```
troessner: Very nice!

lib/reek/core/smell_repository.rb
chastell: Agreed!

lib/reek/core/ast_node_class_map.rb
clemenshelm: Thanks!

lib/reek/core/smell_repository.rb
clemenshelm: I did it!

clemenshelm:
@chastell squashed!

lib/reek/core/smell_repository.rb
```diff
@@ -24,7 +24,9 @@ def initialize(source_description = nil, smell_types = self.class.smell_types)
       end
 
       def configure(klass, config)
-        @detectors[klass].configure_with(config) if @detectors[klass]
+        detector = @detectors[klass]
+        raise ArgumentError, "Cannot configure smell #{klass}" unless detector
```
troessner: "Cannot configure smell ..." is a little confusing for the user since it tells them basically "I have a valid detector but for whatever reasons I can't configure it" when we actually want to say "you give us a smell detector we don't know".

How about changing this to

>> "Cannot configure nknown smell detector #{klass}"

troessner:
Very nice! Just one last comment, then please re-squash and then I'll merge this one.:)

lib/reek/core/smell_repository.rb
mvz: > `"Cannot configure nknown smell detector #{klass}"`

That still sounds confusing to me. How about `"Unknown smell type #{klass} found in confuguration"`?

lib/reek/sexp/sexp_formatter.rb
```diff
@@ -12,10 +12,10 @@ class SexpFormatter
       def self.format(sexp)
         return sexp.to_s unless sexp.is_a? AST::Node
         lines = Unparser.unparse(sexp).split "\n"
-        if lines.length > 1
-          "#{lines.first} ... #{lines.last}"
+        if lines.length < 3
+          lines.join('; ')
```
mvz: That's a very nice solution!

lib/reek/sexp/sexp_formatter.rb
```diff
@@ -12,10 +12,10 @@ class SexpFormatter
...
         else
-          lines.first
+          [lines.first, lines.last].join ' ... '
```
mvz: However, I find this more confusing than  the original.

Hm .. I don't know what the cleanest way to write this whole method would be.

Perhaps it's cleanest to clearly show the three cases: one line, two lines, more lines?

```ruby
case lines.length
when 1: lines.first
when 2: lines.join('; ')
else: [lines.first, lines.last].join ' ... '
end
```

troessner: :+1:

lib/reek/core/smell_repository.rb
troessner: :+1:

lib/reek/sexp/sexp_formatter.rb
```diff
@@ -8,14 +8,15 @@ module Sexp
...
         return sexp.to_s unless sexp.is_a? AST::Node
         lines = Unparser.unparse(sexp).split "\n"
-        if lines.length > 1
-          "#{lines.first} ... #{lines.last}"
-        else
-          lines.first
+        case lines.length
+        when 1 then lines.first
+        when 2 then lines.join('; ')
+        else [lines.first, lines.last].join(' … ')
```
clemenshelm: I tried `lines.values_at(0, -1).join(' … ')` to avoid the duplicate method call, but I think it reads much worse. That's why I ignored the duplicate method call.

spec/reek/sexp/sexp_formatter_spec.rb
```diff
@@ -23,7 +23,13 @@
                 s(:send, nil, :qux)))
       result = described_class.format ast
 
-      expect(result).to eq 'if foo ... end'
+      expect(result).to eq 'if foo … end'
```
clemenshelm: I favor the typographically correct `…` character.

mvz:
Can you remove the old commit messages from the squashed commit please?

clemenshelm:
Done!

troessner:
Saucy!

PR #479, opened 2015-04-25
Added instruction to use reek in ruby code

bibstha:
The file api.md inside docs folder contain instruction to run reek inside a ruby app.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1,0 +1,48 @@
+# Using `reek` inside your Ruby application
+
+`reek` can be used inside another Ruby project.
+
+```bash
+gem install reek
+```
+
+You can use reek inside your Ruby file `check_dirty.rb`
+
+```ruby
+require 'reek'
+require 'reek/source/source_code'
+require 'reek/cli/report/report'
+require 'reek/core/examiner'
+
+source =<<END
+class Dirty
+  # This method smells of :reek:NestedIterators but ignores them
+  def awful(x, y, offset = 0, log = false)
+    puts @screen.title
+    @screen = widgets.map { |w| w.each { |key| key += 3 * x } }
+    puts @screen.contents
+    fail
+  end
+end
+END
+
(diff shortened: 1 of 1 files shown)


chastell:
Thanks!

@mvz / @troessner: I’m merging this as a good starting point for the discussion about what public API should we support for `reek` 3.0 (as in, if this hooks too deep then maybe we should have some good methods for this closer to the top of the namespace).

troessner:
@bibstha thanks a lot!

@mvz @chastell 
By convention, this would have belonged in the wiki, not in the docs folder. But I still highly welcome this because I think we should discuss if  it wouldnt make sense to move everything in the wiki as well. 
IMO this is what we should do. The separation between code and "some" documentation and most of the other documentation in the wiki doesn't make much sense to me.

So I'd suggest moving everything we have in the wiki into the docs folder and then disable the github wiki. WDYT?

chastell:
> I'd suggest moving everything we have in the wiki into the docs folder and then disable the github wiki

I would prefer to have the docs in the code for two reasons:
* as much as I like GitHub, I’m not a huge fan of being tied to it more than necessary/useful,
* smell examples in the repo could be checked with [Ataru](https://github.com/CodePadawans/ataru) whether they’ll still smelly (and whether the refactoring example removes the smell).

Given how wide the links to our wiki are distributed we should probably turn the wiki pages into stubs pointing to the new locations in the docs folder – _but_ this could be automated, as [the GitHub wiki itself is a Git repo](https://help.github.com/articles/adding-and-editing-wiki-pages-locally/) in its own right (this might seem to invalidate the second point above, but I think it would be simpler to have the docs in the same repo as the rest of `reek` and just have a single CI job).

troessner:
>> this might seem to invalidate the second point above, but I think it would be simpler to have the docs in the same repo as the rest of reek and just have a single CI job).

Yes, exactly.

PR #481, opened 2015-04-25
don’t track private methods in the Attributes smell

beanieboi:
this fixes the TODO in the AttributeSmell.

if this gets merged, the wiki needs to get updates too.
https://github.com/troessner/reek/wiki/Attribute

### DIFF
--- a/lib/reek/smells/attribute.rb
+++ b/lib/reek/smells/attribute.rb
@@ -12,10 +12,19 @@ module Reek
     # +attr_reader+, +attr_writer+ and +attr_accessor+ -- including those
     # that are private.
     #
-    # TODO: Eliminate private attributes
     # TODO: Catch attributes declared "by hand"
     #
     class Attribute < SmellDetector
+      ATTR_DEFN_METHODS = [:attr, :attr_reader, :attr_writer, :attr_accessor]
+      VISIBILITY_MODIFIERS = [:private, :public, :protected]
+
+      def initialize(*args)
+        @visiblity_tracker = {}
+        @visiblity_mode = :public
+        @result = Set.new
+        super
+      end
+
       def self.contexts # :nodoc:
         [:class, :module]
       end
@@ -42,14 +51,33 @@ module Reek
       private
 
       def attributes_in(module_ctx)
-        result = Set.new
-        attr_defn_methods = [:attr, :attr_reader, :attr_writer, :attr_accessor]
         module_ctx.local_nodes(:send) do |call_node|
-          if attr_defn_methods.include?(call_node.method_name)
-            call_node.arg_names.each { |arg| result << [arg, call_node.line] }
+          if visibility_modifier?(call_node)
+            track_visibility(call_node)
+          elsif ATTR_DEFN_METHODS.include?(call_node.method_name)
+            call_node.arg_names.each do |arg|
+              @visiblity_tracker[arg] = @visiblity_mode
+              @result << [arg, call_node.line]
+            end
           end
         end
-        result
+        @result.select { |args| recorded_public_methods.include?(args[0]) }
+      end
+
+      def visibility_modifier?(call_node)
+        VISIBILITY_MODIFIERS.include?(call_node.method_name)
+      end
+
+      def track_visibility(call_node)
+        if call_node.arg_names.any?
+          call_node.arg_names.each { |arg| @visiblity_tracker[arg] = call_node.method_name }
+        else
+          @visiblity_mode = call_node.method_name
+        end
+      end
+
+      def recorded_public_methods
+        @visiblity_tracker.select { |_, visbility| visbility == :public }
       end
     end
   end
--- a/spec/reek/smells/attribute_spec.rb
+++ b/spec/reek/smells/attribute_spec.rb
@@ -17,6 +17,47 @@ describe Reek::Smells::Attribute do
       ctx = Reek::Core::CodeContext.new(nil, Reek::Source::SourceCode.from(src).syntax_tree)
       expect(@detector.examine_context(ctx)).to be_empty
     end
+
+    shared_examples_for 'no attribute found' do
+      before :each do
+        ctx = Reek::Core::CodeContext.new(nil, Reek::Source::SourceCode.from(@src).syntax_tree)
+        @smells = @detector.examine_context(ctx)
+      end
+
+      it 'records no attribute' do
+        expect(@smells.length).to eq(0)
+      end
+    end
+
+    context 'declared in a class' do
+      before :each do
+        @src = "class Fred
+          attr :super_private, :super_private2
+          private :super_private, :super_private2
+          private
+          attr :super_thing
+          public
+          attr :super_thing2
+          private
+          attr :super_thing2
+        end"
+      end
+
+      it_should_behave_like 'no attribute found'
+    end
+
+    context 'declared in a module' do
+      before :each do
+        @src = "module Fred
+          attr :super_private, :super_private2
+          private :super_private, :super_private2
+          private
+          attr :super_thing
+        end"
+      end
+
+      it_should_behave_like 'no attribute found'
+    end
   end
 
   context 'with one attribute' do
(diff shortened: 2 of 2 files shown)


lib/reek/smells/attribute.rb
```diff
@@ -12,10 +12,18 @@ module Smells
...
     # TODO: Catch attributes declared "by hand"
     #
     class Attribute < SmellDetector
+      ATTR_DEFN_METHODS = [:attr, :attr_reader, :attr_writer, :attr_accessor]
+
+      def initialize(*args)
+        @visiblity_tracker = {}
+        @visiblity_mode = :public
+        @result = Set.new
+        super(args)
```
chastell: We try to follow Liskov substitution principle, so we prefer a bare `super` where possible (as `super`with explicit arguments suggests there might be a difference in constructor signatures of the current class and its superclass, which, in turn, might suggest a Liskov substitution principle violation).

lib/reek/smells/attribute.rb
```diff
@@ -42,14 +50,37 @@ def examine_context(ctx)
...
+              @result << [arg, call_node.line]
+            end
           end
         end
-        result
+        @result.select { |args| public_methods_from_tracker.include?(args[0]) }
+      end
+
+      def visibility_modifier?(call_node)
+        call_node.method_name == :private || call_node.method_name == :public
```
chastell: What about `protected`?

spec/reek/smells/attribute_spec.rb
```diff
@@ -114,5 +155,82 @@
...
+
+      it 'reports the declaration line number' do
+        expect(@smells[0].lines).to eq([1])
+      end
+
+      it 'reports the correct smell class' do
+        expect(@smells[0].smell_category).to eq(described_class.smell_category)
+      end
+
+      it 'reports the context fq name' do
```
andyw8: 'fq' is 'fully qualified'? I think would be better to avoid the abbreviation.

beanieboi: not an excuse, but i just copied it from the existing examples.
should we fix all occurrences?

lib/reek/smells/attribute.rb
beanieboi: added.

lib/reek/smells/attribute.rb
```diff
@@ -42,14 +52,37 @@ def examine_context(ctx)
...
+
+      def track_visibility(call_node)
+        if modifier_with_attributes?(call_node)
+          call_node.arg_names.each { |arg| @visiblity_tracker[arg] = call_node.method_name }
+        else
+          @visiblity_mode = call_node.method_name
+        end
+      end
+
+      def public_methods_from_tracker
```
troessner: You abstract the "visibility mode" you're interested away by using VISIBILITY_MODE. But in this method name you kind of make this abstraction useless since it explicitly says "public". Maybe this abstraction is over the top? If you just care about not "fat fingering" "public" and still have an abstraction, how about renaming VISIBILITY_MODE TO PUBLIC_VISIBILITY?
Also I'd prefer to drop the "from_tracker" part in the method name since this is an unncessary detail.
So I'd just call this "public_methods".
WDYT?

lib/reek/smells/attribute.rb
```diff
@@ -42,14 +52,37 @@ def examine_context(ctx)
...
+        @visiblity_tracker.select { |_, visbility| visbility == VISIBLITY_MODE }
+      end
+
+      def modifier_with_attributes?(call_node)
+        call_node.children[0].nil? && call_node.children.size > 2
```
troessner: The 0 and the 2 are kind of magic numbers. Can you please TomDoc this method and also explain the 0 and the 2 in this Tomdoc?

lib/reek/smells/attribute.rb
```diff
@@ -12,10 +12,20 @@ module Smells
...
     # TODO: Catch attributes declared "by hand"
     #
     class Attribute < SmellDetector
+      ATTR_DEFN_METHODS = [:attr, :attr_reader, :attr_writer, :attr_accessor]
+      VISIBILITY_MODIFIERS = [:private, :public, :protected]
+      VISIBLITY_MODE = :public
+
+      def initialize(*args)
+        @visiblity_tracker = {}
+        @visiblity_mode = :public
```
troessner: You have a constant VISIBILITY_MODE and an ivar "@visibility_mode". That's highly confusing. Can't we get rid of the ivar?

lib/reek/smells/attribute.rb
andyw8: Ruby already defines `Object#public_methods` :laughing:

troessner: Ah, excellent point!.:D
How about `recorded_public_methods`?

beanieboi: whoops, yeah. i kind of got excited to kill the magic strings. i really like `recorded_public_methods`

lib/reek/smells/attribute.rb
beanieboi: i just realized that i can write this with `call_node.arg_names.any?` in a much cleaner way. this also eliminates the magic numbers.

beanieboi: eliminated the abstraction completely and call `call_node.arg_names.any?` directly.

beanieboi:
squashed and rebased

spec/reek/smells/attribute_spec.rb
```diff
@@ -17,6 +17,47 @@
...
     end
+
+    shared_examples_for 'no attribute found' do
+      before :each do
+        ctx = Reek::Core::CodeContext.new(nil, Reek::Source::SourceCode.from(@src).syntax_tree)
+        @smells = @detector.examine_context(ctx)
+      end
+
+      it 'records no attribute' do
+        expect(@smells.length).to eq(0)
```
mvz: Some of our specs use this old method of checking for smells, but we also have `.should_not reek_of(:Attribute)`, which reduces this whole shared example block to just one line, which can then be inlined.

mvz: OTOH, since this was how the whole spec file was before, we can also leave this change for later.

beanieboi: yes, the attribute spec is pretty weird. i thought about refactoring the spec. i digged into the rabbit hole and stopped half way through. i'm much in favour of redoing the whole spec with the new technique you described.

i'll look into preparing another PR for the spec.

spec/reek/smells/attribute_spec.rb
```diff
@@ -17,6 +17,47 @@
...
+      end
+    end
+
+    context 'declared in a class' do
```
mvz: Can you make these context a bit more specific? You now have to read the actual value assigned to `@src` to understand what's being tested.

mvz:
Can you remove the old commit messages from the squashed commit please?

beanieboi:
@mvz fixed the commit message

troessner:
Awesome pull request @beanieboi !
Merging this one.
@beanieboi are you planning on doing more contributions? If yes, I'd make you a regular contributor to make it easier for you.

beanieboi:
@troessner wow nice!
how should we handle the update of the GitHub Wiki, should it reflect the `develop` branch? currently it says that the Attributes Smell also reports private `attr_*` definitions, which changed with this PR.

i'll have some ideas for further PR's, but it's fine to not be a contributor, let's see how this goes :)

beanieboi:
also big kudos to @phillipp who worked with me on this during #ROSSConf :)

phillipp:
great to see this merged :-) i had a great time working with @beanieboi 

when we reviewed the attributes smell we thought that a smell that complains only about `attr_writer` and `attr_accessor` could be a smell that should be enabled by default, because it allows changing object state from outside. @chastell said that a good way would probably a seperate smell. any thoughts on this?

chastell:
> @chastell said that a good way would probably a seperate smell.

Did I? (I honestly don’t remember.) :smile: I opened #491 to discuss this.

PR #489, opened 2015-04-28
Improve YARD output

mvz:
Several improvements to the YARD output:

* Include all files in `lib/`; we can exclude any classes we don't want to see by marking them `@private` or `private_constant` (I didn't know about the latter option).
* Update Examiner to use the full power of YARD.
* Include extra documentation files.
* Fix up local markdown links.

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,10 +1,7 @@
 --no-private
 --readme README.md
-lib/reek.rb
-lib/reek/*.rb
-lib/reek/smells/*.rb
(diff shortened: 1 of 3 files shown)


mvz:
See https://github.com/lsegal/yard/issues/863 for the problem of local links.

troessner:
Looking good! I took the liberty of squashing your commits to get this merged right away, hope you don't mind.:)
Merged!

PR #492, opened 2015-05-01
Tomdoc SourceCode#syntax_tree.

troessner:


mvz:
I think we need to discuss Tomdoc vs YARD, because I'm running into some issues making YARD play nice with Tomdoc in my pull request #479. More details later.

troessner:
We can agree on yarddoc everywhere. For me Tomdoc just looks nicer..;)

mvz:
In that case, I vote for yarddoc, since tomdoc seems to be unmaintained (both the spec and the yard plugin).

The problem I'm running into with the plugin is mainly that Tomdoc doesn't respect the `@private` annotation for classes, so adding the plugin makes several classes appear that are only used internally by the smell detectors, even if the `--no-private` option is passed.

troessner:
Fixed that in #496 - closing this one.
I'll update all other Tomdoc usages when I stumble across them or is this kind of a blocker for you now?

mvz:
> is this kind of a blocker for you now?

No, I think if we gradually migrate to YARD only it'll be fine.

Issue #497, opened 2015-05-02
rspec enhancements


beanieboi:
i would like to change 2 rspec configuration options.

### 1. disable monkeypatching
`config.disable_monkey_patching!`
disables the old should syntax by patching object. this enforces the new expect-syntax which does not rely on monkeypatching.

### 2. verify partial doubles
from the docs

> Prevents you from mocking or stubbing a method that does not exist on a real object. This is generally recommended, and will default to `true` in RSpec 4.

it's really nice that we don't have any violations of this in the current code. :dancer: 
i find such things hard to spot in code reviews, enabling this config options would fail the build if you mock/stub methods that don't exist. especially when you remove code during refactoring, this option fails the spec when the removed method is still mocked/stubbed.

```
config.mock_with :rspec do |mocks|
  mocks.verify_partial_doubles = true
end
```

to enable 2 we don't have to do anything, for option 1 we need to update some specs.

WDYT?

troessner:
:+1: for both suggestions, sounds like 2 clear improvements.
Lets go for that.

mvz:
:+1:

chastell:
:+1: for both – being a [Bogus](https://github.com/psyho/bogus) contributor I welcome RSpec’s verified doubles very much. If stubbing/mocking turns out to mask some issues we can always go with Bogus’s contract tests. ;)

beanieboi:
@chastell is it ok for you to close this issue and reopen if something is not working with the new rspec configuration?

troessner:
>> @chastell is it ok for you to close this issue and reopen if something is not working with the new rspec configuration?

I'm not @chastell but I don't think he'll mind.:)
Closing this one.

PR #498, opened 2015-05-02
rspec enhancements

beanieboi:
as proposed in https://github.com/troessner/reek/issues/497 this is the PR for the enhancements.

i intentionally used 2 commits here because these are 2 different features to me. it also makes sense to me, when you look at reverting commits, if you want to revert the monkeypatching commit you don't want to necessarily revert the partial double check.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -52,4 +52,8 @@ RSpec.configure do |config|
   config.run_all_when_everything_filtered = true
   config.include FactoryGirl::Syntax::Methods
   config.include Helpers
+
+  config.mock_with :rspec do |mocks|
+    mocks.verify_partial_doubles = true
+  end
 end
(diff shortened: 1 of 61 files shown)


spec/spec_helper.rb
```diff
@@ -52,4 +52,13 @@ def ast(*args)
   config.run_all_when_everything_filtered = true
   config.include FactoryGirl::Syntax::Methods
   config.include Helpers
+
+  config.disable_monkey_patching!
+
+  config.mock_with :rspec do |mocks|
+    # Prevents you from mocking or stubbing a method that does not exist on
```
andyw8: I think the comment isn't needed – the reader can refer to the RSpec documentation if they're not familiar with this method.

beanieboi:
removed the comment from `spec_helper.rb`

mvz:
Looks good to me :shipit:

chastell:
:+1:

PR #487, opened 2015-04-27
Refactor sexp.

troessner:
Major changes:
- Stopped the proliferation of methods from SexpNode across our codebase. Now just "each_node" and "find_nodes" is used for traversing the AST outside of the SexpNode. 
- Tomdoc'ed all-the-things
- Make SexpFormatter the only authority for formatting Sexp's.

PR #500, opened 2015-05-03
Refactor ReferenceCollector call counting

chastell:
I was travelling when https://github.com/troessner/reek/pull/487 was reviewed, so this fixes my minor nitpick for that. :)

### DIFF
--- a/lib/reek/core/reference_collector.rb
+++ b/lib/reek/core/reference_collector.rb
@@ -12,14 +12,19 @@ module Reek
       end
 
       def num_refs_to_self
-        result = 0
-        [:self, :zsuper, :ivar, :ivasgn].each do |node_type|
-          @ast.each_node(node_type, STOP_NODES) { result += 1 }
+        (explicit_self_calls + implicit_self_calls).size
+      end
+
+      private
+
+      def explicit_self_calls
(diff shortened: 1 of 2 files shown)


lib/reek/core/reference_collector.rb
```diff
@@ -12,14 +12,19 @@ def initialize(ast)
...
-        end
-        @ast.each_node(:send, STOP_NODES) do |call|
-          result += 1 unless call.receiver
+        (explicit_self_calls + implicit_self_calls).size
+      end
+
+      private
+
+      def explicit_self_calls
+        [:self, :zsuper, :ivar, :ivasgn].flat_map do |node_type|
```
troessner: Why not just "map" here? Using "flat_map" here is confusing since it implies that it actually is important to use "flat_map" and not just "map".

chastell: It is important, otherwise this will return an `Array` of `Arrays` of calls (and won’t work); note that `each_node` returns an `Array`.

troessner: (facepalm) yes, of course. Nevermind.;)
Let me re-review this bad boy tomorrow.

troessner:
Very nice, merged!

Issue #495, opened 2015-05-02
Cleanup Attributes Smell spec


beanieboi:
Hey,

while working on #481, we (@phillipp and i) came to the conclusion that  https://github.com/troessner/reek/blob/develop/spec/reek/smells/attribute_spec.rb is not a nice spec to work with.

i have some goals in my head for the new spec, besides a cleanup:
- remove the usage of shared examples and use custom matcher to preserve the correct stack trace for failed expectations.
- make it easier to add new specs without duplicating a lot of spec code

/cc #491

chastell:
:+1: – more readable (even at a cost of slight duplication) specs will be most welcome.

troessner:
Merged #501 so I'm closing this one.

PR #501, opened 2015-05-03
Cleanup attribute smell spec

beanieboi:
as proposed in #495 here is a cleanup of the attributes spec.

i used the `ShouldReekOf` custom matcher to avoid shared specs and duplication.

is there some sort of code coverage in place?
i had to enable the Attribute smell by default to get it working with `ShouldReekOf`, is there a better way to do this?

### DIFF
--- a/spec/reek/smells/attribute_spec.rb
+++ b/spec/reek/smells/attribute_spec.rb
@@ -1,37 +1,40 @@
 require_relative '../../spec_helper'
 require_relative '../../../lib/reek/smells/attribute'
-require_relative '../../../lib/reek/core/module_context'
 require_relative 'smell_detector_shared'
 
 RSpec.describe Reek::Smells::Attribute do
-  before :each do
+  let(:config) do
+    {
+      Attribute: { Reek::Core::SmellConfiguration::ENABLED_KEY => true }
+    }
+  end
+
+  before(:each) do
     @source_name = 'dummy_source'
     @detector = build(:smell_detector, smell_type: :Attribute, source: @source_name)
+  end
+
+  around(:each) do |example|
+    with_test_config(config) do
+      example.run
+    end
   end
 
   it_should_behave_like 'SmellDetector'
 
   context 'with no attributes' do
-    it 'records nothing in the module' do
-      src = 'module Fred; end'
-      ctx = Reek::Core::CodeContext.new(nil, Reek::Source::SourceCode.from(src).syntax_tree)
-      expect(@detector.examine_context(ctx)).to be_empty
+    it 'records nothing' do
+      expect('
+        class Klass
+        end
+      ').to_not reek_of(:Attribute)
     end
+  end
 
-    shared_examples_for 'no attribute found' do
-      before :each do
-        ctx = Reek::Core::CodeContext.new(nil, Reek::Source::SourceCode.from(@src).syntax_tree)
-        @smells = @detector.examine_context(ctx)
-      end
-
-      it 'records no attribute' do
-        expect(@smells.length).to eq(0)
-      end
-    end
-
-    context 'declared in a class' do
-      before :each do
-        @src = "class Fred
+  context 'with attributes' do
+    it 'records nothing' do
+      expect('
+        class Klass
           attr :super_private, :super_private2
           private :super_private, :super_private2
           private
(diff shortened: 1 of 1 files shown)


lib/reek/smells/attribute.rb
```diff
@@ -30,7 +30,7 @@ def self.contexts # :nodoc:
       end
 
       def self.default_config
-        super.merge(Core::SmellConfiguration::ENABLED_KEY => false)
+        super.merge(Core::SmellConfiguration::ENABLED_KEY => true)
```
troessner: You can enable smells just for a spec run with our handy "with_test_config" helper like here: https://github.com/troessner/reek/blob/develop/spec/reek/smells/nested_iterators_spec.rb#L138

spec/reek/smells/attribute_spec.rb
```diff
@@ -12,26 +11,18 @@
...
-        ctx = Reek::Core::CodeContext.new(nil, Reek::Source::SourceCode.from(@src).syntax_tree)
-        @smells = @detector.examine_context(ctx)
-      end
-
-      it 'records no attribute' do
-        expect(@smells.length).to eq(0)
-      end
+    it 'records nothing' do
+      expect('
+        class Fred
```
troessner: This "Fred" thing in our specs is driving me cracy. Can we please switch that to "Klass"?

beanieboi: sure :)

lib/reek/smells/attribute.rb
mvz: I'm thinking we sould enhance the reek_of matcher to automatically enable the smell to be deteced.

troessner: Excellent idea, but that probably warrants a separate pull request.:)

mvz: Oh, definitely :-).

beanieboi:
rbx-2 is now the only failing platform, i'll look into this later.
is there any other feedback?

spec/reek/smells/attribute_spec.rb
```diff
@@ -40,210 +43,80 @@
...
-      end
-
-      it_should_behave_like 'no attribute found'
+        end
+      ').to_not reek_of(:Attribute)
     end
 
-    context 'declared in a module' do
-      before :each do
-        @src = "module Fred
```
mvz: I don't see the module examples anymore in the refactored version.

beanieboi: i'm not sure how we should handle this. the only code that is related to class/module is `self.contexts`. the actual implementation does not rely on classes or modules at all. we need a test to make sure that the smell is running in the context of a module or a class, is there a better way in reek to do this? 

before it felt like there is something special happening for classes and modules. that gave me the impression to duplicate all tests cases for modules. which in fact doesn't really make sense.

mvz: Just a single test to make sure this detector also checks modules should be enough. Just enough to justify the implementation of `self.contexts`.

beanieboi: added

beanieboi:
rbx-2 fixed itself.

spec/reek/smells/attribute_spec.rb
```diff
@@ -40,210 +43,88 @@
...
-      before :each do
-        @src = "class Fred; private; attr :#{@attr_name}
+    it "doesn't record protected attributes" do
+      src = '
+        class Klass
+          private
+          attr :my_attr
           public
-          attr :#{@attr_name}
+          attr :my_attr
```
mvz: I just noticed: the attribute `:my_attr` is defined twice here.

spec/reek/smells/attribute_spec.rb
```diff
@@ -40,210 +43,88 @@
...
+      src = '
+        class Klass
+          private
+          attr :my_attr
           public
-          attr :#{@attr_name}
-        end"
-      end
-      it_should_behave_like 'two attributes found'
+          attr :my_attr
```
mvz: Also here, the attribute `:my_attr` is defined twice.

mvz:
Just two minor nits left :-).

chastell:
(FWIW: rbx didn’t as much fix itself as I reran it; it seems we’re triggering rbx to blow up on every/most first build runs, but rerunning it makes it fix itself…)

beanieboi:
sorry for the delay. i fix the last to minors.

Issue #491, opened 2015-04-28
Attribute: make writers reek by default?


chastell:
Continuing the discussion from https://github.com/troessner/reek/pull/481:

The Attribute smell is disabled by default – and (in its current state) I’d say rightly so, as there are many cases where reading attributes makes a lot of sense (e.g., value objects).

@beanieboi and @phillipp rightly suggested that attribute writers (and, by extension, accessors) are more smelly than attribute readers, and we might want to consider making them reek by default.

Thus, two questions:
* should attribute writers reek by default?
* if so, should it be a separate smell or a configuration on the existing Attribute smell?

My tentative answers are ‘yes’ and ‘a configuration on the existing Attribute smell’, but they’re _very_ tentative. :)

phillipp:
I think we got to the conclusion that a seperate smell would be easier to seperate from the original smell  (like in tools and configs) and is a valid edge case of the original small.

troessner:
>>
Thus, two questions:
should attribute writers reek by default?
if so, should it be a separate smell or a configuration on the existing Attribute smell?
My tentative answers are ‘yes’ and ‘a configuration on the existing Attribute smell’, but they’re very tentative. :)

I'm with @chastell on this one.:)

chastell:
> we got to the conclusion that a seperate smell would be easier to seperate from the original smell (like in tools and configs) and is a valid edge case of the original small

Ah, right, and I mentioned the UtilityFunction vs FeatureEnvy case. :)

I thought about this a bit more and think that the ‘this has nothing to do with this class whatsoever (and can be moved _anywhere_ just by cutting and pasting)’ vs ‘this has less to do with the current class than this other object (so maybe should be moved _there_, but will require edits)’ difference between UtilityFunction and FeatureEnvy is much bigger than the reader vs writer distinction of attribute access. If I were looking for a smell that handles attribute readers or writers I’d look for a common one.

I wonder how many codebases have _quite valid_ attribute writers… I’d vote for ‘let’s make it configurable, enable flagging attribute writers by default and find out’. :) @phillipp / @beanieboi: wanna give it a try?

beanieboi:
looks good to me, i would suggest something like the following for the config:

```
Attribute:
  enabled: false
  exclude: []
  ignore: :getter | :setter
```

what do you think?

chastell:
I think I prefer ‘reader’ and ‘writer’ (as these are the words used in `attr_*` methods).

[...]

PR #506, opened 2015-05-08
Attributes smell no readers

beanieboi:
this is the PR for #491 

this PR:
- enables the Attribute smell by default
- does not report the usage of `attr` and `attr_reader` as a smell.


spec/reek/smells/attribute_spec.rb
```diff
@@ -47,26 +35,10 @@ class Klass
...
-        end
-      ').to reek_of(:Attribute, name: 'my_attr')
-    end
-
-    it 'records reader attribute' do
-      expect('
-        class Klass
-          attr_reader :my_attr
+          attr_writer :my_attr
         end
```
troessner: I'd prefer to have one test that checks the inverse: That we don't consider attr_reader smelly.:)

troessner:
Awesome!
Just one comment and then we're good to AFAIS.:) (and travis ofc)

beanieboi:
rbx-2 failed again on Travis :(

@chastell do you why this happens? i couldn't find a Issue for rbx-2, i'll open one.

mvz:
I triggered a rebuild on rbx-2 but it failed again.

spec/reek/smells/attribute_spec.rb
mvz: And a test for `attr` as well, of course.

This bit is actually a little tricky, because `attr` can define setters as well, if used in the following way:

```ruby
attr :y, true
```

lib/reek/smells/attribute.rb
```diff
@@ -30,7 +30,7 @@ def self.contexts # :nodoc:
       end
 
       def self.default_config
-        super.merge(Core::SmellConfiguration::ENABLED_KEY => false)
+        super.merge(Core::SmellConfiguration::ENABLED_KEY => true)
```
mvz: Since enabled is the default, this whole method can go.

features/samples.feature
```diff
@@ -61,7 +61,15 @@ Feature: Basic smell detection
     Then the exit status indicates smells
     And it reports:
     """
-    optparse.rb -- 111 warnings:
+    optparse.rb -- 128 warnings:
+      OptionParser declares the attribute banner (Attribute)
```
mvz: The description is a little odd if we only flag writers as smelly. How about changing the message, e.g., to `declares the writeable attribute`.

lib/reek/smells/attribute.rb
```diff
@@ -15,7 +15,7 @@ module Smells
     # TODO: Catch attributes declared "by hand"
     #
     class Attribute < SmellDetector
-      ATTR_DEFN_METHODS = [:attr, :attr_reader, :attr_writer, :attr_accessor]
+      ATTR_DEFN_METHODS = [:attr_writer, :attr_accessor]
```
mvz: Please also change the description in the comment above this line.

chastell:
> rbx-2 failed again on Travis :(

I restarted the build and it passed this time. I have no clue why this happens to `reek` builds; if you want to open an issue with Rubinius then by all means please do. :)

features/samples.feature
beanieboi: +1

spec/reek/smells/attribute_spec.rb
beanieboi: oh! i didn't know that. i'll look into this.

mvz: I didn't know about it either until two days ago, actually :smile:. I looked it up because I was unsure about what `attr` does.

mvz:
I've opened in issue with Rubinius: https://github.com/rubinius/rubinius/issues/3394

spec/reek/smells/attribute_spec.rb
beanieboi: just a little update, i'm digging into this..

the `true` is not reported as `arg_name`
```
[1] pry(#<Reek::Smells::Attribute>)> call_node
=> (send nil :attr
  (sym :my_attr)
  (true))
[2] pry(#<Reek::Smells::Attribute>)> call_node.arg_names
=> [:my_attr, nil]
```
i could hack something like `call_node[2] == "attr" && call_node[4] == :something`

but on the other hand, i have no idea how to check for `(true)`

```
[6] pry(#<Reek::Smells::Attribute>)> call_node[4]
=> (true)
[7] pry(#<Reek::Smells::Attribute>)> call_node[4].class
=> #<Class:0x007fc7f2b12df0>
```

mvz: It looks like `SendNode#arg_names` is a bit of a hack that assumes all arguments are symbols. You can use `SendNode#args` instead, which will give you the real arguments.

beanieboi: sorry for the delay, couldn't find time in the last week. it's on my list.

troessner:
@beanieboi are you still on this one?

beanieboi:
somehow, yes :)

beanieboi:
just an update from my side. i'm currently in the process of moving between continents, i won't be able to dedicate time to this in the next weeks. it's fine for me to close this for now, i have a local copy of the changes and i will continue. it's also fine if somebody wants to pick of from here.
i'm sorry for the massive delay :(

troessner:
@beanieboi can you please rebase and point this to the master branch?
We then could take over if you don't have time.

beanieboi:
I will.

beanieboi:
how can i point this to master? seems like i can't edit the target branch on GitHub

PR #513, opened 2015-05-10
Only attempt to require actual ruby files

mvz:
This stops reek from attempting to require any `.swp` files that might be present.

### DIFF
--- a/lib/reek/smells.rb
+++ b/lib/reek/smells.rb
@@ -1,6 +1,7 @@
 require 'pathname'
 
 (Pathname.new(__FILE__).dirname + 'smells').children.each do |path|
+  next unless path.extname == '.rb'
   require_relative "smells/#{path.basename('.rb')}"
 end
 


troessner:
Hmmm. Quite frankly, i dont think we should do this. Why would we litter our code base with Editor specific logic?

chastell:
Let me try to broker a solution here: we should just use (a long list of) explicit requires and stop trying to be <del>lazy</del> clever. `<grins, ducks and rolls />`

troessner:
@mvz you can configure the directory vim is using for swap files if i am not mistaken e.g. to /tmp. Wouldnt that solve your problem already?

andyw8:
Would it perhaps be more idiomatic to write this as:

```ruby
Dir.glob("smells/*.rb").each do |path|
  require_relative path
end
```

chastell:
@andyw8: I think we had the `*.rb` globbing in the past, but decided against that as [there are cases where Ruby globbing is problematic](https://github.com/troessner/reek/pull/455) (this wouldn’t be a problem in this case here as long as we keep sensible smell file names, though).

mvz:
@andyw8 I think we basically tried everything already :laughing:. Your version breaks because the current directory isn't `lib/reek`. Injecting the file's location in there could break because Unix allows `*` in file names, and it's not just the smell file names, but any directory name above it that we need to worry about. It's a mess.

> @mvz you can configure the directory vim is using for swap files if i am not mistaken e.g. to /tmp. Wouldnt that solve your problem already?

I'll give that a try.

chastell:
FWIW, in Vim that’s

```VimL
set directory=/tmp//
set undodir=/tmp//
```

#unsolicitedvimadvice

mvz:
@chastell thanks, I dit not know about `undodir`.

mvz:
@troessner your tip worked.

Closing this. We can bicker about how (not) to be clever in `smells.rb` some other time.

mvz:
I'm afraid `set directory` has a nasty side-effect: You cannot edit two files with the same name in different directories at the same time without vim complaining about the swap file existing.

I'm with @chastell on just listing all smells.

chastell:
> You cannot edit two files with the same name in different directories at the same time without vim complaining about the swap file existing.

If `set directory=…` ends with two slashes (like in my example above), then the tempfile will be named using the full path to the original (with `/` replaced by `%`).

> I'm with @chastell on just listing all smells.

`\o/` Just… need… to… convince… one… more… person… ;)

mvz:
> If set directory=… ends with two slashes (like in my example above), then the tempfile will be named using the full path to the original

Ah! That's the secret :-). Thanks, it works. Awesome.

troessner:
>> Just… need… to… convince… one… more… person… ;)

Ok, let's do that and then have a comment above it that says "Smell: Load all those smells automagically".

Just kidding.:)

Let's just require the smells explicitly like you proposed and get rid of this problem and discussion..:)

chastell:
`\o/` I’m on it.

PR #518, opened 2015-05-23
Feature/refactor core dir

### LINKED ISSUE #517: Refactoring /core part 1
I'd propose the following changes to /core:

* Rename StopContext to RootContext
* Move all *_contexts to core/contexts/
* Move SmellConfiguration and SmellRepository to lib/reek/smells

WDYT? Any objections?

troessner:
Partly fixes #517 and in particular implements:

1.) Rename StopContext to RootContext. The file's top comment already said:

>>     # A context wrapper representing the root of an abstract syntax tree.

2.) Move all *_contexts to core/contexts/

Let me summarize my arguments again (more or less c&p'ing from the issue): 
"contexts" should reside on the top level, right next to /core. /core is already kind of a grab-bag for quite some unrelated things already. Not to mention that "core" doesnt mean anything. We could also name it "stuff" and the name would just be as expressive.
So let's rather start working on removing most if not everything under /core into more appropriate directories.


troessner:
@mvz and @chastell this PR and #519 / #520 are blocking me from further refactorings - any chance you can look into those today?

chastell:
I’m ok with this one and @mvz didn’t raise concerns, so merging.

PR #519, opened 2015-05-23
Move smell-related classes out of /core to /smells.

troessner:
Fixes the last part of #517 and move SmellConfiguration and SmellRepository to lib/reek/smells.

Both classes are related to smells, but not to anything else in /core AFAIS.

### DIFF
--- a/lib/reek/smells/attribute.rb
+++ b/lib/reek/smells/attribute.rb
@@ -1,5 +1,4 @@
 require_relative 'smell_detector'
-require_relative '../core/smell_configuration'
 
 module Reek
   module Smells
@@ -31,7 +30,7 @@ module Reek
       end
 
       def self.default_config
-        super.merge(Core::SmellConfiguration::ENABLED_KEY => false)
+        super.merge(SmellConfiguration::ENABLED_KEY => false)
       end
 
       #
--- a/lib/reek/smells/long_parameter_list.rb
+++ b/lib/reek/smells/long_parameter_list.rb
@@ -1,5 +1,4 @@
 require_relative 'smell_detector'
-require_relative '../core/smell_configuration'
 
 module Reek
   module Smells
@@ -21,7 +20,7 @@ module Reek
       def self.default_config
         super.merge(
           MAX_ALLOWED_PARAMS_KEY => DEFAULT_MAX_ALLOWED_PARAMS,
-          Core::SmellConfiguration::OVERRIDES_KEY => {
+          SmellConfiguration::OVERRIDES_KEY => {
             'initialize' => { MAX_ALLOWED_PARAMS_KEY => 5 }
           }
         )
--- a/lib/reek/smells/smell_repository.rb
+++ b/lib/reek/smells/smell_repository.rb
@@ -1,9 +1,9 @@
 require_relative '../smells'
-require_relative '../smells/smell_detector'
+require_relative 'smell_detector'
 require_relative '../configuration/app_configuration'
 
 module Reek
-  module Core
+  module Smells
     #
     # Contains all the existing smells and exposes operations on them.
     #
--- a/lib/reek/core/examiner.rb
+++ b/lib/reek/core/examiner.rb
@@ -5,8 +5,8 @@
 #
 # See also https://github.com/troessner/reek/pull/468
 require_relative 'tree_walker'
-require_relative 'smell_repository'
 require_relative '../cli/warning_collector'
+require_relative '../smells/smell_repository'
 require_relative '../source/source_repository'
 
 module Reek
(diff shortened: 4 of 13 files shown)


chastell:
I’m ok with this one, but I remember @mvz having concerns about moving `SmellRepository` under `Smells`, so I’m deferring this merge to him.

mvz:
I was worried that the moving of SmellConfiguration might unnecessarily break our external API, undefined as it is, but it seems I was mistaken. I suppose it's mostly Examiner that is used externally.

lib/reek/smells/attribute.rb
```diff
@@ -1,5 +1,5 @@
 require_relative 'smell_detector'
-require_relative '../core/smell_configuration'
+require_relative '../smells/smell_configuration'
```
mvz: You can now remove the `../smells/` bit from the require here and in other files under `smells/`.

lib/reek/smells/attribute.rb
```diff
@@ -31,7 +31,7 @@ def self.contexts # :nodoc:
       end
 
       def self.default_config
-        super.merge(Core::SmellConfiguration::ENABLED_KEY => false)
+        super.merge(Smells::SmellConfiguration::ENABLED_KEY => false)
```
mvz: Similarly, the `Smells` could be removed here.

lib/reek/smells/smell_repository.rb
```diff
@@ -1,9 +1,9 @@
 require_relative '../smells'
```
mvz: Perhaps a next step would be to move the contents of `smells.rb` into here.

mvz:
Good to go as far as I'm concerned after the requires and namespacing have been simplified.

lib/reek/smells/attribute.rb
troessner: Turns out this require wasn't even needed.:)

lib/reek/smells/long_parameter_list.rb
```diff
@@ -1,5 +1,5 @@
 require_relative 'smell_detector'
-require_relative '../core/smell_configuration'
+require_relative '../smells/smell_configuration'
```
troessner: Same here, was not needed at all

lib/reek/smells/smell_repository.rb
troessner: Good idea! Then you'd just have to require SmellRepository...

troessner: Let's make that a separate PR to keep things simple.

troessner:
@mvz addressed all your comments. Merge?.:)

troessner:
Seems like a random travis failure, restarting...

lib/reek/smells/smell_repository.rb
mvz: Agreed.

mvz:
Looks like there's more going on in that build failure:

* There's a circular require due to `smells.rb` requiring all files in `smells/`.
* One sample cuke is failing due to different smells being detected. That one seems odd.

troessner:
>> There's a circular require due to smells.rb requiring all files in smells/.

I wonder why this works locally but not on travis. I cant see anything specific in what travis does and in the end it just calls "bundle exec rake" which works for me across multiple ruby versions.

Regarding the problem:

I'd suggest moving all smell types into smells/smell_types, require those in smell_repository and get rid of smells.rb. This also feels like a clean architecture to me.

What do you think about that?

mvz:
> I'd suggest moving all smell types into smells/smell_types, require those in smell_repository and get rid of smells.rb. This also feels like a clean architecture to me.

Perhaps we can start by moving `smell-repository.rb` up to root level? It seems the simpler solution for now. Otherwise, I think `smells/` will end up containing just `smell_repository.rb` and `smell_types/`?

The Travis failures are caused by newer versions of `parser` by the way. I don't know yet what the relevant change is, but it all works fine with `2.2.2.2` but breaks with `2.2.2.5`. :frowning:

mvz:
The problem in parser `2.2.2.5` is probably due to https://github.com/whitequark/parser/issues/194.

troessner:
Then let's pin parser to 2.2.2.2 for now until this issue is fixed. 
Objections?
And afterwards we should really discuss pinning all the gems we use to the version that is currently running. This is not the first time we're having this issue.

mvz:
:+1: on pinning parser.

chastell:
> Then let's pin parser to 2.2.2.2 for now until this issue is fixed.

This makes sense. Note: from a cursory look at the issue it seems like a _feature_ of parser (different comment association), so we’ll hav to _fix_ it on our side…

> And afterwards we should really discuss pinning all the gems we use to the version that is currently running. This is not the first time we're having this issue.

Hm, I think semver exists for a reason and most dependencies actually follow it – I don’t think parser does, though. We should definitely pin parser.

mvz:
@troessner can you rebase to see if the tests pass now?

troessner:
They're passing!

PR #520, opened 2015-05-23
Move WarningCollector to /cli.

troessner:
WarningCollector had nothing to do with anything else in /core but rather with /cli.


chastell:
Good point (IMHO), merging. :)

PR #523, opened 2015-05-27
Pin parser to 2.2.2.2 to fix our CI build.

troessner:


features/samples.feature
```diff
@@ -9,9 +9,9 @@ Feature: Basic smell detection
     Then the exit status indicates smells
     And it reports:
     """
-    inline.rb -- 43 warnings:
+    inline.rb -- 44 warnings:
       CompilationError has no descriptive comment (IrresponsibleModule)
-      File has no descriptive comment (IrresponsibleModule)
+      Dir has no descriptive comment (IrresponsibleModule)
```
troessner: I have no idea why this wasn't failing before but in fact this feature failure was legit. If you check out spec/samples/inline.rb you can see that File DOES have a descriptive comment: https://github.com/troessner/reek/blob/develop/spec/samples/inline.rb#L682 at the end whereas Dir does not have one: https://github.com/troessner/reek/blob/develop/spec/samples/inline.rb#L684 neither at the beginning, nor at the end.
So if you just copy the Dir and File declaration (just the class name and the "end" and the comments) to your editor, save that as a separate file and run reek on it it will rightfullly report the same.

features/samples.feature
```diff
@@ -50,6 +50,7 @@ Feature: Basic smell detection
       Inline::C#parse_signature has the variable name 'x' (UncommunicativeVariableName)
       Inline::C#parse_signature is controlled by argument raw (ControlParameter)
       Inline::C#strip_comments doesn't depend on instance state (UtilityFunction)
+      Module has no descriptive comment (IrresponsibleModule)
```
troessner: Same here. This should have failed before but it didnt since "class Module" indeed does not have a descriptive comment.

troessner:
This should make travis happy again.
The spec failures on develop are actually legit. I have no idea why and quite frankly, I don't think it makes sense to waste time what exactly went wrong there.

mvz:
You didn't pin it firmly enough so Travis still uses 2.2.2.5. You need to put `'2.2.2.2'` instead of `'2.2.2.5'`. 

It's true that some of the newly detected smells make sense, but it stopped considering `File` smelly because it has a last line  `end # class File`, which I think is wrong.

mvz:
BTW, I'm fine with leaving the cuke changes in, but then we should pin to `~> 2.2.2.5` because they will fail with parser 2.2.2.2.

Also, sorry about messing up my comment above. I fixed it.

troessner:
@mvz my bad, i updated the PR

PR #524, opened 2015-05-27
Improve irresponsible module

mvz:
Changes comment handling to only consider preceding comments as potentially descriptive. In particular, the following is not:

```
class Foo
  ...
end # class Foo
```

I haven't considered what to do with the following case, yet:

```
class Foo # handles bars
  ...
end
```

### DIFF
--- a/lib/reek/core/ast_node.rb
+++ b/lib/reek/core/ast_node.rb
@@ -10,8 +10,16 @@ module Reek
         super
       end
 
-      def comments
+      def full_comment
         @comments.map(&:text).join("\n")
+      end
+
+      def leading_comment
+        line = location.line
+        comment_lines = @comments.select do |comment|
+          comment.location.line < line
+        end
+        comment_lines.map(&:text).join("\n")
       end
 
       # @deprecated
--- a/spec/reek/smells/irresponsible_module_spec.rb
+++ b/spec/reek/smells/irresponsible_module_spec.rb
@@ -4,14 +4,6 @@ require_relative '../../../lib/reek/smells/irresponsible_module'
 require_relative 'smell_detector_shared'
 
 RSpec.describe Reek::Smells::IrresponsibleModule do
-  before(:each) do
-    @bad_module_name = 'WrongUn'
-    @source_name = 'dummy_source'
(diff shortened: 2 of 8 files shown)


mvz:
I'd first like #519 to be merged since that will change some class names in the present PR.

spec/reek/smells/irresponsible_module_spec.rb
```diff
@@ -63,14 +48,26 @@ class Bar; end
...
-    expect(smells[0].smell_category).to eq(described_class.smell_category)
-    expect(smells[0].smell_type).to eq(described_class.smell_type)
-    expect(smells[0].parameters[:name]).to eq('Foo::Bar')
-    expect(smells[0].context).to match(/#{smells[0].parameters['name']}/)
+    expect(src).to reek_of :IrresponsibleModule, name: 'Foo::Bar'
+  end
+
+  context 'when a smell is reported' do
+    before do
+      @bad_module_name = 'WrongUn'
```
troessner: What is "WrongUn" supposed to mean?

lib/reek/core/ast_node.rb
```diff
@@ -14,6 +14,10 @@ def comments
         @comments.map(&:text).join("\n")
       end
 
+      def comments_array
```
troessner: Not the biggest fan of hungarian notation. Why not just "comments"?

spec/reek/smells/irresponsible_module_spec.rb
mvz: I believe it was meant to mean 'wrong one' originally. Looks like that whole variable can go, actually.go, actually.

lib/reek/core/ast_node.rb
mvz: I introduced it because `#comments` was already taken and I didn't want to break everything immediately. I'll refactor.

mvz:
Updated and rebased.

mvz:
Ah, noticed one more thing. Hold on.

mvz:
Done.

troessner:
Very nice @mvz - merged!.:)

PR #525, opened 2015-05-28
Require smells explicitly and fix some missing require_relatives

chastell:
As per the discussion in #513, let’s just require all of the smells explicitly.

### DIFF
--- a/lib/reek/smells/attribute.rb
+++ b/lib/reek/smells/attribute.rb
@@ -1,4 +1,6 @@
+require_relative 'smell_configuration'
 require_relative 'smell_detector'
+require_relative 'smell_warning'
 
 module Reek
   module Smells
--- a/lib/reek/smells/boolean_parameter.rb
+++ b/lib/reek/smells/boolean_parameter.rb
@@ -1,4 +1,5 @@
 require_relative 'smell_detector'
+require_relative 'smell_warning'
 
 module Reek
   module Smells
--- a/lib/reek/smells.rb
+++ b/lib/reek/smells.rb
@@ -1,13 +1,24 @@
-require 'pathname'
-
-(Pathname.new(__FILE__).dirname + 'smells').children.each do |path|
-  require_relative "smells/#{path.basename('.rb')}"
-end
-
-module Reek
-  #
-  # This module contains the various smell detectors.
-  #
-  module Smells
-  end
-end
+require_relative 'smells/attribute'
+require_relative 'smells/boolean_parameter'
+require_relative 'smells/class_variable'
+require_relative 'smells/control_parameter'
+require_relative 'smells/data_clump'
+require_relative 'smells/duplicate_method_call'
+require_relative 'smells/feature_envy'
+require_relative 'smells/irresponsible_module'
+require_relative 'smells/long_parameter_list'
+require_relative 'smells/long_yield_list'
+require_relative 'smells/module_initialize'
+require_relative 'smells/nested_iterators'
+require_relative 'smells/nil_check'
+require_relative 'smells/prima_donna_method'
+require_relative 'smells/repeated_conditional'
+require_relative 'smells/too_many_instance_variables'
+require_relative 'smells/too_many_methods'
+require_relative 'smells/too_many_statements'
+require_relative 'smells/uncommunicative_method_name'
+require_relative 'smells/uncommunicative_module_name'
+require_relative 'smells/uncommunicative_parameter_name'
+require_relative 'smells/uncommunicative_variable_name'
+require_relative 'smells/unused_parameters'
+require_relative 'smells/utility_function'
--- a/lib/reek/smells/class_variable.rb
+++ b/lib/reek/smells/class_variable.rb
@@ -1,5 +1,6 @@
 require 'set'
 require_relative 'smell_detector'
+require_relative 'smell_warning'
 
 module Reek
   module Smells
--- a/lib/reek/smells/control_parameter.rb
+++ b/lib/reek/smells/control_parameter.rb
@@ -1,4 +1,5 @@
 require_relative 'smell_detector'
+require_relative 'smell_warning'
 
 module Reek
   module Smells
--- a/lib/reek/smells/data_clump.rb
+++ b/lib/reek/smells/data_clump.rb
@@ -1,4 +1,5 @@
 require_relative 'smell_detector'
+require_relative 'smell_warning'
 
 module Reek
   module Smells
--- a/lib/reek/smells/duplicate_method_call.rb
+++ b/lib/reek/smells/duplicate_method_call.rb
@@ -1,4 +1,5 @@
 require_relative 'smell_detector'
+require_relative 'smell_warning'
 
 module Reek
   module Smells
--- a/lib/reek/smells/feature_envy.rb
+++ b/lib/reek/smells/feature_envy.rb
@@ -1,4 +1,5 @@
 require_relative 'smell_detector'
+require_relative 'smell_warning'
 
 module Reek
   module Smells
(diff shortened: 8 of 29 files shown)


lib/reek/smells/attribute.rb
```diff
@@ -1,4 +1,6 @@
+require_relative 'smell_configuration'
```
mvz: smell_detector already requires smell_configuration, so this one is not needed per se.

chastell: Right – as long as `smell_detector.rb` keeps requiring `smell_configuration.rb`. :)

Disclaimer: the blow is a bit hair-splitting in most cases, but I want to clarify my approach (sorry!).

My general rule-of-thumb is ‘if `attribute.rb` refers to `SmellConfiguration` then it’s coupled to it and should require the file that defines it’ – this goes along my earlier argument ‘if a file starts with a long list of `requires` then it’s because the implementation is too tightly coupled to too many things, and the `requires` should not be hidden to pretend it’s not’.

Or, to rephrase – if `smell_detector.rb` requires `smell_configuration.rb` and `attribute.rb` should not require `smell_configuration.rb`, then `SmellDetector` should have a method that exposes `SmellConfiguration` and `Attribute` implementation should use that method to get it (rather than calling `SmellConfiguration` explicitly).

lib/reek/smells/boolean_parameter.rb
```diff
@@ -1,4 +1,5 @@
 require_relative 'smell_detector'
+require_relative 'smell_warning'
```
troessner: Why not require this in smell_detector?

lib/reek/smells/attribute.rb
mvz: @chastell I agree.

This particular case is a bit odd since we're using constants from SmellConfiguration. I don't particularly like that system, but I also don't know yet how we can improve it.

lib/reek/smells/boolean_parameter.rb
mvz: I think the same argument applies as above. In this case, we *can* easily expose `SmellWarning` through a method (e.g. `create_smell_warning`).

lib/reek/smells/attribute.rb
chastell: I think I’m ok with using constants from `SmellDetector` once `smell_detector.rb` is required, as long as `SmellDetector` is not a namespacing-only `Module`. I think my reasoning is as follows:

* if `SmellDetector` is a class, then
  * non-module/class constants in that class can be expected to be available once `smell_detector.rb` is required;
  * modules/classes embedded inside that class are for private use by that class,
  * modules/classes that are supposed to be used from outside of `SmellDetector` should live in their own files (like `smell_detector/foo.rb`) and be required explicity;
* if `Reek` is a top namespacing module, then either
  * (more realistically) `require 'reek'` should load enough of `reek` to expose whatever we consider to be our public API, or
  * (more of a pipe-dream) all of our public API should be available through methods defined on `Reek` and `require 'reek'` should not be expected to load any of the underlying classes (even though it in practice would to let those methods work).

chastell: (Note that these are just my current ideas, I have tested them only to some extent in rather small personal projects; I’m not sure whether they’re feasible in larger codebases.)

lib/reek/smells/boolean_parameter.rb
troessner: Ok, I read up on your arguments above.

>> we can easily expose SmellWarning through a method (e.g. create_smell_warning).

Doesn't feel to good to me.
Quite frankly, I believe we should have gone down the road of smells/smell_types/.
With SmellDetector requiring SmellConfiguration.

But anyways.:)

The arguments from above are valid for sure. But putting this require in EVERY smell just seems wrong. We should put it in the SmellDetector. Practical usage over architectural correctness.:)

troessner:
Suggestion: 

We require everything smell-related in smells.rb and ONLY in smells.rb. 
And every other e.g. in /core which wants to use anything smell-related has to require smells.rb, and that is that.

lib/reek/smells/boolean_parameter.rb
chastell: > The arguments from above are valid for sure. But putting this require in EVERY smell just seems wrong.

Well, that’s because every smell is coupled to both `SmellDetector` and `SmellWarning`. :)

>  We should put it in the SmellDetector. Practical usage over architectural correctness.:)

Oh definitely. In my pet projects I’m starting to lean towards ‘this is required 20 times across the project, so it’s a core concept of the project, let’s move the require to `lib/foo.rb`’).

chastell:
I have a gut feeling that the mistake we’re making is putting smells (descendants of `SmellDetector`) in the same directory/at the same level as `SmellConfiguration`, `SmellRepository` and `SmellWarning` (which, arguably, are _not_ smells). I’d either move the smells to `smells/detectors` or the other three somewhere else – but that’s a discussion for another PR (maybe `SmellDetector` should be renamed to, y’know, `Smell`?). :)

> We require everything smell-related in smells.rb and ONLY in smells.rb. And every other client e.g. in /core which wants to use anything smell-related has to require smells.rb, and that is that.

Ok, I’ll adjust this PR later today.

chastell:
Does this mean that individual smells should not require anything? If they `require_relative '../smells'` then there will be circular requires again. :)

mvz:
Uuh .. I liked it better when each detector just required `smell_detector`. Since `smell_detector` requires `smell_warning` there really wasn't any practical problem.

If you're going to require `smells`, all tests for individual detectors will have to be adjusted as well. Again, I'd prefer requiring `smell_detector` in each detector.

I don't like it if there is a file where you basically say 'Don't require this file directly, require this other file instead'.

lib/reek/core/examiner.rb
```diff
@@ -6,7 +6,7 @@
 # See also https://github.com/troessner/reek/pull/468
 require_relative 'tree_walker'
 require_relative '../cli/warning_collector'
-require_relative '../smells/smell_repository'
+require_relative '../smells'
```
mvz: Here then, I'd require smell_repository, since that is what Examiner uses directly.

lib/reek/core/tree_walker.rb
```diff
@@ -2,6 +2,7 @@
 require_relative '../context/module_context'
 require_relative '../context/root_context'
 require_relative '../context/singleton_method_context'
+require_relative '../smells'
```
mvz: Not needed?

mvz:
So basically I'd go for the first commit minus the explicit requires of `smell_warning` in each detector.

chastell:
Ok, so we have three preferences (let me know if I got them wrong):

* @mvz’s require what every file needs except for `smell_warning.rb`,
* @troessner’s require only `smells.rb` everywhere,
* @chastell’s require what every file needs.

I’m happy with any of them, so when you two come to a conclusion let me know and I’ll adjust this PR. :)

troessner:
>> Ok, so we have three preferences (let me know if I got them wrong):
@mvz’s require what every file needs except for smell_warning.rb,
@troessner’s require only smells.rb everywhere,
@chastell’s require what every file needs.

Yes! But now we have a stalemate.:D

troessner:
Ok, f**k it, let's go with @chastell 's way. The more explcit the better.
@mvz objections?

chastell:
(It’s not only about explicitness and architectural correctness, but also about having a simple rule for starters and not having to make microdecisions ‘is this class a special snowflake enough to not be required?’. Obviously ‘require what every file needs’ is just the first approximation, the moment it gets unwieldy it should be revisited – but I’d say two requires in a file that hardcodes two non-core classes is not yet _too many_.) ;)

troessner:
@chastell Then let's go for that - I guess you then need to update this pull request significantly, right?.:)

chastell:
I reverted this to the first commit and rebased, let me know if I should adjust anything.

mvz:
I'm ok with @chastell's way :-).

PR #533, opened 2015-06-07
Use current ruby instead of ruby21 for parsing.

troessner:
I believe the Parser::Ruby21 requirement is now obsolete and we should use Parser::CurrentRuby.


mvz:
When I introduced parser I decided to use the version that parsed the then most recent Ruby, since it allows the broadest range in syntax. Using CurrentRuby means reek running on Ruby 2.0 will fail on some 2.1 syntax. Since reek is not a syntax checker, I think that is not what we want. TL;DR: We should probably use Parser::Ruby22 now.

troessner:
Closing this one in favor of #534

PR #536, opened 2015-06-13
Drop unnecessary module qualification in 'ast/' namespace.

troessner:


chastell:
Oh, my kind of clean-up! ;) :+1:

PR #541, opened 2015-06-20
[WIP] Recognize singleton methods defined via module function.

### LINKED ISSUE #27: module_function does not disable Utility Function
A bit like a method on a class, the following should presumably not trigger Utility Function?

    module MyModule
      def reset!
        # ...
      end
      module_function :reset!
    end

troessner:
Please don't review yet, still tweaking things here and there.
Fixes #27

### DIFF
--- a/lib/reek/ast/sexp_extensions.rb
+++ b/lib/reek/ast/sexp_extensions.rb
@@ -249,6 +249,70 @@ module Reek
         end
       end
 
+      # Checking if a method is a singleton method.
+      module SingletonMethod
+        def singleton_method?
+          singleton_method_via_class_self_notation? ||
+            singleton_method_via_module_function?
+        end
+
+        # Ruby allows us to make a method a singleton_method after having defined the
+        # method via `module_function`.
+        #
+        # To check if we used "module_function" for a method we need to check on the parent
+        # level of the method in question if there is a call to "module_function".
+        # Given someting like
+        #   class C
+        #     def m; 3 + 7; end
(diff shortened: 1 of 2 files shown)


troessner:
Ready for review!

chastell:
Just did a cursory look for now, but does it handle the common (at least in my code) case where `module_function` is a scoping operator? The tests don’t cover this and I don’t think the implementation covers this either:

```Ruby
module Foo
  module_function

  def bar
    :bar
  end

  def baz
    :baz
  end
end

Foo.bar   #=> :bar
Foo.baz   #=> :baz
```

chastell:
(I did a partial stab at this before, hoping I can use something like the `Smells::Attribute` tracking of its `VISIBILITY_MODIFIERS`, but didn’t have time to get with it very far.)

troessner:
>> Just did a cursory look for now, but does it handle the common (at least in my code) case where module_function is a scoping modifier? The tests don’t cover this and I don’t think the implementation covers this either:

It doesn't! I knew I would miss a couple of cases.:)
I'll look into it.

chastell:
Awsome. The trick is that the first bare `public`, `protected` or `private` after a bare `module_function` resets the following methods to not be `module_functions` anymore (unless another bare `module_function` happens, of course).

mvz:
I think `Attribute` does some visibility tracking so you may want to take a look at that and perhaps extract common functionality.

troessner:
Closing this one to re-open it against master.

PR #543, opened 2015-06-20
Remove "develop" rake task.

troessner:
I believe nobody is using this anymore and we can get rid of it.


mvz:
We do use this to keep `config/defaults.reek` in sync with reality. However, that file is never read so serves more as documentation. If we get rid of the task we should get rid of the file as well.

troessner:
>> If we get rid of the task we  should get rid of the file as well.

Why? Why not leave it there as documentation?
For instance, I find it quite convenient and helpful to be able to link to this...

mvz:
> Why not leave it there as documentation?

Sure, but then we should leave the task so the file is automatically updated when we change config defaults.

troessner:
>> Sure, but then we should leave the task so the file is automatically updated when we change config defaults.

Good point..:)

-> Closed.

Issue #526, opened 2015-05-28
API for reek v3: The first proposal


troessner:
I thought about our future API for reek 3. I believe that this: https://github.com/troessner/reek/blob/develop/docs/API.md is already a good start.

What I had in mind:

### Set up

```Ruby
require 'reek/api'
```

This will make available:

* All the reporters (see below)
* The Examiner

### Usage

1.) Create a reporter. Can be:

* TextReport
* YAMLReport
* JSONReport
* HTMLReport
* XMLReport

```Ruby
reporter = Reek::CLI::Report::TextReport.new
```

2.) Create an examiner:

```Ruby
examiner = Reek::Examiner.new(source)
```

where `source` can be a File, a string or an array (so basically like it is right now)

3.) Add the examiner to the reporter:

```Ruby
reporter.examiners << examiner
```

4.) Print the result:

```Ruby
puts reporter.show
```

### Exposed API methods:

* Reporter#show
* Reporter.examiners << 
* The initalizers of the reporters and Examiner

### Necessary changes

* Move Reek::Core::Examiner to the top level so it becomes Reek::Examiner
* Create reek/api
* Change Reporter.add_examiner to Reporter.examiners <<

### Spec API

Our spec matchers described [here](https://github.com/troessner/reek/blob/develop/docs/RSpec-matchers.md) should stay like they are.

## [2] Outlook

In the next step I would add a none breaking change that 

1.) exposes our AppConfiguration
2.) exposes our SmellRepository
3.) makes the API more convenient to use like this:

```Ruby
report = Examiner.new source: ,
                      reporter:         TextReport.new,
                      smell_repository: SmellRepository.default,
                      configuration:    AppConfiguration.default
report.show                      
```

This would obviously be a lot harder than [1] since AppConfiguration would be settable on an examiner base. So I'd like to go with [1] to keep things easy and get to a stable version quick and then put out [2] as an add-on.

What do you guys think?

chastell:
I like it in general, but I’m unsure about exposing all of the `Reek::*` plumbing.

I’d consider wrapping the API calls in thin `Reek.foo` (or `Reek::API.foo`) methods – this way we’ll be able to refactor things without having to maintain backward-compatible constants:

```Ruby
require 'reek/api'
reporter = Reek::API.reporter(:text)
examiner = Reek::API.examiner_for(source)
reporter.examiners << examiner
puts reporter.show
```

[...]

PR #544, opened 2015-06-21
Create and document stable API

mvz:
#526 

To view what is left as visible API, run YARD like so:

```bash
yardoc --hide-api private
```

This leaves us with a basic API that matches what we use internally. I think this will improve maintainability, since we won't have a separate API only for third party users.

One eyesore is the deeply nested report namespace. My preference would be to move it up one or two levels so we have either `Reek::Report::TextReport` or `Reek::TextReport`.

Less of an issue are `Reek::Smells::SmellWarning` (since it is not supposed to be created by the API user), and `Reek::Rake::Task`, which could become `Reek::SmellWarning` and `Reek::RakeTask` respectively. I'm OK with leaving them as-is as well.


lib/reek/ast/ast_node.rb
```diff
@@ -4,6 +4,8 @@ module Reek
   module AST
     # Base class for AST nodes extended with utility methods. Contains some
     # methods to ease the transition from Sexp to AST::Node.
+    #
+    # @api private
```
troessner: Might make sense to review / merge #545 a.s.a.p so you can rebase your branch.

lib/reek/cli/report/report.rb
```diff
@@ -1,5 +1,7 @@
 require 'rainbow'
 require 'json'
+require_relative 'formatter'
+require_relative 'heading_formatter'
```
troessner: Already fixed in https://github.com/troessner/reek/pull/540/files#diff-9eaec13b838dd3d833fd276a922f1ac5R3

docs/API.md
```diff
@@ -28,15 +25,24 @@ class Dirty
...
-`reek` can take `source` as `String`, `File` or `IO`.
+`Reek::Examiner.new` can take `source` as `String`, `File` or `IO`, or an array
+of file names.
+
+```
+# Examine a file object
+reporter.add_examiner Reek::Examiner.new(File.new('dirty.rb'))
+# Examine source from an array of file names
+reporter.add_examiner Reek::Examiner.new(['also_dirty.rb', 'clean.rb')
+```
```
troessner: Hmmmmm. Did you take @chastell / my suggestion from here: https://github.com/troessner/reek/issues/526

```Ruby
report = Reek::API.report(source:   source,
                          reporter: :text,
                          smells:   %i(attribute data_clump nil_check),
                          config:   config) # a path to a YAML file or a Hash
puts report.show
```

into account as well? Or rather: Are you going to?
I really think we should get rid of this "report.add_..." thingie for our public API.
Or should I not even comment on that since it's WIP?..:)

mvz: Yeah, my first goal was to get a minimal API in place based on the current code, and to only then see what can be improved.

* Regarding an API namespace: I don't think that is needed. We can just expose methods on `Reek`. We can even use those methods ourselves.
* Regarding accessing reporters by a symbol: I think that is a good idea, since it means not having to fix the reports' full class name.
* Rergarding the all-in-one creation of a reporter: I think that would be overkill. Ideally, the publicly exposed API would be something we could use, and do use, inside reek itself.

lib/reek/ast/ast_node.rb
mvz: I'll rebase.

docs/API.md
mvz: One final thing: I'm not too sure about `report.examiners <<` versus `report.add_examiner`.

The naïve implementation of `report.examiners <<` would basically boil down to `attr_accessor :examiners`, which would expose the array of examiners to the outside world and allow any manipulation. An alternative would be to expose some proxy object that only allows `:<<`, but that would be confusing when people try `report.examiners = report.examiners + some_more_examiners`.

I don't think `add_examiner` is so bad.

mvz:
First 'draft' is ready. I updated the PR description with some instructions.

Extracting the SmellConfiguration will be quite complex and is definitely something for a 3.1 release.

docs/API.md
```diff
@@ -48,3 +50,30 @@ JSONReport
...
 ```
+
+## Accessing the smell warnings directly
+
+You can also access the smells detected by an examiner directly:
+
+```ruby
+require 'reek'
+
+source =<<END
```
troessner: Let's use 

```Ruby
<<-END
```

and then indent the example.

lib/reek/ast.rb
```diff
@@ -0,0 +1,6 @@
+# FIXME: Require this somewhere
+module Reek
+  # @api private
+  module AST
```
troessner: I don't get it, why the empty module? And why require it somewhere when it's empty? Is that a yard-thing?

lib/reek/cli/report/report.rb
```diff
@@ -48,16 +61,16 @@ def show
...
           @examiners.each_with_object([]) do |examiner, result|
             result << summarize_single_examiner(examiner)
           end
         end
 
-        private
-
         def display_summary
-          smells.reject(&:empty?).each { |smell| puts smell }
+          smell_summaries.reject(&:empty?).each { |smell| puts smell }
```
troessner: How about rejecting empty summaries in `smell_summaries` right away?

lib/reek/context.rb
```diff
@@ -0,0 +1,6 @@
+# FIXME: Require this somewhere
```
troessner: Same like above, why?

lib/reek/examiner.rb
```diff
@@ -62,12 +63,10 @@ def smelly?
     private
 
     def run
-      @sources.each do |source|
```
troessner: Awesome, this has been bugging me everytime I came around this piece of source but I never got to fix it.:)

spec/reek/spec/should_reek_of_spec.rb
```diff
@@ -54,23 +54,6 @@
...
-    end
-
-    it 'matches a smelly String' do
-      expect(@matcher.matches?(@smelly_dir)).to be_truthy
-    end
-
-    it 'doesnt match a fragrant String' do
-      expect(@matcher.matches?(@clean_dir)).to be_falsey
-    end
-  end
```
troessner: Why is this spec not necessary anymore?

spec/reek/examiner_spec.rb
```diff
@@ -44,45 +44,6 @@
...
-      end
-    end
-
-    before :each do
-      smelly_dir = Dir['spec/samples/masked_by_dotfile/*.rb']
-      @examiner = described_class.new(smelly_dir)
-    end
-
-    it_should_behave_like 'one smell found'
-  end
```
troessner: I guess the described behaviour here is already spec'ed somewhere else so that you just removed it?

spec/reek/spec/should_reek_spec.rb
```diff
@@ -22,31 +22,6 @@
...
-    it 'masks smells using the relevant configuration' do
-      with_test_config('spec/samples/clean_due_to_masking/masked.reek') do
-        expect(matcher.matches?(masked_dir)).to be_falsey
-      end
-    end
-
-    it 'reports the smells when should_not fails' do
-      matcher.matches?(smelly_dir)
-      expect(matcher.failure_message_when_negated).to match('UncommunicativeVariableName')
-    end
```
troessner: Same question here.

lib/reek/examiner.rb
```diff
@@ -15,23 +15,17 @@ module Reek
...
-    #   if it is a File, the file is opened and parsed for Ruby source code;
-    #   and if it is an Array, it is assumed to be a list of file paths,
-    #   each of which is opened and parsed for source code.
+    #   if it is a File or IO, it is opened and Ruby source code is read from it;
+    #
+    # @param smell_types_to_filter_by [Array<String>]
+    #   List of smell types to filter by.
     #
     def initialize(source, smell_types_to_filter_by = [])
-      @sources      = Source::SourceRepository.parse(source)
```
troessner: Very cool.

troessner:
Reviewed, looks good overall, just a couple of questions and small suggestions then this bad boy is good to go!

lib/reek/ast.rb
mvz: Yeah, it's a YARD thing, but it can go now.

spec/reek/examiner_spec.rb
mvz: This basically tests passing a dir to `Examiner.new`, which is no longer supported.

spec/reek/spec/should_reek_of_spec.rb
mvz: Because we're not allowing passing a dir to `Examiner.new` anymore, the reek_of matchers also don't support passing a directory anymore.

spec/reek/spec/should_reek_spec.rb
troessner: Ok, I guess the answer is the same as above, makes sense.:)

mvz:
Updated and squashed.

What do you all think about the report namespace?

lib/reek/cli/report/report.rb
mvz: I'll update this as well.

troessner:
Awesome, merged.

Issue #542, opened 2015-06-20
Improvements to Ast::AstNode


troessner:
3 suggestions:

1.) We rename Ast::AstNode to Ast::Node so that it's in line with its mother ::Parser::AST::Node.
The slightly different naming is tripping me up all the time.

2.) We get rid of everything below: https://github.com/troessner/reek/blob/develop/lib/reek/ast/ast_node.rb#L27
Because the methods below don't make sense to me. They are rather more confusing than helpful.
E.g. "first" doesnt make sense to me. It's a type and NOT the first element of a collection. The same goes for [] which **would** make sense if not the first element in "elements" would be the type.

3.) The only place where we include SexpNode is AstNode. Which makes this a useless abstraction since the only effect it has is that the code is.....somewhere else. Let's dissolve this abstraction and move all the into Ast::AstNode (or rather Ast::Node if we follow [1]).

WDYT?

chastell:
:+1: as long as it’s spelt `AST::Node`. ;) #capitalisationjunkie

I especially agree on `#first` not making sense and the distinction between `ASTNode` and `SexpNode` being problematic to newcomers and not worth keeping.

mvz:
> E.g. "first" doesnt make sense to me. It's a type and NOT the first element of a collection. The same goes for [] which would make sense if not the first element in "elements" would be the type.

There's a reason I marked them `@deprecated` :smile:.

As long as we get rid of any other `Node` in reek, I'm ok with renaming.

Finally, :+1: on merging `SexpNode` into `AST::Node`. It's a remnant from before we used `parser`.

troessner:
Fixed [1] and [3] in #545 - [2] is a little more complicated and I'll handle this in a separate issue.

troessner:
>> Fixed [1] and [3] in #545 - [2] is a little more complicated and I'll handle this in a separate issue.

Forget to close this one apparently - closed.

PR #545, opened 2015-06-21
Feature/refactor ast node and sexp node

troessner:
This pull request fixes

>> 1.) We rename Ast::AstNode to Ast::Node so that it's in line with its mother ::Parser::AST::Node.
The slightly different naming is tripping me up all the time.

and 

>> 3.) The only place where we include SexpNode is AstNode. Which makes this a useless abstraction since the only effect it has is that the code is.....somewhere else. Let's dissolve this abstraction and move all the into Ast::AstNode (or rather Ast::Node if we follow [1]).

from #542 ([2] seems a little bit more complicated and I'd like to tackle it separately).


troessner:
P.S.: The two commits are intentional and warranted I believe.

mvz:
:+1:

Issue #548, opened 2015-06-22
CodeClimate integration


andyw8:
CodeClimate recently opened up their platform, so it seems it would now be possible to run Reek on it.

http://blog.codeclimate.com/blog/2015/06/19/code-climate-platform/
https://github.com/codeclimate/spec

Has anyone else started looking into this?

troessner:
>> Has anyone else started looking into this?

No - would you like to do that?..:)

troessner:
I just talked to one of the guys via mail (hope he's ok with me making this public, nothing private about it):

```
Hey Timo-

Reek sounds like a great addition to our platform!

We don't have plans to build a Reek analysis engine right now, but any chance you (or your team) might be interested in doing so?

Our new analysis platform is open and extensible, so anyone can write their own Code Climate-compatible analysis engines. We can then run those engines for you (on any project). We designed our engine specification to be very straight-forward, and it often only takes 2-3 hours to stand up a brand new engine from scratch.

If you're interested, just let me know and I can send along some additional info.

Thanks again for the email!
```

I believe we should get this going. CodeClimate is awesome. reek is. 2 + 2 == hell, yes.

troessner:
Alrightie, I'll definitely look into this!

Doesn't seem too complicated:

https://github.com/codeclimate/spec/blob/master/SPEC.md

troessner:
Quoting from that spec:

```
Certain keys of the config object will be passed to all engines, and must be respected in order to conform with the specification. Today, there is only one such key:

exclude_paths -- An Array of file paths (relative to the /code directory) that should be ignored for the purposes of analysis. No Issues should be emitted for excluded files.
```

So let's implement this first. I'm gonna look into it.

troessner:
Now that we have done all the homework I restart looking into this.

andyw8:
I started working on a small Ruby library for building the data structures that the CodeClimate engine needs:

https://github.com/andyw8/codeclimate-engine-rb

andyw8:
My early attempt at an engine: https://github.com/andyw8/codeclimate-reek

troessner:
With #822 merged we can close this one.
Codeclimate is right now working on integrating `Reek`

Issue #529, opened 2015-05-29
Make reek more rails-friendly or add a "rails" mode.


troessner:
See #401

troessner:
I'd like to take this one if nobody objects.:)
Let's discuss over the next days in this thread what a "rails-friendly mode" would mean to us.

troessner:
Alrightie, I thought about this today and here's my basic idea:

I'd like to start of with a Sandi Metz quote: "Make the change easy (warning: that might be hard) and then make the change".:)

The basic idea is to **not** implement something like a "rails-mode" but rather to make configurations possible that were not possible before so that basically every user can make up a "rails-mode" for himself (we can of course still give defaults though).

This default configuration I had in mind would be:

```yaml
"app/controllers":
  IrresponsibleModule:
    enabled: false
  NestedIterators:
    enabled: false
"app/helpers":
  IrresponsibleModule:
    enabled: false
  UtilityFunction:
    enabled: false
```

A couple of nice benefits we get with this approach:

1.) We don't litter our codebase with things that are rails specific
2.) Every user can generate his own settings with us still being able to give a defaults
3.) We get things like a "Sinatra" mode or similar for free
4.) reek config files get easier since you can kill a couple of birds with one stone (e.g. using broader settings than just excluding things based on method names and so on)

What do you guys think of that?

P.S.: If we go with this approach I'd like to discuss the yaml format first. In an ideal world I'd probably then scope the smells under a top level key "Smells" and the settings from above under a top level key "directory directives". But that makes it pretty uncool. So I'd rather just try to check if a config key is a smell -> if it isn't assume it's a directory -> if it isnt raise ArgumentError.

chastell:
RuboCop uses offence names (their idea of ‘smells’) as top-level keys and `Exclude` subkeys that match paths. We use `exclude` subkeys to match class/method names (which is much more flexible except when you don’t namespace your sub-hierarchies, _right, RAAAAILS?_), but I think it’s more readable to keep the config tree smell-, not path-based:

```YAML
IrresponsibleModule:
  skip:
  - app/controllers/**/*
  - app/helpers/**/*

NestedIterators:
  skip:
  - app/controllers/**/*

UtilityFunction:
  skip:
  - app/helpers/**/*
```

troessner:
>>  but I think it’s more readable to keep the config tree smell-, not path-based:

I disagree for the following reasons:

[...]

Issue #408, opened 2015-03-10
possible to set folder in config.reek ?


tansaku:
Hey, reek is totally awesome.

I'm just getting set up with a config.reek file like so:

```yml
IrresponsibleModule:
  enabled: false
  exclude: []
```

I was wondering if there is a way to set the default directory/folder in that file, e.g. I just want reek to check lib by default, so I'd like something like:

```yml
CheckFolder: lib
IrresponsibleModule:
  enabled: false
  exclude: []
```

And than I run 'reek' instead of 'reek lib'. Is that just silly?  Wrong place perhaps?

Many thanks in advance

troessner:
Good idea!
Actually it should be pretty easy to implement this using https://github.com/troessner/reek/blob/master/lib/reek/configuration/app_configuration.rb
Then only thing you'd then have to handle is what should happen when a path is set in the configuation but also passed via CLI.

Care for a pull request?..:)

troessner:
We merged #550 -> closing this one.
I realize that "excluding" is not the same "including" but AFAIS this is close enough.
I'd still be open to PRs that add an explicit "include" option of course.

Issue #349, opened 2014-12-12
Global Exclusions?


jfelchner:
I'm thinking I must just be missing this.  Is there a way to say that I would not like any smell detection on a certain subset of files?  Like `spec/factories/*.rb`?

mvz:
How about making this part of Reek 2.1?

troessner:
Good idea!

mvz:
@troessner are you still looking into this?

troessner:
>> @troessner are you still looking into this?

It's still on my todo list but I want to further look into the refactoring started in #395 - the issue is closed because i didnt want to have issues around for such vague tasks like "Refactor /core." But this refactoring is what I'm gonna do next.

Long story short: Feel free to take over this issue if you like, if not then yes, I'd come back to this issue somewhere in the future.:)

troessner:
Implemented in #550 - closing this.

PR #550, opened 2015-06-28
Make directories excludable via configuration.

troessner:
Relates to #548 #349 #408 and #529 

Apart from that I hate ruby. I've been doing ruby for years now and when i checked the docs here: http://docs.ruby-lang.org/en/2.1.0/Find.html I thought the 

```Ruby
if File.basename(path)[0] == ?.
```

was a syntax error. Of course it isn't in Ruby.

```Bash
[1] pry(main)> ?a
=> "a"
[2] pry(main)> ?ab
SyntaxError: unexpected '?', expecting end-of-input
```
The dark corners of ruby baffle me every time.


spec/reek/configuration/app_configuration_spec.rb
```diff
@@ -46,20 +46,33 @@
...
     it 'resets the configuration' do
-      Reek::Configuration::AppConfiguration.load_from_file(sample_configuration_path)
+      described_class.load_from_file(sample_configuration_path)
 
-      expect(Reek::Configuration::AppConfiguration.configuration).to eq(sample_configuration_loaded)
-      Reek::Configuration::AppConfiguration.reset
-      expect(Reek::Configuration::AppConfiguration.configuration).to eq(default_configuration)
+      expect(described_class.configuration).to eq(sample_configuration_loaded)
+      described_class.reset
+      expect(described_class.configuration).to eq(default_configuration)
```
troessner: Not really part of this PR but cleaned up according to the good old boyscout rule.:)

lib/reek/source/source_locator.rb
```diff
@@ -16,22 +18,37 @@ def initialize(paths)
...
-            find_sources(Dir["#{pathname}/**/*.rb"])
-          else
-            next pathname if pathname.file?
-            $stderr.puts "Error: No such file - #{pathname}"
-            nil
+      def source_paths
+        relevant_paths = []
+        @paths.map do |given_path|
+          print_no_such_file_error(given_path) && next unless path_exists?(given_path)
+          Find.find(given_path) do |path|
```
troessner: Didn't know that thing before this PR but it seems very convenient: http://ruby-doc.org/stdlib-2.2.0/libdoc/find/rdoc/Find.html

lib/reek/cli/option_interpreter.rb
```diff
@@ -10,7 +10,7 @@ module CLI
     # Interprets the options set from the command line
     #
     class OptionInterpreter
-      include CLI::Input
```
troessner: Boyscout rule.

spec/reek/source/source_locator_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+          'spec/samples/source_with_exclude_paths/ignore_me/uncommunicative_method_name.rb',
+          'spec/samples/source_with_exclude_paths/nested/ignore_me_as_well/irresponsible_module.rb'
+        ]
+      end
+
+      it 'does not use excluded paths' do
+        with_test_config(config) do
+          sources = described_class.new([path]).sources
+
+          expect(sources.map(&:description).sort).
```
troessner: I know that this is kind of implicitly checked by the expectation below but this makes it more palpable for the innocent reader.

troessner:
This bad boy is ready for review @mvz && @chastell !

chastell:
Nice!

Every time I encounter the ‘path as `String`’ vs ‘path as `Pathname`’ duality I cringe. Given that `Pathnames` try to be very `String`-like, is there a chance that we try to use `Pathnames` everywhere and `#to_s` them only where needed? By this I mean `SourceLocator#paths` being an `Array` of `Pathnames` (not `Strings`) and `Configuration::AppConfiguration.exclude_paths` being an `Array` of `Pathnames`. If not, then maybe `Configuration::AppConfiguration.excludes_path?` that can take a `Pathname`? This way e.g. `SourceLocator#path_exists?` could be inlined, and, in general, we could start assuming that we have `Pathanames` by default everywhere in `reek`.

lib/reek/source/source_locator.rb
chastell: Yes, it’s nice, but is `Find.prune` thread-safe (it uses `catch`/`throw` under the hood)?

(Also, are we interested in `reek` being thread-safe?)

troessner:
>> Every time I encounter the ‘path as String’ vs ‘path as Pathname’ duality I cringe. Given that Pathnames try to be very String-like, is there a chance that we try to use Pathnames everywhere and #to_s them only where needed? By this I mean SourceLocator#paths being an Array of Pathnames (not Strings) and Configuration::AppConfiguration.exclude_paths being an Array of Pathnames. If not, then maybe Configuration::AppConfiguration.excludes_path? that can take a Pathname? This way e.g. SourceLocator#path_exists? could be inlined, and, in general, we could start assuming that we have Pathanames by default everywhere in reek.

I could not agree more. So let's switch to Pathname everywhere.
However in order to reduce complexity I'd prefer to do this in a follow up pull request, and not in this one.

lib/reek/source/source_locator.rb
troessner: >> (Also, are we interested in reek being thread-safe?)

Good question.
So what could be the worst thing that happens?
Somebody using reek via our API in 2 different threads with one thread f***ing up "Find" in the other thread?

chastell: Somebody using `reek` in one thread and _anything else_ that uses `Find.prune` in another thread. My very quick-and-dirty proof of concept:

```Ruby
def prune
  throw :prune
end

a = Thread.new do
  puts 'a'
  prune
end

b = Thread.new do
  puts 'b'
  catch(:prune) { puts 'caught' }
end

a.join
b.join
```

– if you run this enough times you’ll see `caught` being printed.

That said, I’m not sure whether `reek` is anywhere to being thread-safe everywhere else; I don’t want this to block this PR as long as we track the knowledge somewhere and come back to this issue eventually. :)

chastell: Ok, opened https://github.com/troessner/reek/issues/554 to discuss/track this. :)

chastell:
Opened https://github.com/troessner/reek/issues/555 (and assigned to myself) to track the switch to `Pathnames`, merging this one. :)

mvz:
Nice!

PR #551, opened 2015-06-28
Pull up Report namespace

mvz:
Based on and follow-up for #544. That one should be merged first.

Shows what pulling up Report would look like. External users would now all reporting classes under Reek::Report, which makes more sense I think.


mvz:
Cleaned up and rebased.

troessner:
Well, that's one area where git or ruby or both or I don't know sucks heavily.
Change of namespace means a ton of indentation changes and add / remove of the file for git.
I have no idea if there are changes in the report files themselves but I'll just trust you on this one.:) (Also the specs didn't change).

Merged!

PR #552, opened 2015-06-28
Move report option mapping into Report

mvz:
Follow-up to #551.

### DIFF
--- a/lib/reek/report.rb
+++ b/lib/reek/report.rb
@@ -1,0 +1,71 @@
+require_relative 'report/report'
+require_relative 'report/formatter'
+require_relative 'report/heading_formatter'
+
+module Reek
+  # Reek reporting functionality.
+  module Report
+    module_function
+
+    # @api private
+    REPORT_CLASSES = {
+      yaml: YAMLReport,
+      json: JSONReport,
+      html: HTMLReport,
+      xml: XMLReport,
+      text: TextReport
+    }
+
+    # @api private
+    LOCATION_FORMATTERS = {
+      single_line: SingleLineLocationFormatter,
+      plain: BlankLocationFormatter,
+      numbers: DefaultLocationFormatter
+    }
+
+    # @api private
+    HEADING_FORMATTERS = {
+      verbose: HeadingFormatter::Verbose,
+      quiet: HeadingFormatter::Quiet }
+
+    # @api private
+    WARNING_FORMATTER_CLASSES = {
+      wiki_links: WikiLinkWarningFormatter,
+      simple: SimpleWarningFormatter
+    }
+
+    # Map report format symbol to a report class.
+    #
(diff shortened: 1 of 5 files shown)


troessner:
Please rebase.:)

mvz:
Done.

lib/reek/report.rb
```diff
@@ -0,0 +1,62 @@
...
+    # @api private
+    def self.warning_formatter_class(warning_format)
+      case warning_format
+      when :wiki_links
+        WikiLinkWarningFormatter
+      else
+        SimpleWarningFormatter
+      end
+    end
+  end
```
chastell: Can we turn them all into `module_functions`? They’d be callable on `Report` but private if some class decides to `include Report`.

lib/reek/report.rb
```diff
@@ -0,0 +1,62 @@
...
+      when :json
+        JSONReport
+      when :html
+        HTMLReport
+      when :xml
+        XMLReport
+      else # :text
+        TextReport
+      end
+    end
```
chastell: I’m thinking whether a more compact syntax wouldn’t be nicer to read:

```Ruby
    def self.report_class(report_format)
      case report_format
      when :html then HTMLReport
      when :json then JSONReport
      when :text then TextReport
      when :xml  then XMLReport
      when :yaml then YAMLReport
      else fail "Unknown report format: #{report_format}"
      end
    end
```

troessner: How about getting rid of quite some duplication by doing

```Ruby
REPORT_FORMAT_TO_CLASS = {
  html: HTMLReport
  # ...
}

def self.report_class(report_format)
  fail "Unknown report format: #{report_format}" unless REPORT_FORMAT_TO_CLASS.key?(report_format)
  REPORT_FORMAT_TO_CLASS[report_format]
end
```
?

lib/reek/report.rb
troessner: :+1: 
Let's kill "self".

lib/reek/report.rb
chastell: Right, I thought about this as an option as well, but didn’t want to disrupt the current solution too much. But if so then let’s use `Hash#fetch`, that’s exactly its role:

```Ruby
REPORT_CLASSES = {
  html: HTMLReport,
  # ...
}

module_function

def report_class(report_format)
  REPORT_CLASSES.fetch(report_format) do
    fail "Unknown report format: #{report_format}"
  end
end
```

troessner: Good idea!

troessner:
Can you please point that to master?

mvz:
Sure.

mvz:
I'll create a new PR once I've processed the comments.

Issue #555, opened 2015-06-30
Switch to Pathnames where possible


chastell:
Ruby is a bit idiosyncratic when it comes to canonical representations of a path – half of the stdlib uses `Strings` for this and the other tries hard to support `Pathnames`.

Let’s switch to `Pathnames` where possible.

mvz:
:+1:

troessner:
:+1:

PR #560, opened 2015-07-03
Only use ruby source files.

troessner:
Not sure if this is urgent (I'd expect most scanned directories to be "clean") but I'd rather get this out and release 3.0.2 immediately.
Pinging @mvz and @chastell for a quick review.


chastell:
I’d probably use `Pathname#extname` rather than `Pathname#to_s#end_with?`, but I’ll fix it in #555 – merging!

mvz:
:+1:

troessner:
>> I’d probably use Pathname#extname

TIL. Awesome.:)

PR #566, opened 2015-07-04
Redo Report::HTMLReport#show and its spec

chastell:
I got frustrated by the `Html file saved` printed on spec runs. This PR:
* removes that line (is this backward-incompatible? if so I’ll make it print it and not spam spec runs),
* uses `File.write` and side-steps a line-wrapping need in `Report::HTMLReport#show`,
* redoes the spec to use a (legibly-named) tempfile.


mvz:
Yes, that message bothered me no end! :heart: 

For backward-compatibility, we should probable print that message, if only so reek doesn't just return without any sign it has done anything (except a file appearing, of course).

Given that, I think we should consider outputting the HTML to stdout, since that is what all the other reporters do. Since this is a breaking change, it will have to wait till reek 4.0. I'm in no hurry whatsoever with that!

The implementation looks good, so if you can get that message printed this PR is good to go.

troessner:
Can you please point that to master?

chastell:
@mvz: agree on the HTML goint to stdout in `reek` 4; would fixing the capitalisation of the output be backward-compatible as well?
@troessner: reopened as #568 (*grumble*stupitGitHub*grumble*)

PR #567, opened 2015-07-04
Recognize singleton methods defined via module function.

troessner:
Re-opened coming from #541 since we'll drop gitflow.

### DIFF
--- a/lib/reek/ast/sexp_extensions.rb
+++ b/lib/reek/ast/sexp_extensions.rb
@@ -249,6 +249,70 @@ module Reek
         end
       end
 
+      # Checking if a method is a singleton method.
+      module SingletonMethod
+        def singleton_method?
+          singleton_method_via_class_self_notation? ||
+            singleton_method_via_module_function?
+        end
+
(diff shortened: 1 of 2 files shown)


troessner:
After thinking  about if for a while I believe we should merge this one without adding this feature:

>> Just did a cursory look for now, but does it handle the common (at least in my code) case where module_function is a scoping modifier? The tests don’t cover this and I don’t think the implementation covers this either

in this PR since this would increase the PR size and complexity significantly. Since we can think of this as a separate feature I'd rather adress this in a follow-up issue / pull request, especially since this PR alone already adds "product value" to reek.
WDYT @chastell @mvz

chastell:
Agree; happy to merge if @mvz is ok with this.

mvz:
The Attribute smell keeps track of visibility modifiers entirely outside of the Node classes, so adding the scoping modifier case would probably undo most of this PR or mean a refactoring of Attribute as well.

As long as we keep that in mind I'm OK with merging this.

troessner:
Yep, lets go with that then.

PR #568, opened 2015-07-04
Redo Report::HTMLReport#show and its spec

chastell:
I got frustrated by the `Html file saved` printed on spec runs. This PR:
* properly capitalises that line (is this backward-incompatible? if so I’ll make it keep the awful case),
* uses `File.write` and side-steps a line-wrapping need in `Report::HTMLReport#show`,
* redoes the spec to use a (legibly-named) tempfile.

### DIFF
--- a/lib/reek/report/report.rb
+++ b/lib/reek/report/report.rb
@@ -129,13 +129,11 @@ module Reek
     class HTMLReport < Base
       require 'erb'
 
-      def show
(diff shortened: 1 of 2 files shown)


troessner:
>> properly capitalises that line (is this backward-incompatible? if so I’ll make it keep the awful case),

yeah, let's not overdo it.:) Pretty sure that's ok.

mvz:
Capitalization improvements FTW!

PR #569, opened 2015-07-04
ObjectRefs: track lines for FeatureEnvy

### LINKED ISSUE #480: Last pending spec
I've talked to @chastell about the last pending spec from `spec/reek/smells/feature_envy_spec.rb` line 262 and how good it would be to get it working.

Unfortunately, it turns out making it work is beyond my abilities, so I'd just like to report the findings.

- from what I could tell, this test was first introduced in [this commit](https://github.com/troessner/reek/commit/1344c3662cc37028384a534cc43545352b1fd886#diff-7af6df8b186567d72140f27cb8f9ad5cR232)
- I've tracked that code through the upcoming commits and it turns out that code/feature was never passing tests. I think it was practically logged as a wanted feature.
- with a little bit of code inspection (and guessing), I concluded that the `#lines` was actually coming from [this line](https://github.com/troessner/reek/commit/aedf315d2f0fa95724a67554e2854cf804cbbf40#diff-3dc159b02e4482c773663260e2e12e5aR25) which in turn seems to be delegating to `parser` gem (at which point I gave up).

So, hopefully, this is helpful info!

chastell:
This makes `ObjectRefs` collect `ObjectRef` objects, which track the callsite lines, and makes `FeatureEnvy` use it to fix #480.

### DIFF
--- a/lib/reek/ast/node.rb
+++ b/lib/reek/ast/node.rb
@@ -34,7 +34,7 @@ module Reek
       end
 
       def line
-        loc.line
+        loc.line unless loc.nil?
       end
 
       # @deprecated
--- a/lib/reek/ast/object_refs.rb
+++ b/lib/reek/ast/object_refs.rb
@@ -1,33 +1,30 @@
 module Reek
+  # @api private
   module AST
+    ObjectRef = Struct.new(:name, :line)
     #
     # Manages and counts the references out of a method to other objects.
     #
-    # @api private
-    class ObjectRefs  # :nodoc:
+    class ObjectRefs
       def initialize
-        @refs = Hash.new(0)
+        @refs = Hash.new { |refs, name| refs[name] = [] }
       end
 
-      def record_reference_to(exp)
-        @refs[exp] += 1
+      def most_popular
+        max = @refs.values.map(&:size).max
+        @refs.select { |_name, refs| refs.size == max }
       end
 
-      def references_to(exp)
-        @refs[exp]
+      def record_reference_to(name, line: nil)
+        @refs[name] << ObjectRef.new(name, line)
       end
 
-      def max_refs
-        @refs.values.max || 0
-      end
-
-      def max_keys
-        max = max_refs
-        @refs.select { |_key, val| val == max }
+      def references_to(name)
+        @refs[name]
       end
 
       def self_is_max?
-        max_keys.length == 0 || @refs[:self] == max_refs
+        @refs.empty? || most_popular.keys.include?(:self)
       end
     end
   end
--- a/lib/reek/context/method_context.rb
+++ b/lib/reek/context/method_context.rb
@@ -56,7 +56,7 @@ module Reek
       end
 
       def envious_receivers
-        return [] if @refs.self_is_max?
+        return {} if @refs.self_is_max?
         @refs.max_keys
       end
 
--- a/lib/reek/context/method_context.rb
+++ b/lib/reek/context/method_context.rb
@@ -43,11 +43,11 @@ module Reek
         receiver ||= [:self]
         case receiver[0]
         when :lvasgn
-          @refs.record_reference_to(receiver.name)
+          @refs.record_reference_to(receiver.name, line: exp.line)
         when :lvar
-          @refs.record_reference_to(receiver.name) unless meth == :new
+          @refs.record_reference_to(receiver.name, line: exp.line) unless meth == :new
         when :self
-          @refs.record_reference_to(:self)
+          @refs.record_reference_to(:self, line: exp.line)
         end
       end
 
@@ -57,7 +57,7 @@ module Reek
 
       def envious_receivers
         return {} if @refs.self_is_max?
-        @refs.max_keys
+        @refs.most_popular
       end
 
       def references_self?
--- a/lib/reek/context/singleton_method_context.rb
+++ b/lib/reek/context/singleton_method_context.rb
@@ -8,7 +8,7 @@ module Reek
     # @api private
     class SingletonMethodContext < MethodContext
       def envious_receivers
-        []
+        {}
       end
     end
   end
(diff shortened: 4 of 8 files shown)


chastell:
Not rushing anything, just saying this is ready for review (and will enable me to fix #480, which is that last irritating skipped case on every `rake spec` run). ;)

lib/reek/ast/object_refs.rb
```diff
@@ -1,33 +1,30 @@
...
       def initialize
-        @refs = Hash.new(0)
+        @refs = Hash.new { |refs, name| refs[name] = [] }
       end
 
-      def record_reference_to(exp)
-        @refs[exp] += 1
+      def most_popular
+        max = @refs.values.map(&:size).max
+        @refs.select { |_name, refs| refs.size == max }
```
troessner: How about using Enumerable#max_by?

troessner: Ah, can be multiple entries, forget what i said above.

troessner:
Looks good, please squash and i'll merge!

lib/reek/ast/object_refs.rb
chastell: Yeah, it was super irritating. Went through all of the `enumerable` docs (always a good thing!) to make sure there isn’t a better way. :)

chastell:
Squashed down to two unrelated commits.

Rationale: if `(Singleton)MethodContext#envious_receivers` returning an empty `Hash` (rather than `Array`) in the edge cases turns out to be problematic, `git bisect` will show it right away. Squashing this together with the other commit means this change would have to be combed out of all the rest.

lib/reek/ast/object_refs.rb
```diff
@@ -1,33 +1,30 @@
 module Reek
-  module AST
+  module AST  # :nodoc:
```
mvz: What is this `# :nodoc:` supposed to accomplish? I don't see it having any effect on YARD's output.

lib/reek/ast/object_refs.rb
```diff
@@ -1,33 +1,30 @@
...
+    ObjectRef = Struct.new(:name)
```
mvz: OTOH, this shows up as part of our public API.

lib/reek/ast/object_refs.rb
chastell: Without this (or `# rubocop:disable Style/Documentation`) RuboCop says `AST` is not documented. This is probably a bug in RuboCop – it is triggered by the newly-added `ObjectRef`, and it seems RuboCop no longer treats `AST` as a namespace-only module.

I’m ok with any other way to stop RuboCop failing on undocumented `AST` that you prefer, I just want it to not be picky about it.

lib/reek/ast/object_refs.rb
chastell: Ugh, what’s the most acceptable way to making it not?

lib/reek/ast/object_refs.rb
troessner: We could put this rubocop exception in a rubocop config file so that it doesn't pollute our source code, couldn't we?

lib/reek/ast/object_refs.rb
mvz: Since everything under AST is also private, you can move the `# @api private` on line 7 up to above line 2.

lib/reek/ast/object_refs.rb
chastell: It seems bumping up `# @api private` fixed this. :)

lib/reek/ast/object_refs.rb
chastell: Oh cool, done – and we got rid of the other `:nodoc:` as well.

lib/reek/ast/object_refs.rb
```diff
@@ -1,33 +1,30 @@
...
-        @refs[exp] += 1
+      def most_popular
+        max = @refs.values.map(&:size).max
+        @refs.select { |_name, refs| refs.size == max }
       end
 
-      def references_to(exp)
-        @refs[exp]
+      def record_reference_to(name)
+        @refs[name] << ObjectRef.new(name)
```
troessner: Wait a sec.
ObjectRef just takes the same name. So if we reference, say, "foo" three times "@refs" would be (kind of pseudocode):

```Ruby
"foo" => [ ObjectRef("foo"), ObjectRef("foo"), ObjectRef("foo")]
````

So isn't this just a counter? If yes, couldn't we just drop this whole ObjectRef thingie and replace that via a plain integer count:


```Ruby
@refs[name] += 1
````

and change the initialization to:

```Ruby
@refs = Hash.new { |refs, name| refs[name] = 0 }
```

?

chastell: The counters is what we had previously – but for #480 we need to track line numbers, so the fix for that will have `ObjectRef`s that have a `:name` and a `:line` – and we need FeatureEnvy to get the distinct line numbers – see e0b75b7 + 769ae0b (will be rebased, of course).

troessner: How about adding "line" right away to make this less confusing?

chastell: I made this PR simply fix the issue, maybe this is the least confusing (sorry!). :)

chastell:
(I’d like to make `ObjectRefs#record_reference_to`’s signature to be `record_reference_to(name, line: nil)`, as I prefer kwargs to positional arguments for optional stuff, but for this I need #579 merged.) :)

chastell:
Squashed and rearranged the commits, green ready to merge. :)

lib/reek/ast/node.rb
```diff
@@ -34,7 +34,7 @@ def [](index)
       end
 
       def line
-        loc.line
+        loc.line unless loc.nil?
```
troessner: Why do we need to introduce a statement modifier now?

chastell: Because `loc` is `nil` for cases where we send things like `s(:send, s(:self), :thing)` to `MethodContext#record_call_to`. We seem to do this only in specs (in four places), but I didn’t want to break this use case.

mvz:
Nice!

PR #571, opened 2015-07-04
Output HTMLReport to stdout and move its template

### LINKED ISSUE #570: Write HTMLReport to STDOUT
....I guess we all agree on that, right?.:)
Would be breaking, so labelled with breaking-change-for-reek-4.

chastell:
Fixes #570.


chastell:
The build fails because current JRuby is compatible with Ruby 1.9, which didn’t have `__dir__`. I hope JRuby 9.0.0.0 will get released before `reek` 4 and we’ll be able to simply bump the JRuby version requirement (to match our Ruby 2.0+ support).

troessner:
Nice!

But shouldn't this point to a branch called reek-4 and not to master?

chastell:
> But shouldn't this point to a branch called reek-4 and not to master?

I don’t know, is that our post-gitflow process now? :)

I assumed everything points to master and this will just get merged once master becomes 4.0 material and reek-3 is branched off of it (for potential maintenance releases). But I’m very open to any other process. :)

mvz:
> I assumed everything points to master and this will just get merged once master becomes 4.0 material and reek-3 is branched off of it.

I think that makes the most sense.

troessner:
But then we should have a reek 3 branch right away, right?
I am at the beach today but can do that later unless somebody wants to jump in..;)

mvz:
> But then we should have a reek 3 branch right away, right?

No, we just don't merge this one in. Let's not complicate things by having main development (which I think is still for reek 3) in a different branch.

troessner:
And this where things get complicated and obscure: all of a sudden you cant click merge on a PR because it would introduce a breaking change. So we have to label our PRs like fanatics and mire important, pay attention all the time. 
The potential for errors and confusion here is high.
Also just leaving all breaking changes flying around in PR seems harder to maintain to me than having one dedicated branch. Or how do we handle merge conflicts across all "breaking-changes" pull requests? Those PR will all be affected which makes iur overview very messy.

I strongly vote for not doing it like this but rather have a proper reek-3 branch set up with master being the inofficial reek 4.

chastell:
> all of a sudden you cant click merge on a PR because it would introduce a breaking change.

We can always close such PRs unmerged and reopen them once master is the target release.

> So we have to label our PRs like fanatics

Oh this is actually always good (but I prefer using GitHub milestones for version tracking). I’ve been doing this across projects and you can easily work on open PRs targeting reek 3 only if you’re worried you’ll misclick the merge button. ;)

> Or how do we handle merge conflicts across all "breaking-changes" pull requests?

This is a very valid point; a common branch for reek 4 code would address this.

> I strongly vote for not doing it like this but rather have a proper reek-3 branch set up with master being the inofficial reek 4.

The question is whether it should be `master` for reek 4 and `reek-3` for reek 3 or `master` for reek 3 and `reek-4` for reek 4.

The former (`master` is always the most advanced, i.e., reek 4) is the easier to maintain; as soon as we have breaking changes we branch the previous major release before merging them.

The latter (`master` is current development, i.e., reek 3) is the easier to open PRs against (unless we make `reek-3` the default branch for now) and addresses the problem of ‘the only think for reek 4 we have now is an insignificant, mostly-æsthetic fix so @chastell feels bad he triggered this discussion so soon for such a trivial reason’.

TL;DR: I think our options are:

1. Close this and reopen once we start _serious_ work on reek 4.
2. Designate `master` as reek 4 material, branch `reek-3` off of it and make `reek-3` the default branch. _Please_ let’s merge all open PRs before we do this, I hate the close+reopen dance GitHub makes us do to retarget PRs, as this makes the PR histories chopped across many PRs.
3. Branch `reek-4` and merge this there.

I think the above is in order of my preference, but I’m ok with either, really. :)

mvz:
My vote is for 1, then 3.

mvz:
@chastell maybe you can make a separate PR for moving the template, since that can be merged in immediately.

troessner:
But [1] would limit us mentally. What is serious work on reek4 and what not? We should be able to work on all things all the time and our workflow should invite that.
Lets go with [3] - ok for you @chastell ?

chastell:
I’m ok with 3 (`master` for reek 3, `reek-4` for reek 4). Closing this one, will reopen two (template vs output) against the right targets in a few. :)

PR #575, opened 2015-07-05
Move report option mapping into Report

mvz:
Re-targeted version of #552.

- Delegate option mapping from OptionInterpreter to Report
- Demand a valid option to be passed instead of relying on defaults

### DIFF
--- a/lib/reek/cli/option_interpreter.rb
+++ b/lib/reek/cli/option_interpreter.rb
@@ -1,8 +1,6 @@
 require 'forwardable'
 require_relative 'input'
-require_relative '../report/report'
-require_relative '../report/formatter'
-require_relative '../report/heading_formatter'
+require_relative '../report'
 
 module Reek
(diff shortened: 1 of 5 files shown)


troessner:
Very nice! Just one last thing that i find decreases our overall readability and is in general a little annoying: The "api private" comments that are everywhere. 
Can't we just mark whole files as api private?

mvz:
> Can't we just mark whole files as api private?

In general, yes, but if I do that for `Report`, YARD will consider `TextReport` etc. private as well.

What I could do is move these methods to some class under `Report` instead. WDYT?

troessner:
What I could do is move these methods to some class under Report instead. WDYT?

But then we would structure our code according to our documentation, not the other way round.:)
If YARD sucks that much then maybe we should switch to another tool?
That would / could be a discussion for another thread though.
Merging this one.

Issue #573, opened 2015-07-05
IrresponsibleModule triggers when a class defined in multiple files does not have descriptive comments for all files


beraboris:
When a class is defined in multiple files, IrresponsibleModule will trigger unless all the files have a descriptive comment for that class. This makes no sense. As long as one of the files has a descriptive comment, then it should be good enough.

Defining a class in multiple files is generally uncommon. There is one case where defining a class in multiple files is common.

Let's say you write a gem named `foobar`. In `lib/foobar.rb` you define the `Foobar` class. In `lib/foobar/stuff.rb` you define the `Foobar::Stuff` class. To define `Foobar::Stuff`, you need to reopen the `Foobar` class.

Here's an example:

```
$ ls
bar.rb  foo.rb
$ cat foo.rb 
# Some descritive comment
class Foobar
  # Foo has a comment
  class Foo; end
end
$ cat bar.rb 
class Foobar
  # bar has a comment
  class Bar; end
end
$ reek --version
reek 3.0.3
$ reek *
bar.rb -- 1 warning:
  [1]:Foobar has no descriptive comment (IrresponsibleModule)
1 total warning
```

mvz:
Reek's architecture is not suitable for analysis across files. However, we could limit IrresponsibleModule to not trigger when a class or module is only there as a container for other classes or modules. I.e., no methods are defined in the class or module itself. So, the following would not trigger this smell anymore, regardless of `Foobar` being commented anywhere.

```ruby
class Foobar
  # Foo has a comment
  class Foo; end
end
```

chastell:
> we could limit IrresponsibleModule to not trigger when a class or module is only there as a container for other classes or modules

:+1: – this is in line with ‘let’s not require comments for namespace-only modules’.

mvz:
Fixed in #576.

Issue #485, opened 2015-04-26
IrresponsibleModule does not work on modules


mpapis:
```ruby
module Something
  def do_something
    true
  end
end
```
this does not give warnings about missing comments, there should be at least a comment on the method, but the module could be good too - especially if it was not hacking and the module had some extra purpose

mvz:
I think IrresponsableModule only works on classes currently.

Perhaps we should remove this odd 'smell' altogether and let RuboCop handle this kind of issue?

mvz:
@troessner, @chastell what's your opinion on this?

chastell:
I’m probably a bad person to be asked about this (and bad person in general), as I have both reek’s IrresponsibleModule and RuboCop’s Style/Documentation turned off in my projects. ;)

I think this is a bigger question of whether
* reek should recognise RuboCop and actively try to complement it (and not double RuboCop’s checks), and – orthogonally – whether
* reek should flag this kind of problems as smells.

I think RuboCop and reek work great together, but I don’t mind them flagging the same problems. I’m not sure whether lack of comments is strictly a _code_ smell, though – as in, something that a static code analyser like reek should care about.

chastell:
> I think IrresponsableModule only works on classes currently.

About this: I think this is because modules double as namespaces in Ruby, so if this was to also be a module check then it would have to make an educated guess whether it’s a namespace or not. Potential heuristics:
* top-level module is a namespace, the rest are not, or
* a module with methods is not a namespace.

(I’m still not sure this should be a reek smell in the first place, though.)

troessner:
>> reek should recognise RuboCop and actively try to complement it (and not double RuboCop’s checks), and – orthogonally – whether

I think we shouldn't care about rubocop. It's a great gem, I love it and I use it extensively but we should do what we want to do without having it as a dependency. This doesn't mean that we shouldn't keep it's features in mind of course. And "this exists already in rubocop" can always be a good argument. But it shouldn't be a decisive one.

>> Perhaps we should remove this odd 'smell' altogether and let RuboCop handle this kind of issue?

Veto, like explained above.:)

>> I’m probably a bad person to be asked about this (and bad person in general), as I have both reek’s IrresponsibleModule and RuboCop’s Style/Documentation turned off in my projects. ;)

Heretic!

>> I think IrresponsableModule only works on classes currently.

I'd vote for making it work on modules as well.

troessner:
I think we should make IrresponsibleModule work on modules as well.
[...]

PR #576, opened 2015-07-05
Improve IrresponsibleModule

mvz:
Two commits, fixing two issues: #485 and #573.

### DIFF
--- a/lib/reek/context/module_context.rb
+++ b/lib/reek/context/module_context.rb
@@ -20,6 +20,17 @@ module Reek
       def descriptively_commented?
         CodeComment.new(exp.leading_comment).descriptive?
       end
+
+      # A namespace module is a module (or class) that is only there for namespacing
+      # purposes, and thus contains only nested constants, modules or classes.
+      #
+      # However, if the module is empty, it is not considered a namespace module.
+      #
+      # @return true if the module is a namespace module
+      def namespace_module?
+        contents = exp.children.last
+        contents && contents.find_nodes([:def, :defs], [:class, :module]).empty?
+      end
     end
   end
 end
--- a/spec/reek/smells/irresponsible_module_spec.rb
+++ b/spec/reek/smells/irresponsible_module_spec.rb
@@ -66,6 +66,80 @@ RSpec.describe Reek::Smells::IrresponsibleModule do
     expect(src).to reek_of :IrresponsibleModule, name: 'Foo::Bar'
   end
 
+  it 'does not report modules used only as namespaces' do
+    src = <<-EOS
+      module Foo
+        # Describes Bar
+        class Bar
+          def baz
+          end
+        end
+      end
+    EOS
+    expect(src).not_to reek_of(:IrresponsibleModule)
+  end
+
+  it 'does not report classes used only as namespaces' do
+    src = <<-EOS
+      class Foo
+        # Describes Bar
+        module Bar
+          def qux
+          end
+        end
+      end
+    EOS
+    expect(src).not_to reek_of(:IrresponsibleModule)
+  end
+
+  it 'reports modules that have both nested modules and methods' do
+    src = <<-EOS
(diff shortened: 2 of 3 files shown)


spec/reek/smells/irresponsible_module_spec.rb
```diff
@@ -61,6 +66,80 @@ class BadClass
     expect(src).to reek_of :IrresponsibleModule, name: 'Foo::Bar'
   end
 
+  it 'does not report modules used only as namespaces' do
```
chastell: I don’t think this describes the spec, which seems to be about not reporting a documented class.

spec/reek/smells/irresponsible_module_spec.rb
```diff
@@ -61,6 +66,80 @@ class BadClass
...
+        end
+        # Describes Bar
+        module Bar
+        end
+      end
+    EOS
+    expect(src).to reek_of(:IrresponsibleModule)
+  end
+
+  it 'foo' do
```
chastell: I love foo, but I’d rather have a better description here. :older_man:

lib/reek/context/module_context.rb
```diff
@@ -20,6 +20,11 @@ def node_instance_methods
       def descriptively_commented?
         CodeComment.new(exp.leading_comment).descriptive?
       end
+
+      def namespace_module?
+        return unless (contents = exp.children.last)
+        !contents.find_nodes([:def, :defs], [:class, :module]).any?
+      end
```
chastell: I’m not a huge fan of assignment-in-conditional(-in-parens-to-side-step-RuboCop), nor of `!…any?`. How about:

```Ruby
def namespace_module?
  contents = exp.children.last
  contents && contents.find_nodes([:def, :defs], [:class, :module]).empty?
end
```

spec/reek/smells/irresponsible_module_spec.rb
mvz: Oops. This whole spec should go.

lib/reek/context/module_context.rb
mvz: Yes, that's much nicer.

spec/reek/smells/irresponsible_module_spec.rb
chastell: Hm, I’m not sure that it shuld go – I think it would be beneficial to have explicit spec saying whether comment-less
```Ruby
class CompilationError < RuntimeError; end
```
and
```Ruby
CompilationError = Class.new(RuntimeError)
```
smell of IrresponsibleModule or not.

mvz: Ok. I'll leave it in. The behavior for your second example is undefined. Perhaps that should be the subject of another ticket?

chastell: Oh sure, I don’t have an opinion on this – just wanted to point out there are two syntaxes with the same behaviour, so it might be worth specing them explicitly. :)

mvz: I made a new ticket that will lead to us being able to handle such syntax: #583.

mvz:
Updated and rebased.

lib/reek/context/module_context.rb
```diff
@@ -20,6 +20,11 @@ def node_instance_methods
       def descriptively_commented?
         CodeComment.new(exp.leading_comment).descriptive?
       end
+
+      def namespace_module?
```
troessner: Could we add a short comment here giving an example of what "exp" could look like and / or what we expect it is?

mvz: Makes sense.

mvz: Comment added.

chastell:
:ok_hand: Merging!

PR #577, opened 2015-07-05
Move (and rename) HTMLReport template closer to HTMLReport

chastell:
This allows us to drop the `assets` directory altogether.


troessner:
Following what we discussed in https://github.com/troessner/reek/pull/571, shouldn't this PR point to the reek-4 branch (which I created 30 seconds ago)?

chastell:
As per @mvz’s [comment](https://github.com/troessner/reek/pull/571#issuecomment-118639908) this is the part of that PR that is backwards-compatible and should go into reek 3. :)

PR #579, opened 2015-07-05
Target JRuby 9.0.0.0 (rather than JRuby 1.7)

chastell:
Ok, this one might be controversial, but here goes: I think reek 3 should support JRuby 9.0.0.0 (even though it’s not released yet) rather than JRuby 1.7.

In my opinion one of the main reasons we dropped Ruby 1.9 support in reek 3 was to be able to use Ruby 2.0 syntax, from keyword arguments to things like `__dir__`. JRuby 1.7 supports only Ruby 1.9 syntax (and features), while JRuby 9.0.0.0 supports Ruby 2.2 syntax (and features) – so if we want to support JRuby 1.7 in reek 3 then using Ruby 2.0 features would have to wait for reek 4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,8 @@ rvm:
   - ruby-head
 matrix:
   include:
-    - rvm: jruby-19mode
-      env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
+    - rvm: jruby
+      env: JRUBY_OPTS='--2.0 --server -Xcompile.invokedynamic=false'
     - rvm: jruby-head
       env: JRUBY_OPTS='--server -Xcompile.invokedynamic=false'
   allow_failures:


troessner:
Yes, I agree with this change.
@mvz ?

mvz:
I agree we shouldn't test on JRuby in 1.9 mode. Perhaps we can use the trick shown [here](https://github.com/lotus/controller/pull/85/files) to force 2.0 mode?

chastell:
@mvz that just tests reek against JRuby’s head, which is farther than current 9.0.0.0. I agree that testing against JRuby’s head is a good thing, but I wouldn’t fail the build when the tip of JRuby’s head happens to be unstable (as it sometimes is). Note that we already test against `jruby-head`, which is currently 9.0.0.0-rc1 + some commits (almost 9.0.0.0-rc2).

mvz:
Sorry, I wasn't being clear. I meant using this:

```yaml
matrix:
  include:
    - rvm: jruby
      env: JRUBY_OPTS="--2.0"
```

chastell:
Ah, good point, I didn’t know about this switch – and it does seem to make JRuby support both `__dir__` and kwargs. :) Adjusted and green, please merge!

mvz:
Excellent!

PR #585, opened 2015-07-09
Improve Context classes and TreeWalker

mvz:
- Don't use `method_missing` to bubble up method calls
- Record statement counting and call recievers everywhere
- Make RootContext know about its AST
- Narrow down TreeWalker's exposed set of methods to just `#walk`
- Move `#default_assignments` method to MethodContext
- Use full code example for `config_for` spec

PR #590, opened 2015-07-10
Handle modifier style use of module_function for UtilityFunction

### LINKED ISSUE #582: Recognize singleton methods defined when module_function is a scoping modifier
Follow-up issue to #567 

I'll look into this.

mvz:
This fixes #582.

To be able to track calls to module_function in all cases, these calls are handled during the walking of the tree by TreeWalker. Parent contexts track the currently set visibility and pass it on to any new child contexts. If module_function is used with arguments, the parent context will instead modify the visibility of existing children.

In order to have the final visibility at hand while detecting smells, the smell detection is done after the entire tree has been walked.

Next steps:
* Move smell detection phase out of TreeWalker
* Change Attribute to use this same mechanism.

### DIFF
--- a/lib/reek/context/code_context.rb
+++ b/lib/reek/context/code_context.rb
@@ -14,23 +14,26 @@ module Reek
     class CodeContext
       attr_reader :exp
       attr_reader :num_statements
+      attr_reader :children
+      attr_reader :visibility
 
       # Initializes a new CodeContext.
       #
-      # context - *_context from the `core` namespace
-      # exp - Reek::Source::ASTNode
+      # @param context [CodeContext, nil] The parent context
+      # @param exp [Reek::AST::Node] The code described by this context
       #
-      # Examples:
+      # For example, given the following code:
       #
-      # Given something like:
+      #   class Omg
+      #     def foo(x)
+      #       puts x
+      #     end
+      #   end
       #
-      #   class Omg; def foo(x); puts x; end; end
+      # The {TreeWalker} object first instantiates a {RootContext}, which has no parent.
       #
-      # the first time this is instantianted from TreeWalker `context` is a RootContext:
-      #
-      #   #<Reek::Context::RootContext:0x00000002231098 @name="">
-      #
-      # and `exp` looks like this:
+      # Next, it instantiates a {ModuleContext}, with +context+ being the
+      # {RootContext} just created, and +exp+ looking like this:
       #
       #  (class
       #    (const nil :Omg) nil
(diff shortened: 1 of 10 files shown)


mvz:
Now that #585 has been merged this is ready for review.

chastell:
Looks good, clean and understandable to me (and is most welcome, as I’m a huge user of `module_function` and will be able to clean up my `.reek` files substantially after this is released), but I’ll defer the final review and merge to @troessner. :)

troessner:
Gonna review this bad boy tomorrow on my way to work.:)

lib/reek/context/code_context.rb
```diff
@@ -14,10 +14,12 @@ module Context
     class CodeContext
       attr_reader :exp
       attr_reader :num_statements
+      attr_reader :children
+      attr_reader :visibility
 
       # Initializes a new CodeContext.
       #
-      # context - *_context from the `core` namespace
+      # context - XContext from the `Context` namespace
```
troessner: Why the "X"? How about 

>> context - One of the contexts from the `Context` namespace

?

lib/reek/tree_walker.rb
```diff
@@ -6,12 +6,15 @@
...
   #
-  # Traverses a Sexp abstract syntax tree and fires events whenever
-  # it encounters specific node types.
+  # Traverses an abstract syntax tree and fires events whenever it encounters
+  # specific node types.
   #
   # SMELL: This class is responsible for counting statements and for feeding
   # each context to the smell repository.
   #
+  # TODO: Make TreeWalker responsible only for creating Context objects, and
```
troessner: Meaning we would have 3 tree traversal for each run, TreeDresser, TreeWalker and ???

spec/reek/smells/utility_function_spec.rb
```diff
@@ -73,11 +73,56 @@ def simple(arga)
...
+        expect(src).not_to reek_of(:UtilityFunction)
+      end
+
+      it 'does not report when module_function modifier is called in separate scope' do
+        src = <<-EOF
+          class C
+            begin
+              module_function
+            end
+            def m(a) a.to_s; end
```
troessner: As always, nice specs!

troessner:
Reviewed, this looks awesome. Just one small thing: In order to fully get what's going on here I had to read the specs, just from the code it wasn't obvious to me. Not sure if that is a good or a bad sign though.:D

lib/reek/tree_walker.rb
mvz: Well, yes and no. The final traversal will be a traversal over contexts, not the tree.

mvz:
I improved the documentation and rebased.

I hope the documentation makes things clearer, but maybe the code can still be improved. Is there any particular part you find unclear? In any case, we will probably revisit this code soon and improve it further.

lib/reek/tree_walker.rb
troessner: Ok, but we'll still have:

1.) Traverse AST to adorn it with sexp extensions
2.) Traverse it again to generate a tree of contexts
3.) Traverse the context tree

Right?

troessner:
Looks good to me, merged!

lib/reek/tree_walker.rb
mvz: Yes. I'm not too happy with it, but I don't see a clear path at the moment to make things simpler.

PR #591, opened 2015-07-11
Actually reset AppConfiguration

chastell:
HOW ABOUT A HAIKU TRIBUTE TO HALF AN HOUR OF DEBUGGING

singleton config
blows up simple tests which pass
in isolatiooon
:tada:

simpsons-fire

### DIFF
--- a/lib/reek/configuration/app_configuration.rb
+++ b/lib/reek/configuration/app_configuration.rb
@@ -52,6 +52,7 @@ module Reek
 
         def reset
           @configuration.clear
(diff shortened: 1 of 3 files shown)


troessner:
Yes, I agree. Let's create separate app_configuration objects for all-the-things. 
Might be a bigger refactoring though.

chastell:
I opened #594 to track this; I’ll be happy to take a look once I’m done with pathnaming all the things (which, in turn, triggered this issue). :)

PR #593, opened 2015-07-11
Make smells configurable on a directory base.

### LINKED ISSUE #594: Don’t use singleton configuration
State-holding singletons are problematic, [configuration ones doubly so](https://github.com/troessner/reek/pull/591). Let’s turn `AppConfiguration` into a class and use it instead of the singleton, injecting where necessary (this won’t be trivial, but should be doable).

troessner:
Ready for review! Pinging  @mvz && @chastell 

This PR implements directory specific smell configuration like this:

```Yaml
"web_app/app/controllers":
  NestedIterators:
    enabled: false
"web_app/app/helpers":
  UtilityFunction:
    enabled: false
```

Please see the updated README and the file features/configuration_files/directory_specific_directives.feature (in this PR) for an overview.

As part of this, I'd like to introduce some new terminology into reek, the "directive", which is a specific configuration. 
Either for a given directory or for "everything else" which would be the "default directive" (so basically the global smell configuration from before this PR).

This PR furthermore:

* fixes #594
* implements the rails friendly mode (#529) as a side effect since this mode is just a set of directory specific configurations
* is the precondition for the codeclimate engine (#548) since a lot (most?) of codeclimates ruby projects are rails projects and 
  most people would disable reek quickly when they got spammed by NestedIterators (controllers) and UtilityFunction (helpers)

Regarding the specs, @mvz suggested something like

```Ruby
expect('class Foo; end').to reek_of(:DataClump).with_config(...)
```

but I couldn't find any "proper" documentation about it. I mean, if you search for "rspec custom settings" and so on there are a few hits for sure, but none of them seems to be a stable, solid way of doing this. 

We could implement this anyways for sure, but I'd rather go with the solution in this pull request since it's obvious and straight-forward what's going on.


features/configuration_files/masking_smells.feature
```diff
@@ -3,20 +3,6 @@ Feature: Masking smells using config files
   As a developer
   I want to mask some smells using config files
 
-  Scenario: empty config file is ignored
```
troessner: Deleted as a suggestion in this PR - I can re-introduce it if we decide so.
I deleted it because to me, this feature doesn't make much sense. An empty config file shouldn't cause any errors / warnings at all.
E.g. for quick tests I sometimes use the pattern:

```Bash
echo "" > config.reek
reek ...
git co . # for git checkout .
```

And then this check is highly annoying. But it's not about my usage pattern, it's more about that we should warn if necessary, not just on all-the-things.
WDYT?

lib/reek/cli/application.rb
```diff
@@ -21,7 +22,7 @@ def initialize(argv)
         begin
           @options = options_parser.parse
           @command = ReekCommand.new(OptionInterpreter.new(@options))
-          initialize_configuration
+          @configuration = Configuration::AppConfiguration.new @options
```
troessner: As a next step we should get rid of the option hell in /cli which always trips me up. There's the OptionInterpreter, the OptionParser and that options OpenStruct. That's just madness.:)

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,111 @@ module Reek
...
     #
     # @api private
-    module AppConfiguration
-      NON_SMELL_TYPE_KEYS   = %w(exclude_paths)
-      EXCLUDE_PATHS_KEY     = 'exclude_paths'
-      @configuration        = {}
-      @has_been_initialized = false
+    class AppConfiguration
+      EXCLUDE_PATHS_KEY = 'exclude_paths'
+      attr_reader :exclude_paths, :default_directive, :directory_directives
```
troessner: There's the terminology I'm reerring to in the PR description.

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,111 @@ module Reek
...
-          if path.size.zero?
-            report_problem('Empty file', path)
-            return
-          end
+      # @param source_via [String] - the source of the code inspected.
+      # Might be a string, STDIN or from file. We're only interested in the source
+      # when it's coming from file, since a directory_directive doesn't make sense
+      # for anything else
+      # @return [Hash | nil] the configuration for the source or nil
+      def directory_directive_for_source(source_via)
```
troessner: This is called "description" everywhere else which I think is horribly misleading.
We should change that everywhere but this PR was already big, so I'd rather do this separately.

lib/reek/configuration/configuration_file_finder.rb
```diff
@@ -37,6 +42,20 @@ def find_by_dir(start)
           return found if found
         end
       end
+
+      def load_from_file(path)
```
troessner: This was on AppConfiguration before where it doesn't belong IMHO.

lib/reek/examiner.rb
```diff
@@ -21,13 +21,16 @@ class Examiner
...
+    # @param filter_by_smells [Array<String>]
     #   List of smell types to filter by.
     #
-    def initialize(source, smell_types_to_filter_by = [])
-      @source       = Source::SourceCode.from(source)
-      @collector    = CLI::WarningCollector.new
-      @smell_types  = eligible_smell_types(smell_types_to_filter_by)
+    def initialize(source,
+                   filter_by_smells = [],
+                   configuration: Configuration::AppConfiguration.new)
```
troessner: API change, but not a breaking one.

lib/reek/examiner.rb
```diff
@@ -63,16 +66,18 @@ def smelly?
     private
 
     def run
-      smell_repository = Smells::SmellRepository.new(description, @smell_types)
+      smell_repository = Smells::SmellRepository.new(source_description: description,
```
troessner: One of the most confusing usages of "description" for sure.

lib/reek/smells/smell_repository.rb
```diff
@@ -9,29 +9,37 @@ module Smells
...
-        smell_types.each do |klass|
-          @detectors[klass] = klass.new(source_description)
+      def initialize(source_description: nil,
+                     smell_types: self.class.smell_types,
+                     configuration: Configuration::AppConfiguration.new)
+        self.source_description = source_description
+        self.typed_detectors    = nil
+        self.detectors          = {}
+        self.configuration      = configuration
+        self.smell_types        = smell_types
```
troessner: Not directly part of this PR but according to the boyscout rule I also reduced the usage of instance variables and preferred proper setters.

lib/reek/spec.rb
```diff
@@ -85,8 +85,10 @@ module Spec
     #
     #   expect(src).to reek_of(Reek::Smells::DuplicateMethodCall, name: '@other.thing')
     #
-    def reek_of(smell_category, smell_details = {})
-      ShouldReekOf.new(smell_category, smell_details)
+    def reek_of(smell_category,
+                smell_details = {},
+                configuration = Configuration::AppConfiguration.new)
```
troessner: API change, but none breaking. @mvz suggested to go with something like:

```Ruby
expect('class Foo; end').to reek_of(:DataClump).with_config(...)
```

But I couldn't find anything obvious regarding this topic. Since it's not breaking, we can go with this solution and then maybe improve this later.

spec/reek/tree_walker_spec.rb
```diff
@@ -2,38 +2,6 @@
...
-  @to_sql += case
-    when block_given? then yield.to_sql
-    else args.to_sql
-  end
-  self
-end
-EOS
-    expect(src).not_to reek
-  end
-end
```
troessner: Those specs didn't make any sense here and are superfluous as well AFAIS.

features/configuration_files/masking_smells.feature
mvz: Agreed, en empty file should be fine.

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,108 @@ module Reek
...
+      #
+      # @param options [OpenStruct]
+      #   e.g.: #<OpenStruct report_format=:text,
+      #                      location_format=:numbers,
+      #                      colored=true,
+      #                      smells_to_detect=[],
+      #                      config_file=#<Pathname:config/defaults.reek>,
+      #                      argv=["lib/reek/spec"]>
+      def initialize(options = nil)
+        self.directory_directives = {}
```
mvz: We never finished our discussion about this style, did we :smile:?

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,108 @@ module Reek
...
+      end
 
-        def raise_error(reason, path)
-          raise ConfigFileException, message(reason, path)
+      def handle_directory_directive(path, config)
+        pathname = Pathname.new path.chomp('/')
+        raise ArgumentError, 'Configuration error:'\
+          "No smell or path for #{pathname}" unless pathname.exist?
+        directory_directives[pathname] = config.each_with_object({}) do |(key, value), hash|
+          hash[Reek::Smells.const_get(key)] = value
```
mvz: Shouldn't we somehow use `try_to_recognize_smell_type` here to avoid hard crashes?

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,108 @@ module Reek
...
+      def handle_default_directive(klass, config)
+        default_directive[klass] = config
+      end
 
-        def raise_error(reason, path)
-          raise ConfigFileException, message(reason, path)
+      def handle_directory_directive(path, config)
+        pathname = Pathname.new path.chomp('/')
+        raise ArgumentError, 'Configuration error:'\
+          "No smell or path for #{pathname}" unless pathname.exist?
```
mvz: I would prefer to avoid the line continuation character by making this a multiline `unless`.

features/configuration_files/directory_specific_directives.feature
```diff
@@ -0,0 +1,171 @@
+Feature: Directory directives
```
mvz: Given our recent discussion about Cucumber, I think this should end up being the only configuration scenario, because it is the most complex example.

lib/reek/smells/smell_repository.rb
```diff
@@ -9,29 +9,37 @@ module Smells
     #
     # @api private
     class SmellRepository
-      attr_reader :detectors
+      attr_reader :detectors, :typed_detectors, :configuration, :source_via, :smell_types
```
mvz: Are we accessing all these values outside of this class? If not, I don't think we should expand this class' API by adding these readers.

lib/reek/source/source_locator.rb
```diff
@@ -8,14 +8,16 @@ module Source
     #
     # @api private
     class SourceLocator
+      attr_reader :paths, :configuration
```
mvz: Are we reading these values outside of the class?

spec/reek/tree_walker_spec.rb
mvz: Agreed :+1:

lib/reek/spec.rb
mvz: Agreed. I opened a follow-up ticket: #606.

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,108 @@ module Reek
...
-            return
-          end
+      # @param source_via [String] - the source of the code inspected.
+      # Might be a string, STDIN or Filename / Pathname. We're only interested in the source
+      # when it's coming from file since a directory_directive doesn't make sense
+      # for anything else.
+      # @return [Hash | nil] the configuration for the source or nil
+      def directory_directive_for_source(source_via)
+        return unless source_via
+        return unless Source::SourceCode.source_from_file?(source_via)
```
mvz: We can just leave this out I think? The code below will be just fine if `"stdin"` or `"string"` is passed in.

lib/reek/smells/smell_repository.rb
```diff
@@ -9,29 +9,37 @@ module Smells
...
+                     configuration: Configuration::AppConfiguration.new)
+        self.source_via         = source_description
+        self.typed_detectors    = nil
+        self.detectors          = {}
+        self.configuration      = configuration
+        self.smell_types        = smell_types
+
+        initialize_detectors
+        directive = configuration.directory_directive_for_source(source_via) ||
+          configuration.default_directive
```
mvz: I think we should allow merging directives, e.g., if a smell is enabled in the default directive, and configured further in a directory directive.

lib/reek/smells/smell_repository.rb
```diff
@@ -42,12 +50,20 @@ def examine(scope)
...
+        smell_types.each do |klass|
+          detectors[klass] = klass.new(source_via)
+        end
+      end
+
       def smell_listeners
-        unless @typed_detectors
-          @typed_detectors = Hash.new { |hash, key| hash[key] = [] }
-          @detectors.each_value { |detector| detector.register(@typed_detectors) }
+        unless typed_detectors
```
mvz: This was basically a nice memoization pattern, but adding writers makes it more complicated.

lib/reek/smells/smell_repository.rb
mvz: Perhaps we can move the `||` logic to AppConfiguration and just do `directive = configuration.directive_for(source_via)`.

lib/reek/smells/smell_repository.rb
```diff
@@ -42,12 +50,20 @@ def examine(scope)
 
       private
 
+      attr_writer :detectors, :typed_detectors, :configuration, :source_via, :smell_types
+
+      def initialize_detectors
```
mvz: Since `detectors` is now a method, we can move this code in there.

lib/reek/examiner.rb
mvz: I've always wondered why SmellRepository had a description parameter. Now finally we have some kind of use for it.

spec/reek/configuration/app_configuration_spec.rb
```diff
@@ -13,72 +14,61 @@
...
+        }
+      end
+      let(:source_via) { 'foo/bar/bang/dummy.rb' }
 
-      expect(described_class.configuration).to eq(sample_configuration_loaded)
-      described_class.reset
-      expect(described_class.configuration).to eq(default_configuration)
-      expect(described_class.exclude_paths).to be_empty
+      it 'returns the corresponding directive' do
+        allow(subject).to receive(:directory_directives).and_return directory_directives
```
mvz: I would prefer to just pass the relevant configuration to the initializer.

spec/reek/configuration/app_configuration_spec.rb
```diff
@@ -13,72 +14,61 @@
...
+      let(:baz_config)  { { Reek::Smells::IrresponsibleModule => { enabled: false } } }
+      let(:bang_config) { { Reek::Smells::Attribute => { enabled: true } } }
 
-  describe '.reset' do
-    it 'resets the configuration' do
-      described_class.load_from_file(sample_configuration_path)
+      let(:directory_directives) do
+        {
+          Pathname.new('foo/bar/baz')  => baz_config,
+          Pathname.new('foo/bar/bang') => bang_config
```
mvz: What happens if we have configuration for `foo/bar`?

mvz:
I think this is the way to go in general. We should really finish our fundamental discussion about getters and setters versus instance variables, but I'm fine with delaying that until later.

I'd like to see a little more specs regarding overlapping paths and other corner cases.

@chastell any thoughts?

lib/reek/configuration/app_configuration.rb
troessner: No, we didn't :smile:

lib/reek/configuration/app_configuration.rb
troessner: Good point, I'll kill it.

lib/reek/configuration/app_configuration.rb
troessner: :+1:

lib/reek/configuration/app_configuration.rb
troessner: Hmm, I did that on purpose:
`try_to_recognize_smell_type` is used for the top level where you either have a smell type or a directory and it shouldn't raise on a directory.
Meaning this allows us to do:

```Yaml
IrresponsibleModule:
  # ....
'foo/bar':
  # 

Here, however, we know that we are in an directory directive. This means that here we can and should IMHO help the user since we now **know** that we're definitely talking about smell types and thus can detect errors like this:

```Yaml
'foo/bar':
  IrresponsibleMModule: # Note the double M
    # ....
```

I believe we should keep it this way.

lib/reek/source/source_locator.rb
troessner: Removing them.

features/configuration_files/directory_specific_directives.feature
troessner: I agree. But let's refactor our specs and features separately, not in this PR.

lib/reek/smells/smell_repository.rb
troessner: Good point, changed.

lib/reek/smells/smell_repository.rb
troessner: >> I think we should allow merging directives, e.g., if a smell is enabled in the default directive, and configured further in a directory directive.

I disagree. Then we would have (again) a rather obscure configuration mechanism where the user often would not be entirely sure of what's happening and where we'd have to give lengthy examples of how to use that. The way it is right now, it's simple and easy
: 
- You're in directory for which there is a directive -> this gets used.
- You're not: default directive is used.

I strongly suggest to not re-introduce such a behaviour again. We had this back in the day where we allowed multiple configuration files and back then we decided that the complexity that came along with it was not necessary.
Not to mention that the assumption was (and still is for me), that no user would use this mechanism but rather have a little configuration duplication.

troessner: >> Perhaps we can move the || logic to AppConfiguration and just do directive = configuration.directive_for(source_via).

Good idea, changed!

lib/reek/smells/smell_repository.rb
troessner: I don't follow - in where? The getter or the setter? Both places don't make much sense to me...

spec/reek/configuration/app_configuration_spec.rb
troessner: What do you mean? `directory_directives=` is private and the initializer for AppConfiguration requires the unholy option struct with a pathname. I'm not against changing this code to be more flexible, I'd love it if we 1.) refactored the options hell and 2.) made app_configuration so flexible that we can pass in different things and it still works out.
However, I'd like to do that in a follow-up pull request, since this is pretty big. For [1] there is already an issue. I'll open up another for [2].

troessner: Addendum: See #608

spec/reek/configuration/app_configuration_spec.rb
troessner: Whatever comes first for "detect" here: https://github.com/troessner/reek/pull/593/files#diff-5a8a1c2b1f6f3e41dfe85414f9c43266R69
Excellent question. Wondering how we could solve this.
How about adapting the linked code to not just use plain pattern matching but to look for the closest pattern?

troessner:
>> I'd like to see a little more specs regarding overlapping paths

Agree, but let's first discuss how we handle overlapping paths at all. 
To quote myself:

>>
https://github.com/troessner/reek/pull/593/files#diff-5a8a1c2b1f6f3e41dfe85414f9c43266R69
How about adapting the linked code to not just use plain pattern matching but to look for the closest pattern?

lib/reek/configuration/app_configuration.rb
mvz: Right. I agree, but we should perhaps have a better error message than the `NameError` we would get now.

lib/reek/smells/smell_repository.rb
mvz: > The way it is right now, it's simple and easy:
> 
> * You're in directory for which there is a directive -> this gets used.
> * You're not: default directive is used.

Yes, this makes perfect sense. :+1:

lib/reek/smells/smell_repository.rb
mvz: Into the getter as a memoization pattern.

lib/reek/configuration/app_configuration.rb
troessner: How about "You are trying to configure smell type #{....} but we can't find one with that name. Please make sure you spelled it right. In doubt, check 'config/defaults.reek' in the reek repository for a list of all available smell types."
?

lib/reek/smells/smell_repository.rb
troessner: >> Into the getter as a memoization pattern.

Done and cleaned up afterwards.

troessner:
Let's continue discussing in not-code specific comments, responding to github's outdated diff comments is an incredible pain in the ass.

From what I can see the following is left to do (i believe I addressed all other comments):

- The error message when loading nested smells:
  "You are trying to configure smell type #{....} but we can't find one with that name. Please make sure you spelled it right. In doubt, check 'config/defaults.reek' in the reek repository for a list of all available smell types." and spec'ing that
- handling the overlapping paths. See above
- spec more edge cases (suggestions?)

spec/reek/configuration/app_configuration_spec.rb
mvz: I'm always a little worried when mocking the object under test, but we can leave it until we've improved the initializer.

spec/reek/configuration/app_configuration_spec.rb
mvz: > How about adapting the linked code to not just use plain pattern matching but to look for the closest pattern?

Sounds good.

mvz:
Yes:
* Nested smells: seems the right way to go.
* Overlapping paths: Let's pick the most specific as you suggested.
* Edge cases:
  * Overlapping paths in both orders of keys
  * Full file path including the file name

features/configuration_files/masking_smells.feature
chastell: Agreed, let’s allow it.

lib/reek/smells/smell_repository.rb
chastell: Did it yield a significant performance improvement? (Not saying it wasn’t nice or was unnecessary, just wondering.)

README.md
```diff
@@ -144,42 +144,63 @@ of how many `*.reek` files you might have on your filesystem.
...
+# A lot of smells allow fine tuning their configuration. You can look up all available options
+# in the corresponding smell documentation in /docs. In most cases you probably can just go
+# with the defaults we set in config/defaults.reek.
 DataClump:
   max_copies: 3
   min_clump_size: 3
 
+### Directory specific configuration
+
+# You can configure smells on a directory base.
```
chastell: ‘…on a directory basis’? ‘…on a per-directory basis’?

README.md
```diff
@@ -144,42 +144,63 @@ of how many `*.reek` files you might have on your filesystem.
...
+# E.g. the classical rails case: controllers smell of NestedIterators (see /docs/Nested-Iterators.md) and
```
chastell: classical rails → classic Rails

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,115 @@ module Reek
...
+      #     enabled: false
+      # exclude_paths:
+      #   - "app/controllers"
+      #
+      # this would result in the following configuration:
+      #
+      # exclude_paths = [Pathname('spec/samples/two_smelly_files')]
+      # default_directive = { Reek::Smells::IrresponsibleModule=>{"enabled"=>false} }
+      # directory_directives = { Pathname("spec/samples/three_clean_files/") =>
+      #                          {Reek::Smells::UtilityFunction=>{"enabled"=>false}}}
```
chastell: I know it borders on minutiæ, but can we either keep or drop spaces inside the braces and around the hashrockets here?

(I understand this was copy-’n’-pasted from irb, but I’d rather have a RuboCop-kosher syntax with spaces – or at least not have them everywhere rather than this mish-mash.)

lib/reek/configuration/app_configuration.rb
chastell: We also have

```Ruby
@configuration = Configuration::AppConfiguration.new @options
```

in this PR. :smile: 

(My personal preference, as mentioned before, is to use instance variable asignments in the constructor and getters/setters everywhere else; for immutable objects – which I highly prefer – this approach means no `self.` anywhere.)

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,115 @@ module Reek
...
-          raise_error('Not a hash', path) unless @configuration.is_a? Hash
-        end
+      # @param source_via [String] - the source of the code inspected
+      # Might be a string, STDIN or Filename / Pathname. We're only interested in the source
+      # when it's coming from file since a directory_directive doesn't make sense
+      # for anything else.
+      # @return [Hash | nil] the configuration for the source or nil
+      def directory_directive_for_source(source_via)
+        return unless source_via
+        base_path = Pathname.new(source_via).split[0]
```
chastell: `Pathname#split#[0]` is `Pathname#dirname`. :)

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,115 @@ module Reek
...
+        hit = directory_directives.keys.detect { |pathname| pathname.to_s =~ /#{base_path}/ }
+        directory_directives[hit] unless hit.nil?
```
chastell: How does this exactly work (i.e., why the regexp match)? Can you show an example of a match?

(The regexp match of two `Pathnames` cast to `Strings`, especially without the `base_path` going through `Regexp.escape`, worries me.)

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,115 @@ module Reek
...
-        end
+      def load(options)
+        configuration_file = ConfigurationFileFinder.find_and_load(options: options)
 
-        def exclude_paths
-          @exclude_paths ||= begin
-            @configuration.fetch(EXCLUDE_PATHS_KEY, []).map do |string|
-              Pathname.new(string.chomp('/'))
+        configuration_file.each do |key, value|
+          case key
```
chastell: How about:

```Ruby
case key
when EXCLUDE_PATHS_KEY
  handle_exclude_paths(value)
when Reek::Smells.const_defined?(key)
  handle_default_directive(key, value)
else
  handle_directory_directive(key, value)
end
```

lib/reek/configuration/app_configuration.rb
```diff
@@ -5,89 +5,115 @@ module Reek
...
+        pathname = Pathname.new path.chomp('/')
+        unless pathname.exist?
+          raise ArgumentError, "Configuration error: No smell or path for #{pathname}"
         end
-
-        def raise_error(reason, path)
-          raise ConfigFileException, message(reason, path)
+        directory_directives[pathname] = config.each_with_object({}) do |(key, value), hash|
+          hash[Reek::Smells.const_get(key)] = value
         end
```
chastell: `#each_with_object` always trips me up a bit; how about:

```Ruby
config = config.map { |key, val| { Reek::Smells.const_get(key) => val } }.reduce(:merge)
directory_directives[pathname] = config
```

or

```Ruby
config = Hash[config.map { |key, val| [Reek::Smells.const_get(key), val] }]
directory_directives[pathname] = config
```

(`Hash.[]` would have to be used for Ruby 2.0 compatibility, as `Array#to_h` was added in Ruby 2.1.)

lib/reek/source/source_code.rb
```diff
@@ -38,9 +41,9 @@ def initialize(code, description, parser = Parser::Ruby22)
       def self.from(source)
         case source
         when File     then new(source.read, source.path)
-        when IO       then new(source.readlines.join, 'STDIN')
+        when IO       then new(source.readlines.join, IO_IDENTIFIER)
         when Pathname then new(source.read, source.to_s)
-        when String   then new(source, 'string')
+        when String   then new(source, STRING_IDENTIFIER)
         end
```
chastell: Is this really better? We don’t use `IO_IDENTIFIER` nor `STRING_IDENTIFIER` anywhere else, so all this change means in practice is ‘I’ll have to look up what these two are every time I read the `SourceCode.from` method’ and ‘every time I encounter `'STDIN'` and I’ll grep for it I’ll find `SourceCode::IO_IDENTIFIER` and will have to grep for that next’.

chastell:
I agree on the stuff above (overlapping, edge cases). I made a few nitpics, but like this otherwise.

Apologies for derailing you with the `SAMPLES_PATH` commit! :)

lib/reek/configuration/app_configuration.rb
troessner: >> We also have
@configuration = Configuration::AppConfiguration.new @options
in this PR. :smile:

Yes, I only changed it where it was actually possible without being inconsistent within one file, at least i tried :smile: 
We should agree on one style and then change that in lib/ in one big pull request.

lib/reek/configuration/app_configuration.rb
troessner: Good catch!

lib/reek/configuration/app_configuration.rb
troessner: I need to restructure this in any case to handle overlapping paths correctly, let's postpone this discussion until then.

lib/reek/source/source_code.rb
troessner: >> Is this really better?

I believe so.

>> We don’t use IO_IDENTIFIER nor STRING_IDENTIFIER anywhere else

True, but for me it's about intention revealing code in this case. For me the intention is "we use this literal for identifying xyz and this is will always be like that"
So for me it's about clarity and the detour, at least in this case, doesn't hurt that badly IMHO.

On a sidenote: 
I believe we should get rid of all magic strings and numbers in our source code, but this would be a different discussion.

lib/reek/configuration/app_configuration.rb
troessner: >> case
when key == EXCLUDE_PATHS_KEY
  handle_exclude_paths(value)
when Reek::Smells.const_defined?(key)
  handle_default_directive(key, value)
else
  handle_directory_directive(key, value)
end

Great idea! I needed to go with:

```Ruby
          when smell_type?(key)
            handle_default_directive(key, value)
```

with 

```Ruby
      def smell_type?(key)
        Reek::Smells.const_get key
      rescue NameError
        false
      end
```

because of the quite annoying way "const_defined?" is implemented:

```Bash
[1] pry(main)> Reek::Smells.const_defined? 'asdfas'
NameError: wrong constant name asdfas
from (pry):1:in `const_defined?'
```

But that does the job as well.

lib/reek/configuration/app_configuration.rb
troessner: >> #each_with_object always trips me up a bit;

I disagree - I believe, each_with_object is almost always a cleaner solution compared to reduce/inject.
Matz seems to agree, otherwise they probably wouldn't have introduced each_with_object to begin with :smile: 
Since that is highly subjective I'd rather leave it like that.

lib/reek/source/source_code.rb
mvz: How about createing a slightly different initializer for SourceCode which makes the object itself more aware of the origin of its source? This would make it much easier to have different code paths for each case later on as well, instead of relying on the string contents of the description (This is of course definitely something for a separate PR).

lib/reek/configuration/app_configuration.rb
chastell: Yeah, let’s not block this PR with this discussion and have a separate clean-up PR.

lib/reek/configuration/app_configuration.rb
chastell: Ok, let’s leave this in. I’ll moan about preferring `each.with_object` (`Enumerator` composition) to `each_with_object`, but that’s so minor I won’t. ;)

lib/reek/source/source_code.rb
chastell: @troessner: fair enough. :)

@mvz: I was thinking just that yesterday when I was pondering on #604, but the more I tried to describe the `Source` object (that has an origin and can be read from, and is subclassed for the different subtypes), the more it matched the `SourceCode` object we already have, so maybe we can make that a bit more elastic?

chastell:
Thanks for the fixes, @troessner – I’m happy with this PR as-is (with the further fixes postponed to future PRs). @mvz, feel free to merge if you’re ok with this.

troessner:
Let me improve the pattern matching part and add the additional specs from above and then LGTM..;)

I'll ping you guys then for re-review.

troessner:
@mvz && @chastell ready for re-review!
I put all changes into a separate second commit to make it easier to review: https://github.com/troessner/reek/commit/b717d1179e22bc158fb011d18260ce75575f720a

**What I did:**

* Abort on non-existent directory
* Abort on non-existent smell in directory directive
* Abort on file as directory directive
* Specs and feature improvements
* handling the overlapping paths in feature directives to use the most specific one

Let me know if this works for you - then I'll squash my 2 commits into one and the LGTM :)

chastell:
One teeny-tiny nitpick. LGTM!

mvz:
:+1:

mvz:
Oh, and :shipit: after fixing @chastell's nit.

troessner:
>> One teeny-tiny nitpick. LGTM!

Done and squashed to one commit!
Could somebody please hit the red.....errr, green button?.:)

troessner:
I'd release 3.1. right away - objections?

mvz:
No objections. Awesome!

troessner:
Aaaaaaaand .....3.1 is out..:)

PR #595, opened 2015-07-12
Handle casgn style class definitions

mvz:
Introduces a new ModuleContext for classes defined by constant assignment, like so:

```ruby
Foo = Class.new do
  def bar(a, b)
  end
end
```

This means smells on methods inside the class definition will have the correct context reported.

Also makes IrresponsibleModule detect when such classes have no comment. Other smells still have to be adjusted to consider :casgn contexts.

This makes reek handle the following sample code (from #341) correctly:

```ruby
# Reports the wrong context (A instead of B)
module A
  B = Struct.new(:c) do
    def initialize
    end

    def foo(a, b)

    end
  end
end
```

### DIFF
--- a/lib/reek/ast/sexp_extensions.rb
+++ b/lib/reek/ast/sexp_extensions.rb
@@ -180,6 +180,10 @@ module Reek
         def arg_names
           args.map { |arg| arg[1] }
         end
+
+        def object_creation_call?
+          method_name == :new
+        end
       end
 
       Op_AsgnNode = SendNode
@@ -302,6 +306,7 @@ module Reek
         end
 
         include MethodNodeBase
+
         def full_name(outer)
           prefix = outer == '' ? '' : "#{outer}#"
           "#{prefix}#{SexpFormatter.format(receiver)}.#{name}"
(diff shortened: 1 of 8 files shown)


mvz:
Ready for review.

lib/reek/ast/sexp_extensions.rb
```diff
@@ -391,14 +391,8 @@ def simple_name
...
       end
 
-      # Utility methods for :module nodes.
-      module ModuleNode
-        def name() self[1] end
-
-        def simple_name
-          name.is_a?(::Parser::AST::Node) ? name.simple_name : name
-        end
-
```
troessner: Removing code from the completely overloaded sexp_extensions.rb === awesome.

lib/reek/ast/sexp_extensions.rb
```diff
@@ -409,12 +403,49 @@ def text_name
...
 
+      # Utility methods for :module nodes.
+      module ModuleNode
+        include ModuleNodeBase
+        def name() self[1] end
+
+        def simple_name
+          name.is_a?(::Parser::AST::Node) ? name.simple_name : name
+        end
+      end
```
troessner: Damn, I relished to early.:-/

lib/reek/ast/sexp_extensions.rb
```diff
@@ -409,12 +403,49 @@ def text_name
         end
       end
 
+      # Utility methods for :module nodes.
+      module ModuleNode
+        include ModuleNodeBase
+        def name() self[1] end
```
troessner: Wait, that's deprecated, isn't it? Should be "children[0]", shouldn't it?

lib/reek/ast/sexp_extensions.rb
```diff
@@ -409,12 +403,49 @@ def text_name
...
+            MODULE_DEFINERS.include?(call.receiver.simple_name)
+        end
+
+        def name
+          children[1]
+        end
+
+        def value
+          children.last
+        end
```
troessner: Very nice. Getting rid of our bad abstractions step by step.

lib/reek/ast/sexp_extensions.rb
```diff
@@ -409,12 +403,49 @@ def text_name
         end
       end
 
+      # Utility methods for :module nodes.
+      module ModuleNode
+        include ModuleNodeBase
+        def name() self[1] end
+
+        def simple_name
+          name.is_a?(::Parser::AST::Node) ? name.simple_name : name
```
troessner: I know it's legacy code but can we please document somewhere (here?) with a speaking example what the difference between "name" and "simple_name" is? And then get rid of it asap.
A "name" that has an attribute called "simple_name" is just wrong on so many levels.:)

lib/reek/ast/sexp_extensions.rb
```diff
@@ -409,12 +403,49 @@ def text_name
...
+
+        def defines_module?
+          call = case value.type
+                 when :block
+                   value.call
+                 when :send
+                   value
+                 end
+          call &&
+            call.method_name == :new &&
```
troessner: We have a guard clause like quite a few times in our code base. Could we move this into one speaking method called "initializer?" or something like that?

spec/reek/smells/irresponsible_module_spec.rb
```diff
@@ -140,6 +152,37 @@ class Foo < Bar; end
...
+  it 'does not report constants that are not classes' do
+    src = <<-EOS
+      module Qux
+        Foo = 23
+        Bar = Hash.new
+        Quuz = 'foo'.freeze
+      end
+    EOS
+    expect(src).not_to reek_of(:IrresponsibleModule)
+  end
```
troessner: Nice specs!

troessner:
Very nice!
Just a couple of minor comments and then let's get this on the road!

mvz:
Updated and rebased.

troessner:
Nice one, merged!

PR #596, opened 2015-07-12
Use Pathnames where possible

### LINKED ISSUE #555: Switch to Pathnames where possible

chastell:
This:
* uses `__dir__` wrapped in a `Pathname` instead of `File.expand_path` (where possible),
* makes `AppConfiguration.load_from_file` take a `Pathname` instead of a `String` (is this a backward-incompatible change? I can make it take either…),
* makes `AppConfiguration.exclude_paths` return `Array<Pathname>`,
* makes `SourceLocator#sources` return `Array<Pathname>`,
* uses `Kernel.Pathname` (rather than `Pathname.new`) consistently throughout the codebase.

Once we drop Ruby 2.0 and 2.1 support we can start using the nicer-looking `Pathname#/` instead of `Pathname#+` and `Pathname#write` instead of `File.write`.

Fixes #555.

### DIFF
--- a/features/step_definitions/sample_file_steps.rb
+++ b/features/step_definitions/sample_file_steps.rb
@@ -26,7 +26,7 @@ Given(/^a smelly file with inline masking called 'inline.rb'$/) do
 end
 
 Given(/^the "(.*?)" sample file exists$/) do |file_name|
-  full_path = File.expand_path file_name, 'spec/samples'
+  full_path = Pathname("#{__dir__}/../../spec/samples/#{file_name}")
   in_current_directory { FileUtils.cp full_path, file_name }
 end
 
@@ -135,7 +135,7 @@ When(/^I run "reek (.*?)" in the subdirectory$/) do |args|
 end
 
 Given(/^a masking configuration file in the HOME directory$/) do
-  set_env('HOME', File.expand_path(File.join(current_directory, 'home')))
+  set_env 'HOME', Pathname("#{__dir__}/../../#{current_directory}/home").to_s
   write_file('home/config.reek', <<-EOS.strip_heredoc)
     ---
     DuplicateMethodCall:
--- a/lib/reek/report/report.rb
+++ b/lib/reek/report/report.rb
@@ -1,5 +1,6 @@
-require 'rainbow'
 require 'json'
+require 'pathname'
+require 'rainbow'
 require_relative 'formatter'
 require_relative 'heading_formatter'
 
@@ -129,9 +130,9 @@ module Reek
     class HTMLReport < Base
       require 'erb'
 
-      def show(target_path = 'reek.html')
-        template_path = File.expand_path('../html_report.html.erb', __FILE__)
-        File.write target_path, ERB.new(File.read(template_path)).result(binding)
+      def show(target_path = Pathname('reek.html'))
+        template_path = Pathname("#{__dir__}/html_report.html.erb")
+        File.write target_path, ERB.new(template_path.read).result(binding)
         puts 'HTML file saved'
       end
     end
--- a/lib/reek/source/source_code.rb
+++ b/lib/reek/source/source_code.rb
@@ -18,7 +18,7 @@ module Reek
       # Initializer.
       #
       # code        - ruby code as String
-      # description - in case of STDIN this is "STDIN" otherwise it's a filepath as String
+      # description - 'STDIN', 'string' or a filepath as String
       # parser      - the parser to use for generating AST's out of the given source
       def initialize(code, description, parser = Parser::Ruby22)
         @source      = code
@@ -27,8 +27,8 @@ module Reek
       end
 
       # Initializes an instance of SourceCode given a source.
-      # This source can come via 3 different ways:
-      # - from files a la `reek lib/reek/`
+      # This source can come via 4 different ways:
+      # - from Files or Pathnames a la `reek lib/reek/`
       # - from IO (STDIN) a la `echo "class Foo; end" | reek`
       # - from String via our rspec matchers a la `expect("class Foo; end").to reek`
       #
@@ -37,9 +37,10 @@ module Reek
       # @return an instance of SourceCode
       def self.from(source)
         case source
-        when File   then new(source.read, source.path)
-        when IO     then new(source.readlines.join, 'STDIN')
-        when String then new(source, 'string')
+        when File     then new(source.read, source.path)
+        when IO       then new(source.readlines.join, 'STDIN')
+        when Pathname then new(source.read, source.to_s)
+        when String   then new(source, 'string')
         end
       end
 
--- a/lib/reek/source/source_locator.rb
+++ b/lib/reek/source/source_locator.rb
@@ -1,4 +1,5 @@
 require 'find'
+require 'pathname'
 
 module Reek
   module Source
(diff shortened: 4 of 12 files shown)


features/step_definitions/sample_file_steps.rb
```diff
@@ -26,7 +26,7 @@ def a
 end
 
 Given(/^the "(.*?)" sample file exists$/) do |file_name|
-  full_path = File.expand_path file_name, 'spec/samples'
+  full_path = Pathname(__dir__) + '../../spec/samples' + file_name
```
troessner: Hmmm. I really dislike the "+" for concatenation. Could we use string interpolation here and below? I believe thats also more idiomatic in regards to our code base.

chastell: I wasn’t sure whether there isn’t an edge-case where `Pathname` would escape something properly that concatenation wouldn’t, but I couldn’t come up with anything so I switched them all to interpolated Strings. :)

lib/reek/source/source_locator.rb
```diff
@@ -11,60 +12,56 @@ class SourceLocator
       #
       # paths - a list of paths as Strings
       def initialize(paths)
-        @pathnames = paths.
-          map { |path| Pathname.new(path.chomp('/')) }.
-          flat_map { |pathname| current_directory?(pathname) ? pathname.entries : pathname }
+        @paths = paths.map(&method(:Pathname)).flat_map do |path|
```
troessner: I know i merged the PR, but after thinking about i just came back to say that I believe this is bad. Its valid ruby for sure but incredibly obscure. I mean, quick, how exactly do you call that feature? Right, no idea..;)

WDYT about bejng rather explicit here by just using plain and boring "regular" map?

lib/reek/source/source_code.rb
```diff
@@ -37,9 +37,10 @@ def initialize(code, description, parser = Parser::Ruby22)
       # @return an instance of SourceCode
       def self.from(source)
         case source
-        when File   then new(source.read, source.path)
-        when IO     then new(source.readlines.join, 'STDIN')
-        when String then new(source, 'string')
+        when File     then new(source.read, source.path)
+        when IO       then new(source.readlines.join, 'STDIN')
+        when Pathname then new(source.read, source.to_s)
```
mvz: I didn't really pay much attention to this PR, but perhaps I should have.
I'd really like to not add more options to `.from`. I find the fact that Pathname is also apparently something to read from rather confusing; It is sort of similar to File, but not really.

lib/reek/source/source_locator.rb
```diff
@@ -11,60 +12,56 @@ class SourceLocator
...
+          current_directory?(path) ? path.entries : path
+        end
       end
 
       # Traverses all paths we initialized the SourceLocator with, finds
       # all relevant ruby files and returns them as a list.
       #
-      # @return [Array<File>] - Ruby files found
+      # @return [Array<Pathname>] - Ruby paths found
```
mvz: Can't we go back to returning an Array of File here?

lib/reek/source/source_locator.rb
mvz: Or even just a one-stage `flat_map` without the chaining?

features/step_definitions/sample_file_steps.rb
```diff
@@ -135,7 +135,7 @@ def initialize
 end
 
 Given(/^a masking configuration file in the HOME directory$/) do
-  set_env('HOME', File.expand_path(File.join(current_directory, 'home')))
+  set_env 'HOME', Pathname("#{__dir__}/../../#{current_directory}/home").to_s
```
mvz: I'm a little confused by this change. File.expand_path uses the working directory, which is conveniently set by Aruba.  `__dir__`, however, refers to the current file's directory. This means this will now break if Aruba moves its temporary directories elsewhere, or if we move the step definitions.

Also, why not use Pathname.join? Or `Pathname(File.expand_path ... )`?

lib/reek/report/report.rb
```diff
@@ -129,9 +130,9 @@ def show
     class HTMLReport < Base
       require 'erb'
 
-      def show(target_path = 'reek.html')
-        template_path = File.expand_path('../html_report.html.erb', __FILE__)
-        File.write target_path, ERB.new(File.read(template_path)).result(binding)
+      def show(target_path = Pathname('reek.html'))
+        template_path = Pathname("#{__dir__}/html_report.html.erb")
```
mvz: I prefer `Pathname.join(__dir__, 'html_report.html.erb')`.

spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -16,44 +16,44 @@
 
     it 'returns the file in current dir if config_file is nil' do
       options = double(config_file: nil)
-      current = Pathname.new('spec/samples')
+      current = Pathname('spec/samples')
```
mvz: I suppose it's a matter of taste, but I'm not very fond of methods that look like constants, so I actually prefer `Pathname.new` here.

spec/reek/report/xml_report_spec.rb
```diff
@@ -27,8 +28,8 @@
     end
 
     it 'prints non-empty checkstyle xml' do
-      sample_path = File.expand_path 'checkstyle.xml', 'spec/samples'
-      expect { instance.show }.to output(File.read(sample_path)).to_stdout
+      sample_path = Pathname("#{__dir__}/../../samples/checkstyle.xml")
```
mvz: Perhaps we can create a helper method in `spec_helper.rb` that gives us paths in the samples directory, so we don't have to mentally handle the `..`?

lib/reek/source/source_locator.rb
chastell: Oh sure thing, the fix is here: https://github.com/troessner/reek/pull/598. :)

chastell: @mvz: Good point, sorry I missed that – GitHub didn’t refresh the comments after you wrote yours. I’ll fix along the other things you mentioned.

lib/reek/source/source_code.rb
chastell: Hm, to me `Pathname` is a representation of a path, so reading from (and writing to) it makes sense to me.

I agree the fewer options to `SourceCode.from` the better, but I also believe `Pathname` is strictly superior to `String` for representation of paths. I’m reluctant to drop `String` from here as `Examiner#initialize` calls `Source::SourceCode.from(source)` directly on whatever it gets as its first parameter, and – for better or worse ;) – we expose `Examiner.new` as part of our current API. Accepting both (for now) here seems like a lesser evil.

lib/reek/source/source_locator.rb
chastell: We can, but `SourceLocator` taking paths, analysing and filtering them, then returning the right ones seems like a class with a single responsibility dealing with a collection of one other type. Why would `Files` be better here (other than ‘this is what it returned before’)?

spec/reek/configuration/configuration_file_finder_spec.rb
chastell: Ok, I’ll switch to that syntax. :)

lib/reek/source/source_code.rb
mvz: The `String` option here is strictly for strings that contain the source code itself, so that is a different case.

How about making it the responsibility fo `.from`'s caller (i.e., `CLI::Input`) to convert the `Pathname` to a `File`?

mvz: Sorry, I meant making it the reponsibility of the caller of `SourceLocater#sources` to convert the result to `File`, so that `CLI::Input#sources` always returns a list of `IO` objects.

lib/reek/source/source_locator.rb
mvz: Yes, this is a good application of the SRP. See my other comment on letting `CLI::Input` convert the result into `File` objects.

features/step_definitions/sample_file_steps.rb
chastell: Argh, good catch, apologies – in all the other places switching to `__dir__` meant the code will work regardless of whether it was ran in reek’s checkout dir or not and I got a bit too over-zealous. Will fix!

lib/reek/report/report.rb
chastell: We have three options:

1. `Pathname.new(__dir__).join('html_report.html.erb')` – your preference
1. `Pathname.new("#{__dir__}/html_report.html.erb")` – @troessner’s preference
1. `Pathname.new(__dir__) + 'html_report.html.erb'` – the one I used initially (but I’m ok with either)

Ruby 2.2 adds `Pathname.new(__dir__) / 'html_report.html.erb'` but we can’t use it yet. :)

If you two could figure out which one you want I’ll change all places accordingly. :)

spec/reek/report/xml_report_spec.rb
chastell: Good point, I’ll add it. :)

lib/reek/report/report.rb
mvz: I wish Pathname did a little bit more its best to be like File and allowed `Pathname.join('foo', 'bar')`. None of these look very appealing, but let's go with 2. for now since that's what it is now.

lib/reek/source/source_code.rb
chastell: Oh, `IO` objects, I can get behind that. `IO` protocol in Ruby is one of the parts I really like. :)

chastell:
@mvz: I made all-but-`IO` fixes in #600. I’ll do the ‘`SourceCode.from` should just take `IO` objects’ in a separate PR. Thanks for the post-merge review!

PR #598, opened 2015-07-13
SourceLocator: drop Method#to_proc’s cleverness

chastell:
This might not be as readable as it should, [as per this comment](https://github.com/troessner/reek/pull/596#discussion_r34425701), so let’s drop it. :)


troessner:
I just found another good reason for dropping it:

https://github.com/JuanitoFatas/fast-ruby/blob/master/code/general/block-apply-method.rb

;)

chastell:
I look forward to the point in time when `SourceLocator#initialize` becomes reek’s speed bottleneck. ;)

PR #600, opened 2015-07-16
Pathname switch fixes

chastell:
Four changes requested by @mvz in a post-merge review of #596:

* `Pathname` → `Pathname.new` switch,
* proper expansion of `$HOME` for Aruba-based features,
* single `flat_map` call in `SourceLocator#initialize`,
* switching all specs to (once-computed) `SAMPLES_PATH`.

I’ll do the ‘`SourceCode.from` should just take `IO` objects’ in a separate PR.

As these are four fairly disjunct fixes I made four commits; let me know if that’s too many because `<%= reasons %>`. ;)


mvz:
:+1:

Issue #483, opened 2015-04-26
Assigning class variables smell


mpapis:
two use cases:

1. assign instance variable directly without a writer (`@something = something`) - I'm actually not sure about this one
2. assigning instance variable in methods - example:

    ```ruby
    attr_reader :something
    def methodA(something)
      @something = something
      methodB
    end
    def methodB
      puts something
    end
    ```
    it might be prefered to set up instance variables in `initialize` so any refactoring would not run the code into missing variables

#stopMichalPapis

mvz:
Can you elaborate a bit more about why these would be smelly? I think point 2 would be caught by ruby's warnings if @something wasn't assigned yet. Point 1 I think conflicts with the Attribute smell, where we're trying to avoid writers (which I think is the better practice).

mpapis:
ah so point 1. can be ignored

I did not get any warnings for point 2. if there is an extra work to get those warnings then maybe it's better to also detect it as not everybody would know about it and would rely on `reek` to provide the best practices not running ruby with extra flags

troessner:
Just so I get this right:

@mpapis you're voting for:

- using and setting instance variables only in initializers 
- when updating them in other methods writer should be used instead of instance variables

right?

andyw8:
I read a interesting post on this recently:
http://designisrefactoring.com/2015/03/29/organizing-data-self-encapsulation/

troessner:
Just read it, interesting, thanks @andyw8 !

troessner:
Well, from what I can see a there are 3 possibilities:

[1] DirectAccess smell
[2[ Mixed Access smell
[3] SettingInstanceVariablesDirectly smell 

Not sure if I would go for [1] and [2], that seems a little overzealous to me.
What do you think about [3]?

mvz:
Yes, that was a very interesting article, @andyw8. I'm partial to Mixed access myself, and I think the problems he shows with adding a `last_name` setter for `Doctor` can be avoided by being consistent with `Person` and setting `@last_name` in the initializer.

Be that as it may, @troessner, what would SettingInstanceVariablesDirectly detect?

mvz:
@mpapis, I think Ruby would emit a warning at runtime if you call `methodB` before `methodA`. However, I do agree that it's nicer to write the code so this temporal dependency doesn't exist.

So, we don't want, e.g.,

```ruby
def methodA(something)
  @something = something
  methodB
end

def methodB
  puts @something
end
```

But we may want to allow:

[...]

PR #610, opened 2015-07-21
Use private_attr to silence ‘private attribute?’ warnings and refactor SmellRepository a bit

chastell:
Current master is no longer warning-clean, as it started throwing warnings about private attributes (don’t get me started how wrong the warning is).

This makes this warning go away by using the private_attr gem. This also includes some related refactorings of `SmellRepository`.

### DIFF
--- a/lib/reek/configuration/app_configuration.rb
+++ b/lib/reek/configuration/app_configuration.rb
@@ -1,4 +1,5 @@
 require 'pathname'
+require 'private_attr/everywhere'
 require_relative './configuration_file_finder'
 
 module Reek
@@ -40,9 +41,9 @@ module Reek
       #                      config_file = #<Pathname:config/defaults.reek>,
       #                      argv = [ "lib/reek/spec" ]>
       def initialize(options = nil)
-        self.directory_directives = {}
-        self.default_directive    = {}
-        self.exclude_paths        = []
+        @directory_directives = {}
+        @default_directive    = {}
+        @exclude_paths        = []
 
         load options
       end
@@ -56,7 +57,7 @@ module Reek
 
       private
 
-      attr_writer :exclude_paths, :default_directive, :directory_directives
+      private_attr_writer :exclude_paths
 
       # @param source_via [String] - the source of the code inspected
       # Might be a string, STDIN or Filename / Pathname. We're only interested in the source
--- a/lib/reek/smells/smell_repository.rb
+++ b/lib/reek/smells/smell_repository.rb
@@ -1,3 +1,4 @@
+require 'private_attr/everywhere'
 require_relative '../smells'
 require_relative 'smell_detector'
 require_relative '../configuration/app_configuration'
@@ -16,10 +17,10 @@ module Reek
       def initialize(source_description: nil,
                      smell_types: self.class.smell_types,
                      configuration: Configuration::AppConfiguration.new)
-        self.source_via         = source_description
-        self.typed_detectors    = nil
-        self.configuration      = configuration
-        self.smell_types        = smell_types
+        @source_via      = source_description
+        @typed_detectors = nil
+        @configuration   = configuration
+        @smell_types     = smell_types
 
         configuration.directive_for(source_via).each do |klass, config|
           configure klass, config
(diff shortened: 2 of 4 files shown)


troessner:
I find the warning highly annoying as well, but I'm not the biggest fan of polluting our code with code that is specific to test runs.
Unfortunately I don't have a better solution to offer.
@mvz what's your take on this?

chastell:
The only good thing is that it will finally go away in Ruby 2.3 (and I’ll be first to pull private_attr out once that is released): https://github.com/ruby/ruby/commit/32a5a098e39e377350e4696d2d9f2d8930f1ec41

mvz:
Uhm, I hate to say this, but this ties in with our discussion in #483; I don't really use attribute writers and always try to make my classes lazily immutable by setting instance variables in only one location.

So, my solution would be to replace the private setters with ivar assignment.

chastell:
But the problem is not with private setters, but with any private accessors (including getters).

mvz:
Ah, I see some of them are accessors and not just setters. Since they are not actually mutated in the class the writer part is not needed and we can just use the ivar set in the initializer, right?

mvz:
> we can just use the ivar set in the initializer

In the end, I would prefer a reader here, but since we don't subclass much I think we can deal with just using the ivar here.

I suppose protected attributes don't trigger a warning?

chastell:
Oh, good catch – protected attributes do not warn. I’m much more into making them protected than falling back to ivar access. :)

chastell:
(I’m happy to fix this PR as needed if/when we have a consensus, of course.)

mvz:
Ok, so I would opt for:
* Getting rid of the setters altogether
* Making the getters protected

chastell:
Ready for review – I made the getters protected and turned the setters into ivars (they happen to be used in constructors and for memoization only).

I also refactored two `SmellRepository` methods (hence separate commits). Let me know if you don’t like the refactorings, as readability is a subjective thing. :)

lib/reek/configuration/app_configuration.rb
```diff
@@ -86,7 +84,7 @@ def load(options)
       end
 
       def handle_exclude_paths(paths)
-        self.exclude_paths = paths.map do |path|
+        @exclude_paths = paths.map do |path|
```
troessner: I don't get it. In https://github.com/troessner/reek/issues/612 we had agreed on "use instance vars in the constructor and for memoization and getters / setters for everything else", hadn't we?

lib/reek/smells/smell_repository.rb
```diff
@@ -43,25 +42,21 @@ def examine(scope)
...
-          end
-          @detectors
-        end
+        @initialized_detectors ||= smell_types.map do |klass|
+          { klass => klass.new(source_via) }
+        end.reduce({}, :merge)
       end
 
-      private
+      protected
```
troessner: Strong veto. We just do this to silence a warning we only see when we run the spec?
Sorry, but for me this is a bad change. SmellRepository does not have any children. Making this protected is highly misleading then since I'd naturally assume it does.

lib/reek/configuration/app_configuration.rb
chastell: Ah, right. This is what we get for not having immutable objects. ;) Fixed.

lib/reek/smells/smell_repository.rb
chastell: > Strong veto.

Ok, when you two agree on how we should deal with the warnings please let me know and I’ll update this PR accordingly. :)

> We just do this to silence a warning we only see when we run the spec?

No, to silence warnings we otherwise throw at anyone who uses the reek API.

troessner: >> No, to silence warnings we otherwise throw at anyone who uses the reek API.

Ok, very good point. Then how about switching back to that gem you proposed to introduce before? @mvz WDYT?

mvz: > Then how about switching back to that gem you proposed to introduce before?

Given that I can see the point of private readers, I'd say yes.

> SmellRepository does not have any children. Making this protected is highly misleading then since I'd naturally assume it does.

That's not why it's misleading: Given Ruby's semantics, it suggests SmellRepository instances need to access each other's configuration and other values.

troessner: >> That's not why it's misleading: Given Ruby's semantics, it suggests SmellRepository instances need to access each other's configuration and other values.

Yes, you are right.;)

chastell:
Ready for review once again. :)

troessner:
Nice, merged!

PR #574, opened 2015-07-05
Attributes smell no readers

beanieboi:
this is the PR for #491

no against master, prev PR: https://github.com/troessner/reek/pull/506

this PR:

enables the Attribute smell by default
does not report the usage of attr and attr_reader as a smell.

PR #613, opened 2015-07-21
Attributes smell no readers

mvz:
Follow-up for #574.

This one cleans things up and adds support for `attr :foo, true`.

I've left @beanieboi's work, with some trivial fixes to make the build green, in a separate commit.

I'd like to unify the visibility handlers, but I'll leave that for a separate PR.

### DIFF
--- a/docs/Attribute.md
+++ b/docs/Attribute.md
@@ -2,7 +2,11 @@
 
 ## Introduction
 
-A class that publishes a getter or setter for an instance variable invites client classes to become too intimate with its inner workings, and in particular with its representation of state.
+A class that publishes a setter for an instance variable invites
+client classes to become too intimate with its inner workings, and in
+particular with its representation of state.
+
+The same holds to a lesser extent for getters, but Reek doesn't flag those.
 
 ## Example
 
@@ -20,24 +24,16 @@ end
 reek test.rb
 
 test.rb -- 1 warning:
-  [2]:Klass declares the attribute dummy (Attribute)
+  [2]:Klass declares the writable attribute dummy (Attribute)
 ```
 
 ## Support in Reek
 
-Right now this smell is disabled by default since it is highly subjective.
+This detector it raises a warning for every public `attr_writer`,
+`attr_accessor`, and `attr` with the writable flag set to `true`.
 
-When this detector is enabled it raises a warning for every `attr`, `attr_reader`, `attr_writer` and `attr_accessor` -- including those that are private.
+Reek does not raise warnings for read-only attributes.
 
 ## Configuration
-
-If you want to enable it you can do so by placing
-
-```yaml
-Attribute:
-  enabled: true
-```
-
-in your reek configuration file.
 
 `Attribute` supports only the [Basic Smell Options](Basic-Smell-Options.md).
--- a/features/samples.feature
+++ b/features/samples.feature
@@ -63,7 +63,14 @@ Feature: Basic smell detection
     Then the exit status indicates smells
     And it reports:
     """
-    optparse.rb -- 111 warnings:
+    optparse.rb -- 119 warnings:
+      OptionParser declares the writable attribute banner (Attribute)
+      OptionParser declares the writable attribute default_argv (Attribute)
+      OptionParser declares the writable attribute program_name (Attribute)
+      OptionParser declares the writable attribute release (Attribute)
+      OptionParser declares the writable attribute summary_indent (Attribute)
+      OptionParser declares the writable attribute summary_width (Attribute)
+      OptionParser declares the writable attribute version (Attribute)
       OptionParser has at least 42 methods (TooManyMethods)
       OptionParser has the variable name 'f' (UncommunicativeVariableName)
       OptionParser has the variable name 'k' (UncommunicativeVariableName)
@@ -148,6 +155,7 @@ Feature: Basic smell detection
       OptionParser::List#update has 5 parameters (LongParameterList)
       OptionParser::List#update has approx 10 statements (TooManyStatements)
       OptionParser::List#update has the variable name 'o' (UncommunicativeVariableName)
+      OptionParser::ParseError declares the writable attribute reason (Attribute)
       OptionParser::ParseError#set_option is controlled by argument eq (ControlParameter)
       OptionParser::Switch#add_banner has the variable name 's' (UncommunicativeVariableName)
       OptionParser::Switch#initialize has 7 parameters (LongParameterList)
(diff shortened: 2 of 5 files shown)


mvz:
Ready for review.

docs/Attribute.md
```diff
@@ -20,24 +22,16 @@ end
...
 test.rb -- 1 warning:
-  [2]:Klass declares the attribute dummy (Attribute)
+  [2]:Klass declares the writeable attribute dummy (Attribute)
 ```
 
 ## Support in Reek
 
-Right now this smell is disabled by default since it is highly subjective.
+When this detector is enabled it raises a warning for every public
+`attr_writer`, `attr_accessor`, and `attr` with the writeable flag set to true.
```
chastell: Not 100% sure, but maybe true → \`true\`?

lib/reek/smells/attribute.rb
```diff
@@ -9,22 +9,18 @@ module Smells
     # invites client classes to become too intimate with its inner workings,
     # and in particular with its representation of state.
     #
-    # Currently this detector raises a warning for every +attr+,
-    # +attr_reader+, +attr_writer+ and +attr_accessor+ -- including those
-    # that are private.
+    # This detector raises a warning for every public
+    # +attr_writer+, +attr_accessor+, and +attr+ with the writable set to true.
```
chastell: Let’s standardise on either spelling: ‘writable’ or ‘writeable’. I vote for ‘writable’, but I won’t mind if the other one wins.

chastell: Also let’s standardise on the code-marking sequence (backticks vs plusses) and mark `true` (or not). :)

chastell:
I have two minor nitpicks, otherwise it looks good – but I’ll defer the architectural quality check to @troessner. :)

lib/reek/smells/attribute.rb
mvz: Since this is RDoc format, the code-marking character is the plus. I think we use backticks in several places but that doesn't actually yield the proper result in the docs.

docs/Attribute.md
mvz: Agreed.

lib/reek/smells/attribute.rb
mvz: Writable seems to be the correct spelling. I'll fix it.

troessner:
Please gimme a until the weekend to review this as well.:)

features/samples.feature
```diff
@@ -183,7 +191,13 @@ Feature: Basic smell detection
...
     And it reports:
     """
-    redcloth.rb -- 95 warnings:
+    redcloth.rb -- 101 warnings:
+      RedCloth declares the writeable attribute filter_html (Attribute)
+      RedCloth declares the writeable attribute filter_styles (Attribute)
+      RedCloth declares the writeable attribute hard_breaks (Attribute)
+      RedCloth declares the writeable attribute lite_mode (Attribute)
+      RedCloth declares the writeable attribute no_span_caps (Attribute)
+      RedCloth declares the writeable attribute rules (Attribute)
```
troessner: I think we should get rid of this whole redcloth/inline.rb thingie, but let's discuss this in my upcoming PR

docs/Attribute.md
```diff
@@ -2,7 +2,9 @@
 
 ## Introduction
 
-A class that publishes a getter or setter for an instance variable invites client classes to become too intimate with its inner workings, and in particular with its representation of state.
+A class that publishes a getter or setter for an instance variable invites
+client classes to become too intimate with its inner workings, and in
```
troessner: We should also mention here that we only check for setters, not getters. Otherwise this is confusing.

lib/reek/smells/attribute.rb
```diff
@@ -32,21 +28,19 @@ def self.contexts # :nodoc:
...
-        super.merge(SmellConfiguration::ENABLED_KEY => false)
-      end
-
       #
       # Checks whether the given class declares any attributes.
       #
       # @return [Array<SmellWarning>]
       #
       def examine_context(ctx)
+        @visiblity_tracker = {}
```
troessner: 1. This voolates our new styleguide regarding instance variables..:)
2. Why not leave it in the constructor. Looks to me like it belongs there.

lib/reek/smells/attribute.rb
```diff
@@ -54,17 +48,35 @@ def examine_context(ctx)
       private
 
       def attributes_in(module_ctx)
+        result = Set.new
```
troessner: Can we use a more speaking variable instead of the generic "result"?

lib/reek/smells/attribute.rb
```diff
@@ -54,17 +48,35 @@ def examine_context(ctx)
...
-          elsif ATTR_DEFN_METHODS.include?(call_node.method_name)
-            call_node.arg_names.each do |arg|
-              @visiblity_tracker[arg] = @visiblity_mode
-              @result << [arg, call_node.line]
-            end
-          end
+          result += track_attributes(call_node)
         end
-        @result.select { |args| recorded_public_methods.include?(args[0]) }
+        result.select { |args| recorded_public_methods.include?(args[0]) }
```
troessner: Can we document here what "args" are structurally and what we expect args [0] to be?

lib/reek/smells/attribute.rb
```diff
@@ -54,17 +48,35 @@ def examine_context(ctx)
...
+      end
+
+      def track_attributes(call_node)
+        if attribute_writer? call_node
+          return track_arguments call_node.args, call_node.line
+        end
+        track_visibility call_node if visibility_modifier? call_node
```
troessner: Fantastic readability. If this was only everywhere like that in our codebase..;)

lib/reek/smells/attribute.rb
```diff
@@ -54,17 +48,35 @@ def examine_context(ctx)
...
+        []
+      end
+
+      def attribute_writer?(call_node)
+        method_name = call_node.method_name
+        ATTR_DEFN_METHODS.include?(method_name) ||
+          method_name == :attr && call_node.args.last.type == :true
```
troessner: How about moving the second part of the guard into a speaking method  with a proper yard doc that explains the odd "attr" syntax briefly?

spec/reek/smells/attribute_spec.rb
```diff
@@ -20,99 +12,123 @@
...
+    it "doesn't record private attributes" do
+      src = '
         class Klass
-          attr_accessor :my_attr
+          private
+          attr_writer :attr1
+          attr_accessor :attr2
+          attr :attr3
+          attr :attr4, true
+          attr_reader :attr5
```
troessner: Nice specs..;)

lib/reek/smells/attribute.rb
mvz: 1: Yes. It is ~~fixed~~ completely rewritten in #614.
2: We can't leave it in the constructor because visibility tracking needs te be reset for every context.

troessner: >> 2: We can't leave it in the constructor because visibility tracking needs te be reset for every context.

Ok, understood.

troessner:
Didn't realize that this awesome pul request was ready for merge until I just re-scanned it.

Merged!

PR #614, opened 2015-07-22
Unify visibility detectors

mvz:
Branched off #613.

Still needs some clean-ups of FIXMEs etc.


troessner:
#613 is merged, so let's get this baby ready for merge as well! :)

mvz:
Ready to merge.

lib/reek/tree_walker.rb
```diff
@@ -94,6 +94,13 @@ def process_send(exp)
       if visibility_modifier? exp
         element.track_visibility(exp.method_name, exp.arg_names)
       end
+      if attribute_writer? exp
+        exp.args.each do |arg|
+          next unless arg.type == :sym
+          inside_new_context(Context::MethodContext, arg) do
+          end
```
chastell: I assume that this is needed for `TreeWalker#inside_new_context`’s side-effects (appending the child context) – but the empty block looks strange.

I think `inside_new_context` should either `yield if block_given?` or there should be a comment inside this empty block why it’s empty.

chastell:
I really like this PR, I just have a minor nitpick. I’m ok with merging this as-is, but would love @troessner to take a look just in case. :)

lib/reek/tree_walker.rb
mvz: I agree.

troessner:
>> I really like this PR, I just have a minor nitpick. I’m ok with merging this as-is, but would love @troessner to take a look just in case. :)

I'll happily review this little honeybadger but I'm a little blocked right now, please gimme a day or two.:)

lib/reek/tree_walker.rb
```diff
@@ -211,6 +218,14 @@ def visibility_modifier?(call_node)
       VISIBILITY_MODIFIERS.include?(call_node.method_name)
     end
 
+    # FIXME: Move to SendNode?
+    def attribute_writer?(call_node)
+      method_name = call_node.method_name
+      ATTR_DEFN_METHODS.include?(method_name) ||
+        method_name == :attr && call_node.args.last.type == :true
```
troessner: Hmmm, can we somehow put 

```Ruby
method_name == :attr && call_node.args.last.type == :true
```

into a separate, appropriately named method and in this have method explaining this? I assume that most people are not familiar with 

```Ruby
attr :foo, true
```

How about

```Ruby
def attr_call_special_form(method_name, call_node)
  # Handles the case where we create an attribute writer via:
  # attr :foo, true
  method_name == :attr && call_node.args.last.type == :true
end
```

?

lib/reek/tree_walker.rb
```diff
@@ -211,6 +218,14 @@ def visibility_modifier?(call_node)
       VISIBILITY_MODIFIERS.include?(call_node.method_name)
     end
 
+    # FIXME: Move to SendNode?
```
troessner: Let's remove this comment and rather discuss it via github issue. Comments like that tend to get stale and confusing pretty quick....

lib/reek/ast/sexp_extensions.rb
```diff
@@ -442,6 +442,19 @@ def method_name
...
 
       ZsuperNode = SuperNode
+
+      # Utility methods for :sym nodes.
+      module SymNode
+        def name
+          children.first
+        end
+
+        # FIXME: Move full_name logic to contexts
```
troessner: Let's remove this comment and rather discuss it via github issue. Comments like that tend to get stale and confusing pretty quick....

troessner:
We can remove this https://github.com/troessner/reek/blob/unify-visibility-detectors/lib/reek/smells/attribute.rb#L20 as well now, right?

troessner:
Just a couple of comments, rest looks awesome.:)
LGTM

chastell:
…and this needs to be rebased. :)

mvz:
I'll try to fix and rebase this weekend.

mvz:
Done. Just trying to get rbx to build :crying_cat_face:.

chastell:
Patiently restarted the Rubinius build UNTIL IT WAS GREEN. Can you remove the `FIXME` that’s still there, as per Timo’s request?

mvz:
> Can you remove the FIXME that’s still there, as per Timo’s request?

Dammit, missed that one. Done.

lib/reek/tree_walker.rb
```diff
@@ -191,12 +197,15 @@ def count_statement_list(statement_list)
...
-      scope = klass.new(element, exp)
-      element.append_child_context(scope)
-      push(scope) do
+      push(new_context(klass, exp)) do
         yield
       end
-      scope
+    end
+
+    def new_context(klass, exp)
```
troessner: Nice!
Simple and clean, i like that a lot.

troessner:
Awesome job @mvz !

PR #616, opened 2015-07-22
Reimplement yardoc spec and fix a warning

chastell:
This PR fixes a `yardoc` warning, reimplements its call’s spec and checks the call for a successful return.

(Interestingly, we don’t seem to ever execute it, as we only ever run `spec/reek/**/*_spec.rb` files.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -206,8 +206,8 @@
 
 ### Major Changes
 * Smell detectors can be configured or disabled in code comments
-  * Comment with :reek:smell_name disables the named smell for a class, module or method
-  * Comment with :reek:smell_name:{...} for more detailed configuration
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -207,7 +207,7 @@
 ### Major Changes
 * Smell detectors can be configured or disabled in code comments
   * Comment with :reek:smell_name disables the named smell for a class, module or method
-  * Comment with :reek:smell_name:{...} for more detailed configuration
+  * Comment with :reek:smell_name:… for more detailed configuration
```
mvz: I think this should be
```
* Comment with `:reek:smell_name` disables the named smell for a class, module or method
* Comment with `:reek:smell_name:{...}` for more detailed configuration
```

chastell: Good catch, fixed!

PR #514, opened 2015-05-10
Hotfix: Handle array decomposition

mvz:
Fixes #502.

Basing this off master as it's a bug fix.

PR #617, opened 2015-07-23
Remove unnecessary specs.

troessner:
In my mind, those specs have no value at all and thus should be removed:

* They are way to generic. Just pass in a crappy source file and list the gazillions of warnings has no value for me
* They are duplicated. Every single warning in this spec is already tested separately
* They slow down development. Every time you touch any smell those specs will break which I feel is highly annoying.
* They decrease spec readability just because of the sheer amount of text

Let's remove them and make a big step forward in improving our spec suite.:)


chastell:
A strong :+1: for `inline.rb`, `optparse.rb` and `redcloth.rb` – they always show up when I grep through our codebase for some obvious wrongdoings to fix. ;)

Not sure about `unusual_syntax.rb` – do we cover these cases (especially the array decomposition one) elsewhere? (If so: ax it as well.)

troessner:
>> Not sure about unusual_syntax.rb – do we cover these cases (especially the array decomposition one) elsewhere? (If so: ax it as well.)

Yes, @mvz fixed that recently in #514 :)

mvz:
I'm not so sure. These specs caught serious issues with our Attribute implementation when I was working on that recently. They also give me some idea of the impact a change to a smell may have, and they serve as a reasonable-sized regression test.

Perhaps we can mitigate the problems by, e.g., storing the expected output in a separate file and allowing automated updates of that expected output; `git diff` will then show the change to reek's behavior anyway.

mvz:
> serious issues with our Attribute implementation

By which I mean it was totally broken when multiple classes are defined in the same file, but nobody noticed because the smell was disabled by default.

troessner:
Ok, but then one of those files / scenarios would be enough, wouldn't it?

mvz:
I would go for creating just one scenario with potentially several files and the desired output stored in a separate file (like we do for, e.g., the specs for the XML report).

troessner:
>> I would go for creating just one scenario with potentially several files and the desired output stored in a separate file (like we do for, e.g., the specs for the XML report).

Good idea! Let me look into that.

troessner:
On second thought I think this:

>> I would go for creating just one scenario with potentially several files

is a good idea but this:

>>  and the desired output stored in a separate file (like we do for, e.g., the specs for the XML report).

is not. The reason for this is that we **expect** the feature run to break if we add smells or update existing ones. The same is not true for the xml report (even though it might happen of course).

Or put it the other way round: The further indirection of putting the expected output into a separate file doesn't help us but rather makes our spec handling more inconvenient.

Let's go the middle way: All in one feature run, but we're still checking for the output on stdout.

troessner:
Oh man, our awesome specs are biting me in the ass:

```
+  Inline::C#module_name has the variable name 'md5' (UncommunicativeVariableName)
```

One warning more all of a sudden.
I'll look into it.

troessner:
Ready for re-review but bad branch name, closing this one and re-opening as #630

PR #619, opened 2015-07-23
I accidentally the whole XMLReport

chastell:
So I wanted to make the spec for `XMLReport` a bit more readable and parallelisable (fewer indirections, no `File` stubbing) and ended up with a new `XMLReport` that’s a bit more object-oriented and builds the output lazily rather than procedurally.

Notable changes:

* when a smell is reported for multiple lines we now include a separate `<error>` for every line;
* the XML is now human-readable.

I don’t think they are backward-incompatible, but I’ll reopen against `reek-4` if you believe so.

I’ll squash this one after a review. :)

### DIFF
--- a/lib/reek/report/report.rb
+++ b/lib/reek/report/report.rb
@@ -144,52 +144,45 @@ module Reek
       require 'rexml/document'
 
       def show
-        checkstyle = REXML::Element.new('checkstyle', document)
-
-        smells.group_by(&:source).each do |file, file_smells|
-          file_to_xml(file, file_smells, checkstyle)
-        end
-
-        print_xml(checkstyle.parent)
+        document.write output: $stdout, indent: 2
+        $stdout.puts
       end
 
       private
 
       def document
-        REXML::Document.new.tap do |doc|
-          doc << REXML::XMLDecl.new
+        REXML::Document.new.tap do |document|
+          document << REXML::XMLDecl.new << checkstyle
         end
       end
 
-      def file_to_xml(file, file_smells, parent)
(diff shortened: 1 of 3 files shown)


lib/reek/report/report.rb
```diff
@@ -144,53 +144,46 @@ class XMLReport < Base
...
 
       def show
-        checkstyle = REXML::Element.new('checkstyle', document)
-
-        smells.group_by(&:source).each do |file, file_smells|
-          file_to_xml(file, file_smells, checkstyle)
-        end
-
-        print_xml(checkstyle.parent)
+        document.write $stdout, 2
```
troessner: Magic number alert..;)
I know what the 2 refers to, but it is still baroque.
Any idea how to make this clearer?

troessner:
One comment and then lgtm

lib/reek/report/report.rb
chastell: 1. `document.write output: $stdout, indent: 2`
2. `REXML::Formatters::Pretty.new.write document, $stdout`

Either if fine by me; the former is shorter and more readable, neither is very object-oriented (document shouldn’t really write itself, formatter should just format and return the formatted version, not write…).

troessner:
```Ruby
document.write output: $stdout, indent: 2
```

Let's go with that one!.:)

chastell: Fixed, rebased, squashed and green. :white_check_mark:

mvz:
Nice!

Issue #608, opened 2015-07-19
Make AppConfiguration's initializer flexible enough to deal with different inputs


troessner:
Coming from here: https://github.com/troessner/reek/pull/593/files#diff-373a31cd38fa792eeb0826b025e8da97R69

mvz:
Let's at lease/also get rid of the options parameter and just pass in the path or something.

troessner:
Fix wip in #625

troessner:
#625 is merged -> closing this one.

PR #623, opened 2015-07-28
Update API docs.

troessner:
Ready for review.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1,29 +1,35 @@
 # Using `reek` inside your Ruby application
 
-`reek` can be used inside another Ruby project.
+## Installation
+
+Either standalone via
 
 ```bash
 gem install reek
 ```
 
-## Using a reporter
+or by adding
 
-You can use reek inside your Ruby file `check_dirty.rb`
+```
+gem 'reek'
+```
+
+to your Gemfile.
+
+## Quick start
+
+Code says more than a thousand words:
 
 ```ruby
 require 'reek'
 
-source = <<-END
+source = <<-EOS
   class Dirty
-    # This method smells of :reek:NestedIterators but ignores them
-    def awful(x, y, offset = 0, log = false)
-      puts @screen.title
-      @screen = widgets.map { |w| w.each { |key| key += 3 * x } }
-      puts @screen.contents
-      fail
+    def m(a,b,c)
+      puts a,b
     end
   end
-END
+EOS
 
 reporter = Reek::Report::TextReport.new
 examiner = Reek::Examiner.new(source)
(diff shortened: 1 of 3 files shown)


docs/Configuration-Files.md
```diff
@@ -2,48 +2,83 @@
 
 ## Configuration loading
 
-Configuring `reek` via configuration file is by far the most powerful way.
 
-There are 3 ways of passing `reek` a configuration file:
+Configuring `reek` via a configuration file is by far the most powerful way.
```
troessner: The changes in this file are basically due a copying of https://github.com/troessner/reek#configuration-file into this file.
So yes, it is duplicated. I still think this the right way to go in this special case because:

1.) we want to link to this doc independently 
2.) while maintaining this piece of information in the README for conversion reasons. Or in non-marketing speak: If we move the information from the README to the /docs directory a lot people will jump the ship and not use reek because the README is incomplete and they need to jump over the mental barrier of leaving the README for something that is considered a basic thing.

Alternatively I'd vote for removing /docs/Configuration-Files.md.
WDYT?

troessner: On second thought i'd suggest to kill /docs/Configuration-Files.md in this PR.

chastell:
I like the changes, but Ataru doesn’t.

troessner:
>> I like the changes, but Ataru doesn’t.

Yes, that is a problem indeed. In order to make the examples above work we need a config file since right now we only support  passing in a path for a real file.

Now we could:

1.) Create that file in the example above via ataru which would not only be clumsy but also highly confusing
2.) Create and tear down that file in ataru_setup.rb. There's no way to check in which example we are so this would happen kind of globally which might have weird side effects and cause unrelated examples to fail.
3.) Disable ataru for README.md
4.) Go with [2] for now, fix #608 and the refactor our examples.

I'd go with [4] - WDYT?

troessner:
I removed docs/Configuration-Files.md as well in this PR and updated all references.

mvz:
Not sure how [4] will fully solve it, since it (also?) fails because `dirty.rb` doesn't exist. Perhaps we can go back to the string example for that one and explain the other options for passing source in words?

troessner:
>> Not sure how [4] will fully solve it, since it (also?) fails because dirty.rb doesn't exist. Perhaps we can go back to the string example for that one and explain the other options for passing source in words?

Yes, forgot to mention it but that was my plan indeed.;)

mvz:
Ok :-).

docs/API.md
```diff
@@ -31,16 +41,21 @@ reporter.add_examiner examiner
...
-reporter.add_examiner Reek::Examiner.new(File.new('dirty.rb'))
+dirty.rb -- 4 warnings:
+  [2]:Dirty#m has the name 'm' (UncommunicativeMethodName)
+  [2]:Dirty#m has the parameter name 'a' (UncommunicativeParameterName)
+  [2]:Dirty#m has the parameter name 'b' (UncommunicativeParameterName)
+  [2]:Dirty#m has unused parameter 'c' (UnusedParameters)
 ```
 
-Also, besides normal text output, `reek` can generate output in YAML,
+Note that `Reek::Examiner.new` can take `source` as `String`, `Pathname`, `File` or `IO`.
```
mvz: I think we agreed in https://github.com/troessner/reek/pull/596#discussion_r34437240 and follow-up discussion to remove the Pathname option again?

troessner:
On second thought it probably makes more sense to fix #608 first and put this one on hold. I'll look into that asap.

troessner:
Addressed all of @mvz comments, now let me fix #608 :)

troessner:
Now with #608 I got this one back on track!
Re-review please..:)

docs/API.md
```diff
@@ -51,6 +63,78 @@ HTMLReport
...
+
+Given the example above you should load that as "default directive" which means that it will
+be the default configuration for smell types for which there is
+no "directory directive" (so a directory-specific configuration):
+
+```Ruby
+require 'reek'
+
+default_directive = { Reek::Smells::IrresponsibleModule => { 'enabled' => false } }
+configuration = Reek::Configuration::AppConfiguration.from_map default_directive: default_directive
```
mvz: I'm still not entirely happy with the API in `AppConfiguration.from_map`. Why not allow the following to work:

```ruby
hash = YAML.load_file(Pathname.new('config.reek'))
configuration = Reek::Configuration::AppConfiguration.from_map hash
```

WDYT?

troessner: Yeah, you're right, that's better indeed. However, this PR is a) concerned only with the API docs and (also in branch name) b) merely documenting already existing behaviour.
So how about merging this one and then updating "from_map" in a follow up PR so this doesn't get messy?
I created an issue for this so we don't forget about it:  https://github.com/troessner/reek/issues/662

mvz: Good plan. Merging ...

PR #625, opened 2015-08-02
Refactor configuration loading and make AppConfiguration's initializer more flexible.

troessner:
This makes AppConfiguration's initializer more flexible than before by allowing to pass in hashes as well.
It also refactors quite a lot: Instead of cramming everything into AppConfiguration with hashes and arrays it uses composition and kind of an DCI approach by extending hashes and arrays at runtime to give them the desired behaviour.
More code overall but I believe this is significantly better architecture than before with higher cohesion.
It will also make it significantly easier to maintain this code and to extend it, e.g. with new configuration options.
Precondition for #623

### DIFF
--- a/lib/reek/cli/silencer.rb
+++ b/lib/reek/cli/silencer.rb
@@ -3,11 +3,18 @@ module Reek
     # CLI silencer
     # @api private
     module Silencer
-      def self.silently
-        old_verbose, $VERBOSE = $VERBOSE, nil
-        yield if block_given?
+      module_function
+
+      def silently(stderr: nil, stdout: nil)
+        old_verbose = $VERBOSE
+        $VERBOSE = false
+        $stderr = StringIO.new if stderr
+        $stdout = StringIO.new if stdout
+        yield
       ensure
         $VERBOSE = old_verbose
+        $stderr = STDERR
+        $stdout = STDOUT
       end
     end
   end
--- a/lib/reek/configuration/app_configuration.rb
+++ b/lib/reek/configuration/app_configuration.rb
@@ -1,5 +1,9 @@
 require 'pathname'
 require_relative './configuration_file_finder'
+require_relative './configuration_validator'
+require_relative './default_directive'
+require_relative './directory_directives'
+require_relative './excluded_paths'
 
 module Reek
   # @api private
@@ -10,134 +14,94 @@ module Reek
     #
     # @api private
     class AppConfiguration
+      include ConfigurationValidator
       EXCLUDE_PATHS_KEY = 'exclude_paths'
-      attr_reader :exclude_paths, :default_directive, :directory_directives
+      private_attr_writer :directory_directives, :default_directive, :excluded_paths
 
-      # Given this configuration file:
+      # Instantiate a configuration via given path.
       #
-      # ---
-      # IrresponsibleModule:
-      #   enabled: false
-      # "app/helpers":
-      #   UtilityFunction:
-      #     enabled: false
-      # exclude_paths:
-      #   - "app/controllers"
+      # @param path [Pathname] the path to the config file
       #
-      # this would result in the following configuration:
-      #
-      # exclude_paths = [ Pathname('spec/samples/two_smelly_files') ]
-      # default_directive = { Reek::Smells::IrresponsibleModule => { "enabled" => false } }
-      # directory_directives = { Pathname("spec/samples/three_clean_files/") =>
-      #                          { Reek::Smells::UtilityFunction => { "enabled" => false } } }
-      #
-      #
-      # @param options [OpenStruct]
-      #   e.g.: #<OpenStruct report_format =: text,
-      #                      location_format =: numbers,
-      #                      colored = true,
-      #                      smells_to_detect = [],
-      #                      config_file = #<Pathname:config/defaults.reek>,
-      #                      argv = [ "lib/reek/spec" ]>
-      def initialize(options = nil)
-        @directory_directives = {}
-        @default_directive    = {}
-        @exclude_paths        = []
-
-        load options
+      # @return [AppConfiguration]
+      def self.from_path(path = nil)
+        allocate.tap do |instance|
+          instance.instance_eval { find_and_load(path: path) }
+        end
       end
 
-      # @param source_via [String] - the source of the code inspected
-      # @return [Hash] the directory directive for the source or, if there is
-      # none, the default directive
-      def directive_for(source_via)
-        directory_directive_for_source(source_via) || default_directive
-      end
-
-      private
-
-      private_attr_writer :exclude_paths
-
-      # @param source_via [String] - the source of the code inspected
-      # Might be a string, STDIN or Filename / Pathname. We're only interested in the source
-      # when it's coming from file since a directory_directive doesn't make sense
-      # for anything else.
-      # @return [Hash | nil] the configuration for the source or nil
-      def directory_directive_for_source(source_via)
-        return unless source_via
-        source_base_dir = Pathname.new(source_via).dirname
-        hit = best_directory_match_for source_base_dir
-        directory_directives[hit]
-      end
-
-      def load(options)
-        configuration_file = ConfigurationFileFinder.find_and_load(options: options)
-
-        configuration_file.each do |key, value|
-          case
-          when key == EXCLUDE_PATHS_KEY
-            handle_exclude_paths(value)
-          when smell_type?(key)
-            handle_default_directive(key, value)
-          else
-            handle_directory_directive(key, value)
+      # Instantiate a configuration by passing everything in.
+      #
+      # @param map [Hash] can have the following 3 keys:
+      # 1.) directory_directives [Hash] for instance:
+      #   { Pathname("spec/samples/three_clean_files/") =>
+      #     { Reek::Smells::UtilityFunction => { "enabled" => false } } }
+      # 2.) default_directive [Hash] for instance:
+      #   { Reek::Smells::IrresponsibleModule => { "enabled" => false } }
+      # 3.) excluded_paths [Array] for instance:
+      #   [ Pathname('spec/samples/two_smelly_files') ]
+      #
+      # @return [AppConfiguration]
+      def self.from_map(map = {})
+        allocate.tap do |instance|
+          instance.instance_eval do
+            self.directory_directives = map.fetch(:directory_directives, {}).
+              extend(DirectoryDirectives)
+            self.default_directive = map.fetch(:default_directive, {}).extend(DefaultDirective)
+            self.excluded_paths = map.fetch(:excluded_paths, []).extend(ExcludedPaths)
           end
         end
       end
 
-      def handle_exclude_paths(paths)
-        self.exclude_paths = paths.map do |path|
-          pathname = Pathname.new path.chomp('/')
-          raise ArgumentError, "Excluded directory #{path} does not exists" unless pathname.exist?
-          pathname
+      def self.default
+        from_path nil
+      end
+
+      def self.new(*)
+        raise NotImplementedError,
+              'Calling `new` is not supported, please use one of the factory methods'
+      end
+
+      # Returns the directive for a given directory.
+      #
+      # @param source_via [String] - the source of the code inspected
+      #
+      # @return [Hash] the directory directive for the source or, if there is
+      # none, the default directive
+      def directive_for(source_via)
+        directory_directives.directive_for(source_via) || default_directive
+      end
+
+      def path_excluded?(path)
+        excluded_paths.include?(path)
+      end
+
+      private
+
+      def directory_directives
+        @directory_directives ||= {}.extend(DirectoryDirectives)
+      end
+
+      def default_directive
+        @default_directive ||= {}.extend(DefaultDirective)
+      end
+
+      def excluded_paths
+        @excluded_paths ||= [].extend(ExcludedPaths)
+      end
+
+      def find_and_load(path: nil)
+        configuration_file = ConfigurationFileFinder.find_and_load(path: path)
+
+        configuration_file.each do |key, value|
+          case
+          when key == EXCLUDE_PATHS_KEY
+            excluded_paths.add value
+          when smell_type?(key)
+            default_directive.add key, value
+          else
+            directory_directives.add key, value
+          end
         end
-      end
-
-      def handle_default_directive(key, config)
-        klass = Reek::Smells.const_get(key)
-        default_directive[klass] = config
-      end
-
-      def handle_directory_directive(path, config)
-        pathname = Pathname.new path.chomp('/')
-        validate_directive pathname
-
-        directory_directives[pathname] = config.each_with_object({}) do |(key, value), hash|
-          abort(error_message_for_invalid_smell_type(key)) unless smell_type?(key)
-          hash[Reek::Smells.const_get(key)] = value
-        end
-      end
-
-      def best_directory_match_for(source_base_dir)
-        directory_directives.
-          keys.
-          select { |pathname| source_base_dir.to_s =~ /#{pathname}/ }.
-          max_by { |pathname| pathname.to_s.length }
-      end
-
-      def smell_type?(key)
-        Reek::Smells.const_get key
-      rescue NameError
-        false
-      end
-
-      def error_message_for_invalid_smell_type(klass)
-        "You are trying to configure smell type #{klass} but we can't find one with that name.\n" \
-          "Please make sure you spelled it right (see 'config/defaults.reek' in the reek\n" \
-          'repository for a list of all available smell types.'
-      end
-
-      def error_message_for_missing_directory(pathname)
-        "Configuration error: Directory `#{pathname}` does not exist"
-      end
-
-      def error_message_for_file_given(pathname)
-        "Configuration error: `#{pathname}` is supposed to be a directory but is a file"
-      end
-
-      def validate_directive(pathname)
-        abort(error_message_for_missing_directory(pathname)) unless pathname.exist?
-        abort(error_message_for_file_given(pathname)) unless pathname.directory?
       end
     end
   end
--- a/lib/reek/configuration/configuration_validator.rb
+++ b/lib/reek/configuration/configuration_validator.rb
@@ -1,0 +1,35 @@
+module Reek
+  module Configuration
+    #
+    # Configuration validator module.
+    #
+    module ConfigurationValidator
+      private
+
+      def smell_type?(key)
+        Reek::Smells.const_get key
+      rescue NameError
+        false
+      end
+
+      def error_message_for_missing_directory(pathname)
+        "Configuration error: Directory `#{pathname}` does not exist"
+      end
+
+      def error_message_for_file_given(pathname)
+        "Configuration error: `#{pathname}` is supposed to be a directory but is a file"
+      end
+
+      def validate_directory(pathname)
+        abort(error_message_for_missing_directory(pathname)) unless pathname.exist?
+        abort(error_message_for_file_given(pathname)) if pathname.file?
+      end
+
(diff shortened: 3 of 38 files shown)


lib/reek/configuration/app_configuration.rb
```diff
@@ -33,19 +33,24 @@ class AppConfiguration
...
+      # @param  default_directive [Hash] see above for an example
+      # @param  exclude_paths [Array] see above for an example
+      def initialize(path: nil,
+                     directory_directives: nil,
+                     default_directive: nil,
+                     exclude_paths: nil)
+        if path && [directory_directives, default_directive, exclude_paths].any?
+          raise(ArgumentError,
+                'You can either pass a path or single configuration values but not both')
+        end
```
chastell: I feel somewhat bad about this, as it border on control coupling: if it’s either-or, then maybe we should have two constructors, one that takes a path and one that takes the other three? (A theoretical use case for having both would be ‘parse this path but set `exclude_paths` to the provided override’, but I’m not sure whether it would be useful in practice.)

(Also: this does not seem to be tested in any way.) ;)

troessner: >> I feel somewhat bad about this, as it border on control coupling: if it’s either-or, then maybe we should have two constructors, one that takes a path and one that takes the other three?

Yep, that was what I had in my first try but it didn't look that good. But you're right, that is de facto a control couple. Lemme try again.

>> (Also: this does not seem to be tested in any way.) ;)

Whoopsie, busted!.:)

troessner:
Back to wip.

troessner:
Ready for review!

lib/reek/cli/silencer.rb
```diff
@@ -4,10 +4,13 @@ module CLI
     # @api private
     module Silencer
       def self.silently
-        old_verbose, $VERBOSE = $VERBOSE, nil
-        yield if block_given?
+        original_stderr = $stderr
+        original_stdout = $stdout
+        $stderr = $stdout = StringIO.new
```
chastell: Hmmm. It seems you need this version (that gulps stdout/stderr, rather than turning off warnings) to use in specs – but wouldn’t using this in specs mean that any spec failures won’t show on spec runs?

lib/reek/configuration/app_configuration.rb
```diff
@@ -11,135 +15,89 @@ module Configuration
...
+      # @param path [Pathname] the path to the config file
       #
-      # this would result in the following configuration:
-      #
-      # exclude_paths = [ Pathname('spec/samples/two_smelly_files') ]
-      # default_directive = { Reek::Smells::IrresponsibleModule => { "enabled" => false } }
-      # directory_directives = { Pathname("spec/samples/three_clean_files/") =>
-      #                          { Reek::Smells::UtilityFunction => { "enabled" => false } } }
+      # @return [AppConfiguration]
+      def self.from_path(path: nil)
```
chastell: Given the method’s name, wouldn’t it be simpler to make the path a regular attribute?

lib/reek/configuration/app_configuration.rb
```diff
@@ -11,135 +15,89 @@ module Configuration
...
+                        default_directive: {},
+                        excluded_paths: [])
+        instance = allocate
+        instance.instance_eval do
+          self.directory_directives = directory_directives.extend(DirectoryDirectives)
+          self.default_directive    = default_directive.extend(DefaultDirective)
+          self.excluded_paths       = excluded_paths.extend(ExcludedPaths)
+        end
+        instance
+      end
```
chastell: Hmmm. `.from_path` and `.from_map` look awfully similar – maybe we can lazily instantiate them to the empty defaults on first access?

lib/reek/configuration/app_configuration.rb
```diff
@@ -11,135 +15,89 @@ module Configuration
...
+
+      def self.default
+        from_path path: nil
+      end
```
chastell: `.from_path`’s parameter defaults to `nil`, so this could just call a bare `.from_path`. Not sure if more readable, though…

lib/reek/configuration/directory_directives.rb
```diff
@@ -0,0 +1,54 @@
...
+          end
+        end
+        self
+      end
+
+      private
+
+      def best_match_for(source_base_dir)
+        keys.
+          select { |pathname| source_base_dir.to_s =~ /#{pathname}/ }.
```
chastell: Should it be `/#{Regexp.escape(pathname)}/`?

chastell:
I’m not sure about the ‘let’s extend core classes’ pattern – what’s the advantage of this over POROs that compose the given `Array`/`Hash` and delegate the few useful methods?

(Also: this needs rebasing, probably due to the ivar clean-up. Sorry!)

troessner:
>> I’m not sure about the ‘let’s extend core classes’ pattern – what’s the advantage of this over POROs that compose the given Array/Hash and delegate the few useful methods?

I see quite a few:

1.) Sounds trite, but it is more fitting. I really want a hash like structure, not only for simplicity but in the end, the way that it's used by e.g. SmellRepository IS as a hash, and not something "classy".

2.) In the beginning I had wrapped this with PORO's. The problem with this approach:  As soon as you want to actually access the underlying value (like we do in SmellRepository / Examiner) things start to go south. I ended up with "default_directive.directive", "directory_directives.directives" and "excluded_paths.paths" which not only does sound odd, it's confusing as hell. Calling that underlying value "value" or "data" doesn't make things better, rather worse.

3.) Regarding:

>> I’m not sure about the ‘let’s extend core classes’ pattern

Like I wrote above, it's DCI inspired which is  a well known and understand pattern. My understanding of DCI is summarized pretty well in this SO post: http://stackoverflow.com/a/10560611/5052578 just so we're on the same page here.

I believe the solution above is a concise (less code than with PORO + delegate), elegant (DCI) and rubyesqe way of solving this.

lib/reek/configuration/app_configuration.rb
troessner: But we don't use "path" anywhere. I'd consider it confusing to have an attribute that is nowhere ever used.

lib/reek/configuration/app_configuration.rb
troessner: Yes, I passed the "path: nil" here deliberately to make it crystal-clear what is going on.

lib/reek/configuration/directory_directives.rb
troessner: Ah, yes, good catch!
I switched it to

```Ruby
source_base_dir.to_s.match /#{Regexp.escape(pathname.to_s)}/
```

lib/reek/cli/silencer.rb
troessner: I basically rewrote the method to be more specific, lemme know what you think.

lib/reek/configuration/app_configuration.rb
troessner: Good  idea! Changed and also switched to using "allocate.tap..." in the factory methods.

troessner:
Addressed all comments and rebased!

lib/reek/configuration/app_configuration.rb
mvz: Perhaps @chastell meant making it a regular parameter?

lib/reek/configuration/app_configuration.rb
```diff
@@ -10,135 +14,95 @@ module Configuration
...
-      #                      colored = true,
-      #                      smells_to_detect = [],
-      #                      config_file = #<Pathname:config/defaults.reek>,
-      #                      argv = [ "lib/reek/spec" ]>
-      def initialize(options = nil)
-        @directory_directives = {}
-        @default_directive    = {}
-        @exclude_paths        = []
+      # @return [AppConfiguration]
+      def self.from_map(directory_directives: {},
```
mvz: Why not make `.from_map` take the full contents of a yaml config file as its single parameter, allowing unification of the handling of both cases?

lib/reek/configuration/app_configuration.rb
```diff
@@ -10,135 +14,95 @@ module Configuration
...
+          end
+        end
+      end
 
-        load options
+      def self.default
+        from_path path: nil
       end
 
+      def initialize(*)
```
mvz: I **think** a more common way to do this is to make either `new` or `initialize` private (can't remember which a.t.m.).

mvz:
> DCI

I really need to read more on DCI. Last time I looked at it, which was quite some time ago, all the explanations and examples were very unconvincing. Time to buy and read 'Clean Ruby', I suppose.

That said, I have also read quite a few arguments that subclassing Array and Hash is a very bad idea.

Isn't the D of DCI supposed to be a domain data object rather than a generic type like Array or Hash?

lib/reek/configuration/app_configuration.rb
troessner: >> Why not make .from_map take the full contents of a yaml config file as its single parameter, allowing unification of the handling of both cases?

And then we could move it back into the initializer, so abandon factory methods again. And at this point in time, @chastell would swoop in and suggest to make separate factory methods for this. And then we would have an infinite loop :laughing: 

In all seriousness: I believe the two factory methods are fine since they cater to two completely different use cases.
Not to mention that we had it like that before in this PR and it seems just a taaaaaad like bike shedding.:)

lib/reek/configuration/app_configuration.rb
troessner: But the way above is far more revealing and helpful:

Compare:

>> irb(main):012:0> class Foo; def bar; end; private :bar; end; Foo.new.bar
NoMethodError: private method `bar' called for #<Foo:0x007ff305448fb8>
	from (irb):12

vs.:

>> irb(main):013:0> Reek::Configuration::AppConfiguration.new
NotImplementedError: Calling `new` is not supported, please use one of the factory methods
	from /Users/troessner/dev/reek/lib/reek/configuration/app_configuration.rb:60:in `initialize'

The latter tells you immediately what's going on and what you should do.

troessner:
>> That said, I have also read quite a few arguments that subclassing Array and Hash is a very bad idea.

But I am NOT subclassing here. That was basically my point. :laughing: 
I am aware of the disadvantages of this (e.g. I like: http://words.steveklabnik.com/beware-subclassing-ruby-core-classes) 

I'm extending Hash objects, not classes dynamically at runtime. And this extension is scoped to the existence of the variable :)

lib/reek/configuration/app_configuration.rb
mvz: I don't remember seeing `.from_map` with a single argument before?

Anyway, I just get the feeling we're making a complicated extra factory method with tricky parameters whose use case is just the specs at the moment, and I would like to have a little more common code between creation from file and creation from data.

@chastell any thoughts on this?

lib/reek/configuration/app_configuration.rb
mvz: Your first example doesn't match what I meant, but I understand your point. I just find it a bit confusing that this class is instantiated without calling `#initialize` :smile:.

mvz:
> But I am NOT subclassing here.

Hmmyes. Like I said, I need to read more on DCI, preferably some examples that are less ridiculous than [this one](http://fulloo.info/Examples/RubyExamples/Dijkstra/DijkstraListing.html) :laughing:.

troessner:
>> that are less ridiculous than this one :laughing:.

Yep, true. BTW. I just realized that we're already using the "extend at runtime" pattern already heavily, but only for classes: https://github.com/troessner/reek/blob/master/lib/reek/ast/ast_node_class_map.rb#L18

lib/reek/configuration/app_configuration.rb
troessner: I think allocate is incredibly helpful for exactly this use case since ruby is single-dispatch and not double-dispatch (which I would prefer, but I'm not Matz unfortunately) : factory methods. Actually, I can't think of any other uses cases right now but for this one it really shines :laughing: 

I think this SO post sums it up pretty nicely: http://stackoverflow.com/questions/13059552/classallocate-and-its-uses

>> Despite what the documentation says, the existence of the allocate method is not so much about memory management as it is about providing some finer grained control over the object creation lifecycle. Most of the time, you won't need this feature, but it is useful for certain edge cases.

lib/reek/configuration/app_configuration.rb
troessner: >> I don't remember seeing .from_map with a single argument before?

Ah, then I misunderstood you indeed!
But we're still talking about 2 factory methods, right?

mvz: > But we're still talking about 2 factory methods, right?

Absolutely.

troessner: Ok, understood, makes sense. Lemme change that (probably gonna happen tomorrow since I'm semi-offline today).

lib/reek/configuration/app_configuration.rb
mvz: Wow, I didn't really know about `allocate`. It may be just what I need to fix some long-standing issues with my [GirFFI](https://github.com/mvz/gir_ffi) project.

troessner: >> Wow, I didn't really know about allocate.

And this is exactly what I love about reviewing pull requests. I think I learned 90% of what I know about any programming language using pull requests (especially including yours.:) )

lib/reek/configuration/app_configuration.rb
chastell: Argh, yes, sorry, a regular parameter. (I can see how `path` is a kwarg everywhere else, so I can understand if there’s a need for symmetery – but on its own a `from_path` method should probably just take a regular `path` parameter.)

lib/reek/configuration/app_configuration.rb
chastell: > I think a more common way to do this is to make either `new` or `initialize` private

Oh wow, I almost made exactly this comment and then deleted it before posting. :) I think making `.new` private is the canonical way (both when you want a singleton and when you want to have dedicated factory methods), but – as Timo points out – that would not have the explanatory feature of the current `raise`. Thinking about this a bit more – how about moving the explanatory exception from `#initialize` to `.new`?

chastell: > Wow, I didn't really know about `allocate`.

When you’re at it – check out this and the following slides if you don’t want to get tripped when playing around with `allocate`: http://talks.chastell.net/rubyconf-pt-2014/#/17

lib/reek/configuration/app_configuration.rb
troessner: >> Argh, yes, sorry, a regular parameter. (I can see how path is a kwarg everywhere else, so I can understand if there’s a need for symmetery – but on its own a from_path method should probably just take a regular path parameter.)

Good point, changed.

lib/reek/configuration/app_configuration.rb
troessner: >> Thinking about this a bit more – how about moving the explanatory exception from #initialize to .new?

Good point, changed.

troessner:
Alright, addressed all comments! Merge?.:)

lib/reek/configuration/configuration_validator.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def error_message_for_file_given(pathname)
+        "Configuration error: `#{pathname}` is supposed to be a directory but is a file"
+      end
+
+      def validate_directory(pathname)
+        abort(error_message_for_missing_directory(pathname)) unless pathname.exist?
+        abort(error_message_for_file_given(pathname)) if pathname.file?
+      end
```
chastell: Why `abort` here (which is effectively `exit 1` + a message to stderr) rather than a regular `raise`?

troessner: Because a "raise" would give an ugly, useless stacktrace. The "abort" happens on a configuration error done by the user and we should make it as easy as possible for him to see what's going on.
A stacktrace is far more deterring than just a proper error message.
P.S.: 
I believe we should use "abort" instead of raise for all errors that are presented to the user but that would be another PR / discussion.

lib/reek/cli/silencer.rb
```diff
@@ -3,11 +3,24 @@ module CLI
...
+        if stderr
+          old_stderr = $stderr
+          $stderr = StringIO.new
+        end
+
         yield if block_given?
       ensure
         $VERBOSE = old_verbose
+        $stderr = old_stderr if stderr
+        $stdout = old_stdout if stdout
```
chastell: From the ‘made the comment, deleted, made it again, deleted, made it third time’ department: I’m thinking whether reassigning to `STDERR` and `STDOUT` wouldn’t be more canonical (and significantly simpler). They’re supposed to hold the references to the canonical streams, and the only reason for caching `old_stderr`/`old_stdout` by hand would be if this silencer was itself embedded in another piece of code tha alters `$stderr`/`$stdout`.

Also, rather than Boolean flags, how about making the kwargs the target streams (so the caller would pass `StringIO.new` or whatever it needs rather than `true`)? And how about making this a `module_function`? :)

```ruby
    module Silencer
      module_function

      def silently(stdout: nil, stderr: nil)
        old_verbose = $VERBOSE
        $VERBOSE = false
        $stdout = stdout if stdout
        $stderr = stderr if stderr
        yield
      ensure
        $VERBOSE = old_verbose
        $stderr = STDERR
        $stdout = STDOUT
      end
    end
```

chastell:
Two minor nitpicks (open for discussion) and I’m ok with this PR (great work!). :)

lib/reek/configuration/configuration_validator.rb
chastell: Oh ok, I’m convinced. :)

lib/reek/cli/silencer.rb
troessner: >>  I’m thinking whether reassigning to STDERR and STDOUT wouldn’t be more canonical (and significantly simpler). They’re supposed to hold the references to the canonical streams, and the only reason for caching old_stderr/old_stdout by hand would be if this silencer was itself embedded in another piece of code tha alters $stderr/$stdout.

Ah, I like that.:)

>>  how about making the kwargs the target streams (so the caller would pass StringIO.new or whatever it needs rather than true)?

But so far in 100% of all cases where we use "silently" we just want to, uhm, silence the corresponding stream and we don't care if that is a StringIO, a "true" or a unicorn on a rainbow.
So forcing the caller to pass in something else we generate the impression that this is important, when in fact it isn't. 
So I agree with all of your other suggestions, but I'd rather keep the flags.

troessner: Updated, but with the boolean flags.

chastell:
I’m ok with the refactorings, but today is apparently not a Green Build Day…

troessner:
Rebased, specs green because of my second commit which I would like to leave as separate commit and ready to go!
Merge?

chastell:
> which I would like to leave as separate commit

Normally I would be all ‘hell yeah, more smaller, intention-revealing commits!’ – but in this case it means we knowingly keep a commit that breaks the build. Such commit will show up as a false negative on every `git bisect` run… Is there a chance where you can arrange these two commits so that we’re not creating a known-red commit?

(I can merge this as-is if you don’t share the above concern; I mostly wanted to raise it and see what’s our general viewpoint on known-red commits.)

troessner:
>> Normally I would be all ‘hell yeah, more smaller, intention-revealing commits!’ – but in this case it means we knowingly keep a commit that breaks the build. Such commit will show up as a false negative on every git bisect run…

Yes, you're right.

>> Is there a chance you can arrange these two commits so that we’re not creating a known-red commit?

I just squashed both commits to one!

chastell:
…aaand merged! Thanks, this is a very welcome refactoring!

PR #630, opened 2015-08-09
Refactor our samples cucumber feature.

troessner:


troessner:
Replacement for #617 with proper branch name now.

PR #631, opened 2015-08-09
Refactor our samples cucumber feature.

troessner:
Doesn't seem like much, but for me it makes our spec set up less confusing:

Before we had 3 seemingly arbitrary, separate scenarios for basically checking all-the-things, now we have just one to make it clear that we're checking reek from front to back here.


troessner:
Re-created from #630 since that one was kind of fragged up by my git-fu.

chastell:
Why is the `unusual_syntax.rb` dropped from the spec? And if we drop it from there, shouldn’t we drop the file itself as well?

troessner:
>> Why is the unusual_syntax.rb dropped from the spec?

Because @mvz fixed that recently in #514 and we now have proper specs for this, so there is no need to use an ugly cucumber feature.:)

>> And if we drop it from there, shouldn’t we drop the file itself as well?

Good catch, missed that, removed!

mvz:
Looks good :shipit:.

PR #634, opened 2015-08-10
Update TooManyInstanceVariables default.

troessner:
Decrease it from 9 to 4, since 9 is wayyyyy too high.

### DIFF
--- a/spec/reek/smells/too_many_instance_variables_spec.rb
+++ b/spec/reek/smells/too_many_instance_variables_spec.rb
@@ -10,12 +10,26 @@ RSpec.describe Reek::Smells::TooManyInstanceVariables do
 
   it_should_behave_like 'SmellDetector'
 
+  def default_max_ivars
+    Reek::Smells::TooManyInstanceVariables::DEFAULT_MAX_IVARS
+  end
+
+  def too_many_ivars
+    default_max_ivars + 1
+  end
+
+  def ivar_sequence(count: default_max_ivars, alphabet: ('a'..'z').to_a)
+    alphabet.first(count).map do |name|
+      "@#{name}=#{rand}"
+    end.join(',')
+  end
+
   context 'counting instance variables' do
-    it 'should not report 9 ivars' do
+    it 'should not report for non-excessive ivars' do
       src = <<-EOS
         class Empty
           def ivars
-            @aa=@ab=@ac=@ad=@ae=@af=@ag=@ah=@ai=4
+           #{ivar_sequence}
           end
         end
       EOS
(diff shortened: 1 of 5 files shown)


troessner:
Ready for review!

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -10,12 +10,26 @@
 
   it_should_behave_like 'SmellDetector'
 
+  def default_max_ivars
+    Reek::Smells::TooManyInstanceVariables::DEFAULT_MAX_IVARS
+  end
+
+  def too_much_ivars
+    default_max_ivars + 1
+  end
```
chastell: I’m aware I’m bordering on [obsessing over grammar minutiæ](https://www.youtube.com/watch?v=UJPeg1b-jVo), but let’s go with `too_many_ivars` for countable amounts. :)

chastell:
One minor grammar nitpick and it looks good to me. I’ll see whether bumping Aruba fixes the JRuby build. ಠ_ರೃ

troessner:
>> One minor grammar nitpick and it looks good to me.

Haha, I was re-reviewing my PR in the S-Bahn on my way to work today and I thought "Oh, that is for sure something Piotr will comment on before I have a chance to fix that".:D

troessner:
Addressed your comment - now we only need to make jruby happy, rebase and then we can get this bad boy on the road.:)

troessner:
Green! Merge?

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -26,36 +40,36 @@ def ivars
...
           def ivars_a
-            @aa=@ab=@ac=@ad=@ae
+            #{ivar_sequence}
           end
         end
 
         class Full
           def ivars_b
-            @af=@ag=@ah=@ai=@aj
+            #{ivar_sequence}
```
mvz: I just noticed this changes the meaning of this test because it will use the same sequence instead of a new one.

I understand the need to make the spec not depend on the precise default value, but I feel claritiy is lost. How about adding instead a guard expectation that notices the default has changed and encourages the developer to look at these specs?

troessner: >> How about adding instead a guard expectation that notices the default has changed and encourages the developer to look at these specs?

I honestly have not a single clue what you mean with this sentence :)

mvz: > I honestly have not a single clue what you mean with this sentence :)

LOL!

mvz: I'll make a PR

PR #638, opened 2015-08-13
Instance variable access clean-up

chastell:
This normalises our access to instance variables in `lib` – we only use ivars directly in constructors and for memoization.

A separate PR for cleaning up instance variables in `spec` is forthcoming. :)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,6 @@
-require 'rake/clean'
 require 'bundler/gem_tasks'
+require 'private_attr/everywhere'
+require 'rake/clean'
 
 Dir['tasks/**/*.rake'].each { |t| load t }
 
--- a/lib/reek/cli/warning_collector.rb
+++ b/lib/reek/cli/warning_collector.rb
@@ -8,16 +8,20 @@ module Reek
     # @api private
     class WarningCollector
       def initialize
-        @warnings = Set.new
+        @warnings_set = Set.new
       end
 
       def found_smell(warning)
-        @warnings.add(warning)
+        warnings_set.add(warning)
       end
 
       def warnings
-        @warnings.to_a.sort
+        warnings_set.to_a.sort
       end
+
+      private
+
+      private_attr_reader :warnings_set
     end
   end
 end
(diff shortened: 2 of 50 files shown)


lib/reek/cli/warning_collector.rb
```diff
@@ -8,16 +8,20 @@ module CLI
     # @api private
     class WarningCollector
       def initialize
-        @warnings = Set.new
+        @warnings_set = Set.new
```
mvz: Why the hungarian notation?

lib/reek/cli/warning_collector.rb
```diff
@@ -8,16 +8,20 @@ module CLI
...
       end
 
       def found_smell(warning)
-        @warnings.add(warning)
+        warnings_set.add(warning)
       end
 
       def warnings
-        @warnings.to_a.sort
+        warnings_set.to_a.sort
```
mvz: Ah, that's why :smile:.

Rakefile
```diff
@@ -1,5 +1,6 @@
 require 'rake/clean'
 require 'bundler/gem_tasks'
+require_relative 'lib/reek'
```
mvz: Why do we need this here?

mvz:
Looks good. I'm a little confused by the shuffling around of `require` statements, but I can understand that you don't want to require `private_attr` everywhere.

lib/reek/cli/warning_collector.rb
chastell: Yeah, I was unhappy about the type-based name, but I didn’t find a better solution that would keep being a minimal diff. :)

Rakefile
chastell: We don’t (anymore). Good catch, removed!

troessner:
Mein lieber Scholli!

>> Consider private_attr core reek functionality

Agree.

chastell:
Yeah, I’m looking foward to Ruby 2.3 which won’t warn on private attributes anymore. (Also, fun fact: Ruby 2.2 is the first one that lets subclasses call private inherited getters/setters, so I needed to fix the build for previous versions…).

chastell:
Gaaah, builds locally, blows up on Travis. Lemme fix. :)

troessner:
Merged, nice job!

PR #639, opened 2015-08-14
Check casgn value responds to type

### LINKED ISSUE #622: Reek fails on use of ||= with constants
I ran into an issue with the use of the or equals operator on constants.

Example file (test.rb):
```ruby
module Foo
  FOO ||= :foo
end
```

Output of ```reek test.rb```:
```
$ reek test.rb
/Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/ast/sexp_extensions.rb:408:in `defines_module?': undefined method `type' for :FOO:Symbol (NoMethodError)
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:54:in `process_casgn'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:38:in `process'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:77:in `block in process_default'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:76:in `each'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:76:in `process_default'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:40:in `process'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:77:in `block in process_default'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:76:in `each'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:76:in `process_default'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:47:in `block in process_module'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:191:in `block in inside_new_context'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:199:in `push'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:190:in `inside_new_context'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:46:in `process_module'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:38:in `process'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/tree_walker.rb:27:in `walk'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/examiner.rb:73:in `run'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/examiner.rb:35:in `initialize'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/cli/reek_command.rb:14:in `new'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/cli/reek_command.rb:14:in `block in execute'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/cli/reek_command.rb:13:in `each'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/cli/reek_command.rb:13:in `execute'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/lib/reek/cli/application.rb:34:in `execute'
	from /Library/Ruby/Gems/2.0.0/gems/reek-3.1/bin/reek:11:in `<top (required)>'
	from /usr/bin/reek:23:in `load'
	from /usr/bin/reek:23:in `<main>'
```

HParker:
Seems that the issue is `casgn`'s value did not respond to type when `tree_walker` is trying to decide if it `defines_module?`

since it doesn't respond to type, I think it is safe to say that it does not define a module.

### DIFF
--- a/lib/reek/ast/sexp_extensions.rb
+++ b/lib/reek/ast/sexp_extensions.rb
@@ -405,6 +405,7 @@ module Reek
         MODULE_DEFINERS = [:Class, :Struct]
 
         def defines_module?
+          return false if value.nil?
           call = case value.type
                  when :block
                    value.call
@@ -420,8 +421,18 @@ module Reek
           SexpFormatter.format(children[1])
         end
 
+        # there are two valid forms of the casgn sexp
+        # (casgn <namespace> <name> <value>) and
+        # (casgn <namespace> <name>) used in or-asgn and mlhs
+        #
+        # source = "class Hi; THIS ||= 3; end"
+        # (class
+        #   (const nil :Hi) nil
+        #   (or-asgn
+        #    (casgn nil :THIS)
+        #    (int 3)))
         def value
-          children.last
+          children[2]
         end
       end
 
--- a/spec/reek/ast/sexp_extensions_spec.rb
+++ b/spec/reek/ast/sexp_extensions_spec.rb
@@ -347,3 +347,15 @@ RSpec.describe Reek::AST::SexpExtensions::ModuleNode do
     end
   end
 end
+
+RSpec.describe Reek::AST::SexpExtensions::CasgnNode do
+  context 'with single assignment' do
+    subject do
+      s(:casgn, nil, :Foo)
+    end
+
+    it 'does not define a module' do
+      expect(subject.defines_module?).to eq(false)
+    end
+  end
+end
--- a/spec/reek/tree_walker_spec.rb
+++ b/spec/reek/tree_walker_spec.rb
@@ -95,6 +95,18 @@ RSpec.describe Reek::TreeWalker, 'statement counting' do
       expect(method.num_statements).to eq(6)
     end
 
+    it 'does not count constant assignment with or equals' do
+      source = 'class Hi; CONST ||= 1; end'
+      klass = process_method(source)
+      expect(klass.num_statements).to eq(0)
+    end
+
+    it 'does not count multi constant assignment' do
+      source = 'class Hi; CONST, OTHER_CONST = 1, 2; end'
+      klass = process_method(source)
+      expect(klass.num_statements).to eq(0)
+    end
+
     it 'does not count empty conditional expression' do
       method = process_method('def one() if val == 4; ; end; end')
       expect(method.num_statements).to eq(0)


mvz:
Fixes #622

lib/reek/ast/sexp_extensions.rb
```diff
@@ -405,6 +405,7 @@ module CasgnNode
         MODULE_DEFINERS = [:Class, :Struct]
 
         def defines_module?
+          return false unless value.respond_to?(:type)
```
mvz: I would prefer to avoid `respond_to?`.

My educated guess is that, given the sexp used in the test, we would expect value to be `nil`, whereas actually, it is `:Foo`. I.e., the `:casgn` sexp has the following structure:

```
s(:casgn, <namespace>, <name>, <value>)
```

If the value is empty, parser does not include it. Due to the naïve implementation of `#value`, it then returns the name instead.

So IMHO a better solution is to fix `#value` and make `defines_module?` return false if `value` is `nil`.

mvz:
Hi @HParker thanks for your pull request. Please take a look at my comment on the diff.

HParker:
@mvz if you wouldn't mind taking a look at https://github.com/troessner/reek/commit/18d8074a2833af5f33b444f30cae5198600d50ac I think you are right this is a much better implementation.

lib/reek/ast/sexp_extensions.rb
```diff
@@ -421,7 +422,7 @@ def name
         end
 
         def value
-          children.last
+          children.last if children.size > 2
```
troessner: Can you please add an explanatory comment that shows an example for the S-exp?

HParker: @troessner how would you feel about a comment something like,

```Ruby
# there are two valid forms of the casgn sexp
# (casgn <namespace> <name> <value>) and
# (casgn <namespace> <name>) used in or-asgn and mlhs
```

Here are a couple examples,
```
source = "class Hi; THIS, THAT = 2, 5; end"
Reek::Source::SourceCode.from(source).syntax_tree
(class
 (const nil :Hi) nil
 (masgn
  (mlhs
   (casgn nil :THIS)
   (casgn nil :THAT))
  (array
   (int 2)
   (int 5))))
```

```
source = "class Hi; THIS ||= 3; end"
(class
 (const nil :Hi) nil
 (or-asgn
  (casgn nil :THIS)
  (int 3)))
```

troessner: The last example would be picture-perfect in my mind.:)

mvz: Or you can go with `children[2]` and avoid a lot of complication :smile:.

:+1: on adding the explanation and example.

HParker:
Incorporated feedback. Hope the comment structure works for you all.
Let me know if I need to squish commits or anything before merge.

troessner:
Could you please add a unit test that actually ensures that we can now parse

```Ruby
module Foo
  Bar ||= :hossa
end 
```

properly? Then this would be good to go from my side.:)

HParker:
Would that be in tree_walker_spec? I am having trouble finding a good place for it. Sorry, still learning the structure of the code.

troessner:
>> Would that be in tree_walker_spec? 

Yes, it would.:)

mvz:
Or alternatively, in irresponsible_module_spec.

HParker:
added two specs that mirror the use cases referenced in the issue.

troessner:
Looking good! Just one last thing, could you please squash it to one commit?
Then we can merge this :)

HParker:
oh god what did I do. Hold on.

HParker:
apparently 5 years of git is not enough to know how to interactive rebase. Sorry for the spam. I am going to try to start fresh here.

HParker:
well, I learned some important lessons today. Thankfully I learned them on a small change. Sorry again for the noise. Hopefully everything is good now.

troessner:
>> Hopefully everything is good now.

It is! Restarting the build for rubinius to make travis happy.....as soon as it's green I'll merge.;)

troessner:
Merged, thanks a lot for this @HParker !

Releasing a new version right away to bring this fix live.

PR #648, opened 2015-08-19
Replace the highly misleading "SourceCode#description" with "origin".

### LINKED ISSUE #604: SourceCode#description
Regarding: https://github.com/troessner/reek/blob/master/lib/reek/source/source_code.rb#L16
We use "description" as filler for everything.
Might be a string, STDIN or Filename / Pathname.
I think the name is horribly misleading and we should change it.
Suggestion: source_via
Since this "description" terminology is used across the application, I think it makes sense to discuss this first.
WDYT?

troessner:
Fixes #604

### DIFF
--- a/lib/reek/source/source_code.rb
+++ b/lib/reek/source/source_code.rb
@@ -14,17 +14,17 @@ module Reek
       IO_IDENTIFIER     = 'STDIN'
       STRING_IDENTIFIER = 'string'
 
-      attr_reader :description
+      attr_reader :origin
 
       # Initializer.
       #
-      # code        - Ruby code as String
-      # description - 'STDIN', 'string' or a filepath as String
-      # parser      - the parser to use for generating AST's out of the given source
-      def initialize(code, description, parser: Parser::Ruby22)
-        @source      = code
-        @description = description
-        @parser      = parser
+      # code   - Ruby code as String
+      # origin - 'STDIN', 'string' or a filepath as String
+      # parser - the parser to use for generating AST's out of the given source
+      def initialize(code, origin, parser: Parser::Ruby22)
+        @source = code
+        @origin = origin
+        @parser = parser
       end
 
       # Initializes an instance of SourceCode given a source.
@@ -80,9 +80,9 @@ module Reek
         @syntax_tree ||=
           begin
             begin
-              ast, comments = parser.parse_with_comments(source, description)
+              ast, comments = parser.parse_with_comments(source, origin)
             rescue Racc::ParseError, Parser::SyntaxError => error
-              $stderr.puts "#{description}: #{error.class.name}: #{error}"
+              $stderr.puts "#{origin}: #{error.class.name}: #{error}"
             end
 
             # See https://whitequark.github.io/parser/Parser/Source/Comment/Associator.html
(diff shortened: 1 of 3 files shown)


lib/reek/source/source_code.rb
```diff
@@ -16,16 +16,16 @@ class SourceCode
...
       # code        - Ruby code as String
-      # description - 'STDIN', 'string' or a filepath as String
+      # origin - 'STDIN', 'string' or a filepath as String
       # parser      - the parser to use for generating AST's out of the given source
-      def initialize(code, description, parser = Parser::Ruby22)
+      def initialize(code, origin, parser = Parser::Ruby22)
         @source      = code
-        @description = description
+        @origin = origin
         @parser      = parser
```
chastell: Can you please adjust the whitespace in the comment and in the method?

chastell:
I like it, but I’m not sure what’s the distinction between ‘origin’ and ‘target’ – i.e., why not have ‘origin’ everywhere.

lib/reek/source/source_code.rb
troessner: Fixed!

troessner:
I like it, but I’m not sure what’s the distinction between ‘origin’ and ‘target’ – i.e., why not have ‘origin’ everywhere.

`origin` in a SmellRepository would at least for me mean "where is this SmellRepository coming, e.g. where do we load it from?" (doesn't make much sense if you know what it does, but i am assuming the unsuspecting reader here). `target` tells me "we will apply this SmellRepository to this target", which is exactly what we're doing.
Does this make sense to you? If not I can also change it to `origin` everywhere for consistency reasons.:)

chastell:
Hmmm. I was about to propose `source_origin` (to make the association more obvious), but the more I think about it the more I’m ok with `target`. :)

I’ll merge this when it builds, but it’s past 1 am and neither build started yet and I have a busy day coming. If @mvz beats me to it then by all means merge (or weigh on the above). :)

mvz:
This is a breaking change that should go into reek 4, right?

mvz:
Also ... I'm not too sure about the names `origin` and `target`. I would prefer something more explicit. Maybe `source_name` or something?

Also ... I just pushed my todos branch at #651 for a more radical solution.

troessner:
>> This is a breaking change that should go into reek 4, right?

This part of Examiner and friends has never been communicated [here](https://github.com/troessner/reek/blob/master/docs/API.md), so for me this is not part of our public API...

>> Also ... I just pushed my todos branch at #651 for a more radical solution.

But there it is stil named "description"? 
-> https://github.com/troessner/reek/pull/651/files#diff-3eadf674a9c3b7a7864e5247c51bd801R71

mvz:
> > Also ... I just pushed my todos branch at #651 for a more radical solution.
> But there it is still named "description"? 

True, but it removes this value from SourceRepository entirely.

troessner:
>> True, but it removes this value from SourceRepository entirely.

Yep, but it still has the annoying "description" in it. :laughing: 

I'll leave this branch for reference here, let's get #651 merged and then come back to this branch to remove it completely.

troessner:
Alrightie, I rebased and made the scope of this whole PR significantly smaller!

>> This is a breaking change that should go into reek 4, right?

Nope, not anymore ;)

Please re-review!

chastell:
:+1:

PR #649, opened 2015-08-19
Remove obsolete docs/Configuration-Files.md (because it's now directly in the README).

troessner:
In one of our latest "improve our documentation" pull requests (can't remember which one) we kind of decided to rather have the configuration documentation in our README for conversion reasons but apparently we forgot to delete this now completely outdated doc file.


chastell:
…and it builds on first try, even on Rubinius! Merging!

Issue #650, opened 2015-08-19
Breaking change for reek 4: Clean up Examiner constructor


troessner:
....because right now it has this signature:

```Ruby
    def initialize(source,
                   filter_by_smells = [],
                   configuration: Configuration::AppConfiguration.default)
```

Let's use keyword arguments everywhere.

chastell:
:+1:

mvz:
I'm have a branch with some cleanups that includes a change here. I'll push it soon so it can be reviewed.

mvz:
I just realized that branch contains no changes to Examiner#initialize.

Anyway, :+1: from me.

Once #651 is merged, we can extract the smell detector selection out of this class and just pass in a SmellRepository.

PR #651, opened 2015-08-20
Todos and fixmes

mvz:
This is a branch where I collected some nits I wanted to pick, but it got out of hand and now contains a breaking change that should go into Reek 4.

Highlights:
* Makes smell detectors not depend on the source to examine. They just get passed contexts, which could be all from the same source, or from different sources.
* Uses the parse result to fetch the source name. No more passing description/origin/target :smile:.
* Fully configure smell detectors on initialization, instead of using `#configure_with`.
* More FIXMEs fixed (and some added).

See #650 and #648.

### DIFF
--- a/lib/reek/context/method_context.rb
+++ b/lib/reek/context/method_context.rb
@@ -9,11 +9,6 @@ module Reek
     class MethodContext < CodeContext
       attr_reader :refs
 
-      def envious_receivers
-        return {} if refs.self_is_max?
-        refs.most_popular
-      end
-
       def references_self?
         exp.depends_on_instance?
       end
--- a/lib/reek/smells/smell_detector.rb
+++ b/lib/reek/smells/smell_detector.rb
@@ -14,8 +14,6 @@ module Reek
     #
     # @api private
     class SmellDetector
-      attr_reader :source
-
       # The name of the config field that lists the names of code contexts
       # that should not be checked. Add this field to the config for each
       # smell that should ignore this code element.
@@ -74,10 +72,8 @@ module Reek
         end
       end
 
-      attr_reader :smells_found # SMELL: only published for tests
-
-      def initialize(source, config = self.class.default_config)
-        @source = source
+      def initialize(config = {})
+        config = self.class.default_config.merge(config)
         @config = SmellConfiguration.new(config)
         @smells_found = []
       end
@@ -85,15 +81,6 @@ module Reek
       def register(hooks)
         return unless config.enabled?
         self.class.contexts.each { |ctx| hooks[ctx] << self }
-      end
-
-      # SMELL: Getter (only used in 1 test)
-      def enabled?
-        config.enabled?
-      end
-
-      def configure_with(new_config)
-        config.merge!(new_config)
       end
 
       def examine(context)
@@ -104,16 +91,23 @@ module Reek
         self.smells_found += sm
       end
 
-      def enabled_for?(context)
-        enabled? && config_for(context)[SmellConfiguration::ENABLED_KEY] != false
+      def report_on(report)
+        smells_found.each { |smell| smell.report_on(report) }
       end
 
       def exception?(context)
         context.matches?(value(EXCLUDE_KEY, context, DEFAULT_EXCLUDE_SET))
       end
 
-      def report_on(report)
-        smells_found.each { |smell| smell.report_on(report) }
+      protected
+
+      # NOTE: Needs to be protected so += works for Ruby < 2.2
+      attr_accessor :smells_found
+
+      private
+
+      def enabled_for?(context)
+        config.enabled? && config_for(context)[SmellConfiguration::ENABLED_KEY] != false
       end
 
       def value(key, ctx, fall_back)
@@ -124,11 +118,17 @@ module Reek
         ctx.config_for(self.class)
       end
 
-      protected
-
-      attr_writer :smells_found
-
-      private
+      def smell_warning(options = {})
+        context = options.fetch(:context)
+        exp = context.exp
+        ctx_source = exp.loc.expression.source_buffer.name
+        SmellWarning.new(self,
+                         source: ctx_source,
+                         context: context.full_name,
+                         lines: options.fetch(:lines),
+                         message: options.fetch(:message),
+                         parameters: options.fetch(:parameters, {}))
+      end
 
       private_attr_reader :config
     end
--- a/lib/reek/smells/smell_warning.rb
+++ b/lib/reek/smells/smell_warning.rb
@@ -8,13 +8,15 @@ module Reek
     class SmellWarning
       include Comparable
       extend Forwardable
-      attr_reader :context, :lines, :message, :parameters, :smell_detector
-      def_delegators :smell_detector, :smell_category, :smell_type, :source
+      attr_reader :context, :lines, :message, :parameters, :smell_detector, :source
+      attr_reader :source
+      def_delegators :smell_detector, :smell_category, :smell_type
 
       # FIXME: switch to required kwargs when dropping Ruby 2.0 compatibility
       def initialize(smell_detector, context: '', lines: raise, message: raise,
-                                     parameters: {})
+                                     source: raise, parameters: {})
         @smell_detector = smell_detector
+        @source         = source
         @context        = context.to_s
         @lines          = lines
         @message        = message
@@ -43,11 +45,10 @@ module Reek
         listener.found_smell(self)
       end
 
-      def yaml_hash(warning_formatter = nil)
+      def yaml_hash
         stringified_params = Hash[parameters.map { |key, val| [key.to_s, val] }]
         core_yaml_hash.
-          merge(stringified_params).
-          merge(wiki_link_hash(warning_formatter))
+          merge(stringified_params)
       end
 
       protected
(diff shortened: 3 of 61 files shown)


troessner:
Does it also make reek smell-free? :trollface:

lib/reek/context/method_context.rb
```diff
@@ -9,6 +9,7 @@ module Context
     class MethodContext < CodeContext
       attr_reader :refs
 
+      # TODO: Move to FeatureEnvy
```
troessner: YES!

lib/reek/configuration/configuration_file_finder.rb
```diff
@@ -18,16 +18,11 @@ class ConfigFileException < StandardError; end
...
         load_from_file(find(params))
       end
 
-      # FIXME: switch to kwargs on upgrade to Ruby 2 and drop `params.fetch` calls:
-      # def find(options: nil, current: Pathname.pwd, home: Pathname.new(Dir.home))
-      def find(params = {})
-        options = params.fetch(:options) { nil                    }
-        current = params.fetch(:current) { Pathname.pwd           }
-        home    = params.fetch(:home)    { Pathname.new(Dir.home) }
+      def find(options: nil, current: Pathname.pwd, home: Pathname.new(Dir.home))
```
troessner: This is already fixed in master: https://github.com/troessner/reek/blob/master/lib/reek/configuration/configuration_file_finder.rb#L25

lib/reek/configuration/configuration_file_finder.rb
```diff
@@ -18,16 +18,11 @@ class ConfigFileException < StandardError; end
     module ConfigurationFileFinder
       module_function
 
-      def find_and_load(params = {})
+      def find_and_load(**params)
```
troessner: :+1: - let's use kw arguments where we use hashes.

lib/reek/smells/smell_detector.rb
```diff
@@ -83,14 +81,11 @@ def register(hooks)
       end
 
       # SMELL: Getter (only used in 1 test)
+      # MvZ: Or use it more :-).
```
troessner: Or let's kill it.:)

lib/reek/smells/smell_detector.rb
```diff
@@ -118,6 +113,18 @@ def value(key, ctx, fall_back)
       def config_for(ctx)
         ctx.config_for(self.class)
       end
+
+      def create_smell_warning(options = {})
```
troessner: Can we drop the "create_" terminology and just use "smell_warning"? Feels java-esque :-(

troessner:
I started commenting even though that thing says "wip". Why? Because Yolo, that's why.

mvz:
> I started commenting even though that thing says "wip". Why? Because Yolo, that's why.

And I'm glad you did!

lib/reek/smells/smell_detector.rb
mvz: Hm, seems like there's lots of weirdness going on with the config stuff. I'd like to unify this so it always goes through a SmellConfiguration object rather than having the current mix of SmellConfiguration and Hash.

lib/reek/smells/smell_detector.rb
mvz: Done!

mvz:
Improves Reek's smell count from 118 to 108!

lib/reek/smells/smell_detector.rb
```diff
@@ -76,8 +74,8 @@ def default_smell_category
 
       attr_reader :smells_found # SMELL: only published for tests
 
-      def initialize(source, config = self.class.default_config)
-        @source = source
+      def initialize(config = {})
```
troessner: Yes, very nice, that we now separate this from "source".

lib/reek/smells/smell_detector.rb
```diff
@@ -104,16 +93,23 @@ def examine(context)
...
-        smells_found.each { |smell| smell.report_on(report) }
+      protected
+
+      # NOTE: Writers cannot be private in Ruby < 2.2
+      attr_writer :smells_found
+
+      private
+
+      def enabled_for?(context)
+        config.enabled? && config_for(context)[SmellConfiguration::ENABLED_KEY] != false
```
troessner: If we just use this in 1 tests, how about refactoring this test and get rid of it?

lib/reek/smells/smell_detector.rb
```diff
@@ -104,16 +93,23 @@ def examine(context)
...
       def exception?(context)
         context.matches?(value(EXCLUDE_KEY, context, DEFAULT_EXCLUDE_SET))
       end
 
-      def report_on(report)
-        smells_found.each { |smell| smell.report_on(report) }
+      protected
+
+      # NOTE: Writers cannot be private in Ruby < 2.2
+      attr_writer :smells_found
```
troessner: Why not use 

````Ruby
private_attr_writer :smells_found
```

?

lib/reek/smells/smell_warning.rb
```diff
@@ -83,13 +85,14 @@ def core_yaml_hash
...
-          'smell_type'     => smell_detector.smell_type,
-          'source'         => smell_detector.source
+          'smell_category' => smell_category,
+          'smell_type'     => smell_type,
+          'source'         => source
         }
       end
 
       def wiki_link_hash(warning_formatter)
+        # SMELL:
```
troessner: What smell?

spec/factories/factories.rb
```diff
@@ -19,15 +27,18 @@
...
-      new(smell_detector, context: context,
-                          lines: lines,
-                          message: message,
-                          parameters: parameters)
+      new(smell_detector,
+          source: source,
+          context: context,
+          lines: lines,
+          message: message,
+          parameters: parameters)
```
troessner: Just a  random note: I really think we should refactor our specs to use factories for all-the-things. 
That would clean up our specs tremendously.
Maybe something to keep in mind..:)

troessner:
>> Improves Reek's smell count from 118 to 108!

Awesome!

lib/reek/smells/smell_detector.rb
mvz: Which bit are you talking about here? `enabled_for` is used internally.

lib/reek/smells/smell_detector.rb
mvz: Because that causes `self.smells_found = 0` not to work in Ruby < 2.2.

lib/reek/smells/smell_warning.rb
mvz: I was going to explain it, but I'll fix it instead.

lib/reek/smells/smell_detector.rb
```diff
@@ -76,8 +74,8 @@ def default_smell_category
 
       attr_reader :smells_found # SMELL: only published for tests
```
troessner: If we just use this in 1 tests, how about refactoring this test and get rid of it?

lib/reek/smells/smell_detector.rb
troessner: Arrgh, i had written this comment while commuting on my phone and apparently I slipped some lines down :)

See my new comment above: https://github.com/troessner/reek/pull/651/files#diff-5b6de37fbe6e3791a90ceaf86e847935R75

lib/reek/smells/smell_detector.rb
troessner: >> Because that causes self.smells_found = 0 not to work in Ruby < 2.2.

I don't follow :(

>>   ( $ (master)) grep private_attr_writer lib/ -r
lib//reek/configuration/app_configuration.rb:      private_attr_writer :directory_directives, :default_directive, :excluded_paths
lib//reek/context/code_context.rb:      private_attr_writer :tracked_visibility

Those writer work with all ruby versions we run on travis.

mvz: > Those writer work with all ruby versions we run on travis.

Indeed. Something else is going on. The following demonstrates the problem.

```ruby
class Foo
  def initialize
    self.foo = 1
  end
  def bar
    self.foo += 3
  end

  attr_reader :foo

  private

  attr_writer :foo
end

foo = Foo.new
foo.bar
```

This fails on Ruby 2.1 and 2.0. It seems this is due to how the `+=` operator is handled, because it indeed works fine with just straight assignment.

I'll update the note.

lib/reek/smells/smell_detector.rb
mvz: Done!

lib/reek/smells/smell_warning.rb
```diff
@@ -8,11 +8,13 @@ module Smells
...
       extend Forwardable
-      attr_reader :context, :lines, :message, :parameters, :smell_detector
-      def_delegators :smell_detector, :smell_category, :smell_type, :source
+      attr_reader :context, :lines, :message, :parameters, :smell_detector, :source
+      attr_reader :source
+      def_delegators :smell_detector, :smell_category, :smell_type
 
       def initialize(smell_detector, options = {})
         @smell_detector = smell_detector
+        @source         = options.fetch(:source)
```
mvz: This technically is an API change, but we do not describe smell warning creation in the API docs.

troessner: But then it's not part of our API ;)

troessner:
Can you please squash and rebase? Then I'll happily merge!

mvz:
Squashed and rebased.

PR #652, opened 2015-08-20
Prepare 3.3.0 release.

troessner:


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+## 3.3.0 (2015-08-20)
+
+* (HParker / Adam Hess) add support for or-assign constants
+* (troessner) Update TooManyMethods default value to 15
```
troessner: I think this warrants a minor version upgrade.

mvz: The default value? Not sure ... I think this falls under 'Consistency of detected smells', which only needs a point release.

troessner: But when I, as a reek user, upgrade reek to the next patch version I certainly do NOT expect my CI build to break because we changed smell configurations. And this change will break a huge number of CI builds.

mvz: > But when I, as a reek user, upgrade reek to the next patch version I certainly do NOT expect my CI build to break because we changed smell configurations

Good point. I agree and we should update our versioning policy accordingly.

mvz:
I'm seeing some loading issues in the master version, so I don't think we should release this just yet.

Perhaps we should start creating pre-releases to shake out final bugs before each minor and major release.

chastell:
FWIW, I bumped into that Psych issue on each clean install of a new Ruby version, so I doubt that one is in any way related to reek.

chastell:
Made the build green – but we should probably merge #654 first (so this includes it) and adjust the release date.

mvz:
> I bumped into that Psych issue on each clean install of a new Ruby version, so I doubt that one is in any way related to reek.

No, it definitely isn't, but there was another loading issue: `stringio`. I'll make a PR.

troessner:
>> No, it definitely isn't, but there was another loading issue: stringio. I'll make a PR.

Merged!

Alright guys, let's get 3.3.0 going! Objections?

chastell:
None. I bumped the date in `CHANGELOG.md` to today’s and will merge as soon as it builds.

Issue #653, opened 2015-08-20
private method `load' called for Psych:Module


mvz:
I'm getting this error when running reek on reek:

```
code_comment.rb:31:in `add_to_config': private method `load' called for Psych:Module (NoMethodError)
```

mvz:
This does not occur when running with bundler.

mvz:
Right:

```
[matijs@walnut] ruby -e "require 'yaml'; puts YAML.methods(false).count"
30
[matijs@walnut] ruby -e "require 'reek'; puts YAML.methods(false).count"
1
```

mvz:
Lovely:
```
[matijs@walnut] bundle exec ruby -e "require 'reek'; puts YAML.methods(false).count"
30
```

mvz:
Here's a comparison of features loaded in each case:

```
[matijs@walnut] ruby -w -e "old = \$LOADED_FEATURES.dup; require 'reek/code_comment'; new = \$LOADED_FEATURES.dup; p new - old"
["/home/matijs/.rbenv/versions/2.2.2/lib/ruby/2.2.0/rubygems/path_support.rb",
 "/home/matijs/.rbenv/versions/2.2.2/lib/ruby/2.2.0/rubygems/dependency.rb",
 "/home/matijs/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/psych-2.0.13/lib/psych/exception.rb",
 "/home/matijs/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/psych-2.0.13/lib/psych/syntax_error.rb",
 "/home/matijs/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/extensions/x86_64-linux/2.2.0-static/psych-2.0.13/psych.so",
 "/home/matijs/.rbenv/versions/2.2.2/lib/ruby/2.2.0/yaml.rb",
 "/home/matijs/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/private_attr-1.1.0/lib/private_attr/version.rb",
 "/home/matijs/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/private_attr-1.1.0/lib/private_attr.rb",
 "/home/matijs/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/private_attr-1.1.0/lib/private_attr/everywhere.rb",
[...]

PR #654, opened 2015-08-21
Ensure full psych is loaded

mvz:
This is a workaround for #653 until https://github.com/tenderlove/psych/issues/223 is fixed.

I think we should release this as `3.2.2`, so I created a `stable-3.2` branch to handle this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ## Unreleased
 
+* (mvz) Ensure all YAML methods are loaded (workaround for #653)
+
 ## 3.2.1 (2015-08-17)
 
-* Revert 864f0a9 to hotfix issues/642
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 
 ## Unreleased
 
+* (mvz) Add workaround for issues/653
```
troessner: I'd prefer it to have a description here. If we just use issue numbers we can throw away the changelog and use git again :)

mvz: Good point.

chastell:
In the past [we used the `#xxx` notation to mean issue numbers](https://github.com/troessner/reek/blob/df4156056817ccc51da28947568e9debb264ef66/CHANGELOG.md#minor-changes) – can we keep it consistent?

mvz:
Updated.

PR #657, opened 2015-08-22
Use of StringIO requires stringio

mvz:
This doesn't show up in our tests. Perhaps because of bundler.


chastell:
I got bitten by this in other projects in the past; I think it was rather because RSpec requires it in the meantime.

(I really need to write that linter that goes through all constants in a Ruby file and whines if they’re not defined in one of its requires. Maybe as an optional reek smell.)

PR #658, opened 2015-08-22
Ensure each file can be individually required

mvz:
We've always been careful to ensure each file `require`s its own requirements, but with `private_attr` we've failed to do so. This makes it hard to require individual reek files if needed.

This PR makes sure that for every file in reek's lib, `ruby -e "require '<file>'"` works.

### DIFF
--- a/lib/reek/cli/command.rb
+++ b/lib/reek/cli/command.rb
@@ -1,3 +1,5 @@
+require 'private_attr/everywhere'
+
 module Reek
   module CLI
     #
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,4 @@
 require 'bundler/gem_tasks'
-require 'private_attr/everywhere'
 require 'rake/clean'
 
 Dir['tasks/**/*.rake'].each { |t| load t }
--- a/lib/reek.rb
+++ b/lib/reek.rb
@@ -1,7 +1,6 @@
 #
 # Reek's core functionality
 #
-require 'private_attr/everywhere'
 require_relative 'reek/version'
 require_relative 'reek/examiner'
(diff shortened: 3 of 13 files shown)


mvz:
Suggestions for how to test this and ensure it for the future are welcome :smile:.

lib/reek/cli/command.rb
```diff
@@ -1,3 +1,4 @@
+require 'private_attr/everywhere'
```
troessner: Newline missing here and below.

mvz: Fixed.

troessner:
Travis....lets make rubinius an allowed failure in general. I am really fed up with this shit.

mvz:
> Travis....lets make rubinius an allowed failure in general. I am really fed up with this shit.

Agreed.

chastell:
Agreed on Rubinius as optional in general, but I _think_ currently the only reason why Rubinius and JRuby fail now-and-then is because of #631 – they just sometimes hit a parsing timeout. We should probably investigate (after we make Rubinius optional). :)

chastell:
(I’m not sure about the timeout being the exact culprit, but it seems to always be that particular Cucumber feature which fails.)

chastell:
> Suggestions for how to test this and ensure it for the future are welcome

Well, running a `ruby -e "require '<file>'"` subshell for each of our files is quite doable, but it doesn’t actually address the problem in the method bodies – missing constants (like `StringIO`) will be only discovered when actually executing a method. (Hence my idea of ‘parse all constants in the current file, check whether they’re all either from Ruby core or defined in the required files.)

PR #659, opened 2015-08-22
Remove spec/gem directory.

troessner:
We never used this and in general this is a classical case of over-spec'ing for me.
Let's kill it.


chastell:
…with :fire:

PR #664, opened 2015-08-23
Ensure full psych is loaded

mvz:
So .. #654 got merged but not released so I'm making a new PR for master to get this in the next release.


troessner:
Will merge when green.

PR #665, opened 2015-08-23
Simplify CodeComment implementation

mvz:


chastell:
:+1:

lib/reek/code_comment.rb
```diff
@@ -11,31 +11,26 @@ class CodeComment
...
 
-    protected
+    private
 
     def add_to_config(smell, options)
       options ||= ': { enabled: false }'
-      config.merge! YAML.load(smell.gsub(/(?:^|_)(.)/) { $1.upcase } + options)
-      # TODO: extend this to all configs -------------------^
-      # TODO: extend to allow configuration of whole smell class, not just subclass
+      YAML.load(smell.gsub(/(?:^|_)(.)/) { $1.upcase } + options)
```
troessner: Since we're already on it, could we maybe also refactor this mega-ugly line of code?

chastell: Totally unrelated (as here this is a non-capturing group), but ever since I learned that the ternary logic operator – `foo ? bar : baz` – is called the Elvis operator I have an eye for the `?:` sequence. Here it looks like he’ś even singing.

(I’d vote for dropping the non-capturing here and going with `$2` instead; `/(^|_)(.)/` is more readable IMHO.)

troessner:
Very nice! Can you please squash? Then I'll merge this one.

PR #666, opened 2015-08-23
Switch to keyword arguments

chastell:
I started to work on the ‘switch reek to keyword arguments’ idea. Let me know whether the `Report` example is the right direction.

I also added the ‘work in progress’ label, as I don’t really like pseudo-tags in PR names – but if you’re fond of ‘[WIP]’ then I’ll cave in. ;)

### DIFF
--- a/lib/reek/context/code_context.rb
+++ b/lib/reek/context/code_context.rb
@@ -137,7 +137,7 @@ module Reek
       #
       # @param visibility [Symbol]
       # @param names [Array<Symbol>]
-      def track_visibility(visibility, names = [])
+      def track_visibility(visibility, names)
         if names.any?
           children.each do |child|
             child.visibility = visibility if names.include? child.name
--- a/lib/reek/report/report.rb
+++ b/lib/reek/report/report.rb
@@ -16,13 +16,15 @@ module Reek
       # @api private
       WARNINGS_COLOR = :red
 
-      def initialize(options = {})
+      def initialize(heading_formatter: HeadingFormatter::Quiet,
+                     report_formatter: Formatter, sort_by_issue_count: false,
+                     warning_formatter: SimpleWarningFormatter.new)
         @examiners           = []
+        @heading_formatter   = heading_formatter.new(report_formatter)
+        @report_formatter    = report_formatter
+        @sort_by_issue_count = sort_by_issue_count
         @total_smell_count   = 0
-        @options             = options
-        @warning_formatter   = options.fetch :warning_formatter, SimpleWarningFormatter.new
-        @report_formatter    = options.fetch :report_formatter, Formatter
-        @sort_by_issue_count = options.fetch :sort_by_issue_count, false
+        @warning_formatter   = warning_formatter
       end
 
       # Add Examiner to report on. The report will output results for all
@@ -56,7 +58,7 @@ module Reek
 
       private
 
-      private_attr_reader :examiners, :options, :report_formatter,
+      private_attr_reader :examiners, :heading_formatter, :report_formatter,
                           :sort_by_issue_count, :warning_formatter
     end
 
@@ -89,7 +91,7 @@ module Reek
         result = heading_formatter.header(examiner)
         if examiner.smelly?
           formatted_list = report_formatter.format_list(examiner.smells,
-                                                        warning_formatter)
+                                                        formatter: warning_formatter)
           result += ":\n#{formatted_list}"
         end
         result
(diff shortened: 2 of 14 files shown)


troessner:
>> I also added the ‘work in progress’ label, as I don’t really like pseudo-tags in PR names – but if you’re fond of ‘[WIP]’ then I’ll cave in. ;)

Nope, that's coolio.:)

>> I started to work on the ‘switch reek to keyword arguments’ idea. Let me know whether the Report example is the right direction.

Yes, it is!

chastell:
Ready to review. I tracked all methods that take a parameter that defaults to a `Hash`, all methods that have an `options` or `hash` parameters and all methods that have deafults; surely there are more methods to switch, but with Ruby 2.0 not supporting required kwargs it’ll be a job that will need refixing later anyway (see the `SmellWarning` change).

As some of these can be controversial I put them in separate commits, for ease of reversing. I’m not fond of any one in particular, so feel free to veto any of the changes.

lib/reek/report/report.rb
```diff
@@ -139,7 +136,7 @@ def show
     class HTMLReport < Base
       require 'erb'
 
-      def show(target_path = Pathname.new('reek.html'))
+      def show(target_path: Pathname.new('reek.html'))
```
mvz: I'm not too sure about the cases where it's just one optional argument.

lib/reek/smells/nil_check.rb
```diff
@@ -30,7 +30,7 @@ def examine_context(ctx)
       # A base class that allows to work on all nodes of a certain type.
       #
       class NodeFinder
-        def initialize(ctx, type, detector = nil)
+        def initialize(ctx, type, detector: nil)
```
mvz: I'd make detector a required argument instead. The case where detector is nil blows up anyway.

lib/reek/smells/smell_warning.rb
```diff
@@ -11,12 +11,14 @@ class SmellWarning
       attr_reader :context, :lines, :message, :parameters, :smell_detector
       def_delegators :smell_detector, :smell_category, :smell_type, :source
 
-      def initialize(smell_detector, options = {})
+      # FIXME: switch to required kwargs when dropping Ruby 2.0 compatibility
+      def initialize(smell_detector, context: '', lines: raise, message: raise,
+                                     parameters: {})
```
mvz: :+1:. Nice trick with raise.

lib/reek/context/code_context.rb
```diff
@@ -137,7 +137,7 @@ def config_for(detector_class)
       #
       # @param visibility [Symbol]
       # @param names [Array<Symbol>]
-      def track_visibility(visibility, names = [])
+      def track_visibility(visibility, names: [])
```
mvz: Let's make names a required argument since it's never called without names in the actual code.

spec/reek/source/source_code_spec.rb
```diff
@@ -30,7 +30,7 @@
     let(:source_name) { 'Test source' }
     let(:error_message) { 'Error message' }
     let(:parser) { double('parser') }
-    let(:src) { Reek::Source::SourceCode.new('', source_name, parser) }
+    let(:src) { Reek::Source::SourceCode.new('', source_name, parser: parser) }
```
mvz: This makes the dependency injection stand out more. Nice!

lib/reek/report/report.rb
chastell: My reasoning is: positional arguments in public methods are a strong API restriction, and if they are there they should be the obvious ones (like `from_path(path)`). If something is passed solely (or mostly) for dependency injection / testing then making it a keyword argument means it won’t steal the precious First Positional Argument spot.

(That said, I’m open for reverting this on behalf of ‘this is the obvious argument to this method’; I’m just addressing the broader comment.) :)

lib/reek/smells/nil_check.rb
chastell: A required positional argument or a required keyword argument?

mvz: A required positional argument.

lib/reek/smells/smell_warning.rb
chastell: :) I used to write the below (and do stuff like `lines: req(:lines)`) in my codebases that needed to support Ruby 2.0, but that was a bit too much. ;)

req

chastell:
(Green and ready for a re-review.)

lib/reek/report/report.rb
troessner: With @chastell on this one ;)

lib/reek/smells/smell_warning.rb
troessner: >> :) I used to write the below (and do stuff like lines: req(:lines)) in my codebases that needed to support Ruby 2.0, but that was a bit too much. ;)

:laughing:

troessner:
Let's get this merged! @chastell can you please squash your 10 gazillion commits please?

lib/reek/report/report.rb
mvz: I agree with the API/dependency injection heuristic. It makes a lot of sense.

chastell:
Squashed down to two unrelated ones, I can squash further if you so desire. :)

PR #670, opened 2015-08-25
Attribute no longer disabled by default

tansaku:
I noticed that as of the latest gem that Attribute is no longer disabled by default so thought it a good idea to submit a pull request to update the docs - apologies in advance if this is not the correct procedure to adjust the docs


mvz:
Yes, this is exacly the correct procedure :smile:.

Thanks!

PR #677, opened 2015-08-31
Fix a bunch of smells.

troessner:
Ladies and gentlemen,

may I present to you the ultimate "Make reek smell-free" pull request!

This little honeybadger you see below is still in its infancy but fixes already a dozen warnings (or disables them rightfully).

I won't stop till I drop and then we can finally use "rake test:quality" on travis!

### DIFF
--- a/lib/reek/ast/node.rb
+++ b/lib/reek/ast/node.rb
@@ -37,7 +37,7 @@ module Reek
       end
 
       def line
-        loc.line unless loc.nil?
+        loc && loc.line
       end
 
       # @deprecated
--- a/lib/reek/ast/object_refs.rb
+++ b/lib/reek/ast/object_refs.rb
@@ -3,7 +3,7 @@ require 'private_attr/everywhere'
 module Reek
   # Represents functionality related to an Abstract Syntax Tree.
   module AST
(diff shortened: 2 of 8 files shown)


troessner:
Wanting to fix all smells in one PR was a bad plan as I now realize.

Way smarter: step by step.

So here's the first PR in an series ;)

Review please!

chastell:
I used to put exceptions in a `.reek` file, but recently started moving them to code comments – this way (a) they move along with the method and (b) we could eventually make reek smell-free with the default config.

Also, rather than excluding DuplicateMethodCall I add a count, so that when it goes over the method is smelly again.

chastell:
LGTM otherwise. :)

troessner:
>> I used to put exceptions in a .reek file, but recently started moving them to code comments – this way (a) they move along with the method and (b) we could eventually make reek smell-free with the default config.

Good point! I removed the reek config again and added comments accordingly.

>> Also, rather than excluding DuplicateMethodCall I add a count, so that when it g oes over the method is smelly again.

Added!

troessner:
Please re-review.;)

chastell:
Ausgezeichnet!

Issue #675, opened 2015-08-31
Remove  # @api private comments


troessner:
.....because they:

-  decrease readabiity 
- are counter-intuitive. Because "api private" is the regular case, not the exception. We should rather list exceptions and not the other way round
- is repetitive and feels un-rubyesque

Objections?

chastell:
Yes, please!

troessner:
#678 -> fix ready.

mvz:
See also #578.

mvz:
Done!

Issue #578, opened 2015-07-05
Getting rid of YARD


troessner:
Coming from: https://github.com/troessner/reek/pull/575#issuecomment-118663161
makes me think that YARD is not the right tool for the job.
Do you guys agree and / or do we have an alternative for discussion?

mvz:
I think this would be too drastic. I think YARD is a really nice documentation tool. The problem seems to be with defining our public API. When implementing that, the simplest way to me seemed to be to mark what was part of our 'private' API, but perhaps that was the wrong thing to do. I'd like to look into how other projects deal with that and see if we can find a better solution.

troessner:
>>  I'd like to look into how other projects deal with that and see if we can find a better solution.

Looking forward to this..:)

mvz:
One option is to add `@api public` to all classes that are part of the official API, and then put `--api public` in `.yardopts`. The reason for the latter is that YARD doesn't park the public API in its output. It only does that with the bits marked `@api private`.

Advantage: We only need to mark our public API, instead of everything that's not.

Disadvantage: It becomes difficult to output our entire documentation, since undoing the `--api public` flag is hard or impossible.

An alternative to setting `--api public` would be to create another YARD plugin that does output some text for all items that are marked `@api public`. This actually sounds like something other people might want to use.

mvz:
The more I think about it the more I like the plugin option. It will allow us to have a nice friendly message on objects and methods that are safe to use, instead of an unpleasent message elsewhere.

troessner:
>> Disadvantage: It becomes difficult to output our entire documentation, since undoing the --api public flag is hard or impossible.

Why would we want to output our entire documentation? I fail to see how this is relevant for any non-reek contributors and we just read the documentation when looking at the code.
What's our use case here?

mvz:
> What's our use case here?

Mainly to allow people working on reek (like ourselves) to read the documentation in a nice format. I agree it's not a hugely important use case.

If we agree marking the public API instead of the private one is the way to go, I'll start a branch for that.

troessner:
>> If we agree marking the public API instead of the private one is the way to go, I'll start a branch for that.

:+1: for that. @chastell ?

chastell:
Oh sure, go ahead!

troessner:
>> If we agree marking the public API instead of the private one is the way to go, I'll start a branch for that.

Any updates on this? Sorry for being a pain in the ass but this "api private" marker in our source code annoys me every time I see this...

[...]

PR #678, opened 2015-09-01
Remove annoying "@api private" comments.

troessner:


troessner:
fixes #675

troessner:
travis fails because of a rubocop update, fix up in #679

chastell:
I merged the other PR, but I’d rather @mvz makes a call on this one – I’m not sure what the `@api private` was for but I think we used this in YARD?

mvz:
> I’m not sure what the `@api private` was for but I think we used this in YARD?

Correct. I'm still planning to research an alternative solution (basically, where we mark what's public instead.

mvz:
TL;DR: I'll create a PR for the solution I proposed in #578, but this can be merged in the mean time once it's green.

troessner:
Something is really f**ked up with our rubocop set up:

>> lib/reek/ast/object_refs.rb:4:3: C: Style/Documentation: Missing top-level module documentation comment.
  module AST
  ^^^^^^

But I can't reproduce this locally, neither with the version that is in our gemspec nor with the latest version online. Grrarrrrrr.

chastell:
I can replicate this with RuboCop 0.33.0 – what would be a good comment for `Reek::AST`? :)

troessner:
>> I can replicate this with RuboCop 0.33.0 – what would be a good comment for Reek::AST? :)

Super weird. I can reproduce this now as well, but only on my new mac, not my old. WTF. 
Let me retry.

troessner:
Green! LGTM? ;)

mvz:
:+1:

PR #680, opened 2015-09-01
Ignore `Object#tap` for NestedIterators.

### LINKED ISSUE #676: Make NestedIterators ignore Object#tap
Object#tap should be an exception to the rule because it's no smell to do things like:

```Ruby
class Grmpff
  def gruml
    {}.tap do |hash|
      hash[:foo] = [:holla, :die, :waldfee]
      hash.each do |key, value|
        puts value
      end
    end
  end
end
````

What do you guys think?

troessner:
Fixes #676

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* (troessner) Ignore `Object#tap` for NestedIterators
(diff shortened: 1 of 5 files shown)


troessner:
Killing reeks smell warnings bit by bit! Slaying the dragon!

chastell:
Ausgezeichnet!

Issue #682, opened 2015-09-03
Single smell option broken


mvz:
It seems reek checks the config against the current set in the smell repository rather than the set of all existing detectors:

```
[matijs@walnut] reek --smell UtilityFunction lib
/home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/smells/smell_repository.rb:30:in `configure': Unknown smell type Reek::Smells::UncommunicativeMethodName found in configuration (ArgumentError)
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/smells/smell_repository.rb:24:in `block in initialize'
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/smells/smell_repository.rb:23:in `each'
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/smells/smell_repository.rb:23:in `initialize'
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/examiner.rb:71:in `new'
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/examiner.rb:71:in `run'
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/examiner.rb:35:in `initialize'
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/cli/reek_command.rb:14:in `new'
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/cli/reek_command.rb:14:in `block in execute'
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/cli/reek_command.rb:13:in `each'
	from /home/matijs/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/reek-3.2.1/lib/reek/cli/reek_command.rb:13:in `execute'
[...]

PR #686, opened 2015-09-04
When detecting single smell, allow configuration for other smells

mvz:
#682.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* (mvz) Do not fail on explicit configuration for unselected smells when
+  selecting smells on the command line
+
 ## 3.3.1 (2015-09-03)
 
 * (troessner) Fix file list as argument in rake task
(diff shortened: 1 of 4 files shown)


chastell:
Ok, so we currently raise when there is config for a _valid_ smell that’s not in the currently enabled ones; this is clearly wrong. But this seems to fix it by removing a feature that’s actually useful: having a config for an _invalid_, non-existing smell (e.g., because someone made a typo in the smell’s name).

I think this feature is useful; can we keep raising when encountering configuration for smells that don’t exist at all?

mvz:
> can we keep raising when encountering configuration for smells that don’t exist at all?

We already raise earlier in the process when reading the configuration file from disk.

mvz:
Due to the changes to master this is basically just an update of the cucumber scenarios now.

troessner:
>> Due to the changes to master this is basically just an update of the cucumber scenarios now.

Then I believe we should another branch- and commit name for this, otherwise the history is going to be very confusing ;)

mvz:
Agreed.

PR #690, opened 2015-09-06
Mark public API

mvz:
The implementation is still a little provisory in order to keep our options open, but here is a PR for marking our guaranteed API as such.

You can now mark modules, classes and methods with the tag `@public`. In YARD's output, any that are not marked this way will show up as private with the usual warning. Other output options are marking the public items explicitely as well, or only marking the public items and leaving the private ones unmarked. I think the current choice makes for the most pleasent reading for any users of our API.

I have chosen not to use `@api public` as the marker, because the `@api` tag is transitive, so marking `Reek` as public would make everything below it public as well. This is of course not what we want.

Once we've finalized the output behavior I will clean up the implementation accordingly.

### DIFF
--- a/docs/templates/default/docstring/html/public_api_marker.erb
+++ b/docs/templates/default/docstring/html/public_api_marker.erb
@@ -1,4 +1,3 @@
-<div class="note notetag">
-  <strong>Note:</strong>
+<div class="note public">
   <div class="inline"><p>This <%= object.type %> is part of Reek's public API.</p></div>
 </div>
--- a/docs/templates/default/docstring/html/public_api_marker.erb
+++ b/docs/templates/default/docstring/html/public_api_marker.erb
@@ -1,0 +1,6 @@
+<% if object.tags(:api).any? { |tag| tag.text == 'public' } %>
+  <div class="note notetag">
+    <strong>Note:</strong>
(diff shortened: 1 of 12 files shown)


troessner:
Very nice!
LGTM!
@chastell wdyt?

chastell:
Oh yes, please! One question:

> marking `Reek` as public would make everything below it public as well

– does this mean that everything below, say, `Reek::Report` is public? (I’m not saying this is a bad thing, just something to remember in case we want to factor out a private class somewhere inside a `@public` module/class.)

LGTM otherwise, so feel free to merge at will. :)

mvz:
@troessner that was fast, maybe you missed this:

> Once we've finalized the output behavior I will clean up the implementation accordingly.

:smile:

mvz:
> > marking Reek as public would make everything below it public as well
> – does this mean that everything below, say, Reek::Report is public?

Using `@api public` as a marker, yes. However, I created a custom tag `@public` that has the behavior I want, so now each public item has to be marked explicitely, which gives a nicer and less surprising result.

chastell:
Ah, perfect, thanks!

troessner:
>> @troessner that was fast, maybe you missed this:

Whoopsie, I did read it, then I read the changes and then my finger slipped :laughing: 
Then let's do that in a follow-up PR, sorry for that.

Issue #687, opened 2015-09-04
Alert on "helper" or "util" in class or module names


troessner:
.....because I'm sick and tired of seeing things like lib/helpers or lib/utils.

For me this means: "I have no idea where to put it so i'll just put it somewhere."

For me that's a bad smell.

WDYT?

andyw8:
"Manager" too?

troessner:
>> "Manager" too?

Good idea!

HParker:
I hear it is considered a bad practice to have a pattern name as a class name. I know I am annoyed by a number of our models that go something like `PostDto` or `PostBuilder`.

chastell:
I think helper + util are a bit different than manager, builder, etc. – the first two are ‘I don’t have a clue where else to put this stuff’ neutral groupings, while manager, builder (presenter? decorator?) are pattern names used in class names. I’d split these two general cases into separate smells.

mvz:
I'm with @chastell regarding builder and manager. Those do communicate function.

I think we can group this problem under UncommunicativeModuleName, detect Util and Helper by default, and allow people to add more forbidden name parts if necessary.

troessner:
On fifteenth thought I'm not so sure anymore that this makes sense. Seems like there are quite a lot valid exceptions. Closing this one.

PR #691, opened 2015-09-07
Refactor UncommunicativeMethodName.

troessner:
Sooooo. On my quest to make reek smell-free I stumbled across UncommunicativeMethodName and decided to refactor it. 

I believe the implementation is now significantly clearer, simpler and more expressive since I introduced regexes for the "accept" pattern as well, which means #687 will be trivial to implement.

### DIFF
--- a/lib/reek/smells/uncommunicative_module_name.rb
+++ b/lib/reek/smells/uncommunicative_module_name.rb
@@ -21,13 +21,13 @@ module Reek
       # The name of the config field that lists the regexps of
       # smelly names to be reported.
       REJECT_KEY = 'reject'
-      DEFAULT_REJECT_SET = [/^.$/, /[0-9]$/]
+      DEFAULT_REJECT_PATTERNS = [/^.$/, /[0-9]$/]
 
       # The name of the config field that lists the specific names that are
       # to be treated as exceptions; these names will not be reported as
       # uncommunicative.
       ACCEPT_KEY = 'accept'
-      DEFAULT_ACCEPT_SET = ['Inline::C']
+      DEFAULT_ACCEPT_PATTERNS = [/Inline::C/]
 
       def self.smell_category
         'UncommunicativeName'
@@ -35,12 +35,12 @@ module Reek
 
       def self.default_config
         super.merge(
-          REJECT_KEY => DEFAULT_REJECT_SET,
-          ACCEPT_KEY => DEFAULT_ACCEPT_SET
+          REJECT_KEY => DEFAULT_REJECT_PATTERNS,
+          ACCEPT_KEY => DEFAULT_ACCEPT_PATTERNS
         )
       end
 
-      def self.contexts # :nodoc:
+      def self.contexts
         [:module, :class]
       end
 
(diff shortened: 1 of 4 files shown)


lib/reek/smells/uncommunicative_module_name.rb
```diff
@@ -21,13 +21,13 @@ class UncommunicativeModuleName < SmellDetector
...
       REJECT_KEY = 'reject'
-      DEFAULT_REJECT_SET = [/^.$/, /[0-9]$/]
+      DEFAULT_REJECT_PATTERNS = [/^.$/, /[0-9]$/]
 
       # The name of the config field that lists the specific names that are
       # to be treated as exceptions; these names will not be reported as
       # uncommunicative.
       ACCEPT_KEY = 'accept'
-      DEFAULT_ACCEPT_SET = ['Inline::C']
+      DEFAULT_ACCEPT_PATTERNS = [/Inline::C/]
```
chastell: Hmm, do we really want to push this default onto everyone just for the sake of our single spec example?

lib/reek/smells/uncommunicative_module_name.rb
```diff
@@ -49,21 +49,35 @@ def self.contexts # :nodoc:
...
           lines: [exp.line],
-          message: "has the name '#{name}'",
-          parameters: { name: name })]
+          message: "has the name '#{module_name}'",
+          parameters: { name: module_name })]
+      end
+
+      private
+
+      def acceptable_name?(context:, module_name:, fully_qualified_name:)
```
chastell: Ruby 2.0 doesn’t have required kwargs, so until `reek` 4 it should be

```ruby
def acceptable_name?(context: raise, module_name: raise, fully_qualified_name: raise)
```

troessner: Yep, thats what I wanted to avoid, having those raises everywhere. 

But you're right, we introduced this pattern and we should see it through, will change it.

troessner: Not to mention that 2.0 fails on travis due to the syntax ;)

lib/reek/smells/uncommunicative_module_name.rb
troessner: Hmm, since it was there already I assumed the initial author had put some thought into it ;)
@mvz maybe you know more about this?

lib/reek/smells/uncommunicative_module_name.rb
chastell: Yeah, let’s just add a `FIXME` [like here](https://github.com/troessner/reek/blob/75a41a8792dabca82170a38aa3f2e305510c2e94/lib/reek/smells/smell_warning.rb#L14-L16).

troessner:
>> Hmm, do we really want to push this default onto everyone just for the sake of our single spec example?

I just went ahead and removed it.

But since I managed to typo commit name AND branch (uncommunicative-method-name when I am actually refactoring uncommunicative-module-name -> very grasdackelig) I'll close this one and open up a new.

-> Closed.

PR #693, opened 2015-09-08
Refactor UncommunicativeMethodName.

troessner:
Follow-up to #691 

Adressed all comments - please re-review.

PR #695, opened 2015-09-09
Fix or suppress all smells.

troessner:
Ladies and gentleman, this pull request does 3 things:

1.) Fixes all low-hanging fruits when it comes to smells in our codebase (e.g. DuplicateMethodCall, UncommunicativeParameterName, etc)
2.) Suppresses all remaining smell warnings via comments (not via configuration file. That's how I started but @chastell convinced me that it's better to use comments) so we can finally get to ....
3.) CI integration with `reek` for `reeks` own codebase!

And then we can address our own smell warnings step by step. At the very least the UtilityFunction and FeatureEnvy smells will definitely require some design changes.

What's left?

* [3] is not working yet, I'll fix this in this very PR, but I'll add it as a second commit so you can review already.
* #693 needs to merged first to make reek completely happy in this PR as well

### DIFF
--- a/lib/reek/cli/options.rb
+++ b/lib/reek/cli/options.rb
@@ -32,6 +32,7 @@ module Reek
 
       private_attr_reader :argv, :options, :parser
 
+      # :reek:UtilityFunction
       def color_support?
         $stdout.tty?
       end
@@ -71,6 +72,7 @@ module Reek
         end
       end
 
+      # :reek:TooManyStatements: { max_statements: 6 }
       def set_configuration_options
         parser.separator 'Configuration:'
         parser.on('-c', '--config FILE', 'Read configuration options from FILE') do |file|
@@ -127,6 +129,7 @@ module Reek
         end
       end
 
+      # :reek:TooManyStatements: { max_statements: 7 }
       def set_utility_options
         parser.separator "\nUtility options:"
         parser.on_tail('-h', '--help', 'Show this message') do
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -14,6 +14,8 @@ module Reek
   # Applies all available smell detectors to a source.
   #
   # @public
+  #
+  # :reek:TooManyInstanceVariables: { max_instance_variables: 6 }
   class Examiner
     #
     # Creates an Examiner which scans the given +source+ for code smells.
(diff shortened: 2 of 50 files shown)


lib/reek/examiner.rb
```diff
@@ -35,7 +37,7 @@ def initialize(source,
       @source        = Source::SourceCode.from(source)
       @configuration = configuration
       @collector     = CLI::WarningCollector.new
-      @smell_types   = eligible_smell_types(filter_by_smells)
+      @smell_types   = Smells::SmellRepository.eligible_smell_types(filter_by_smells)
```
mvz: Nice!

mvz:
Nice! :+1:

lib/reek/cli/options.rb
```diff
@@ -71,6 +72,7 @@ def set_alternative_formatter_options
         end
       end
 
+      # :reek:TooManyStatements: : { max_statements: 6 }
```
chastell: I know it’s super nitpicky, but can we do away with the double colons? They creep me out. :)

chastell:
very nice

troessner:
Alrightie!

>> [3] is not working yet, I'll fix this in this very PR, but I'll add it as a second commit so you can review already.

Done, please see the second commit.

>> #693 needs to merged first to make reek completely happy in this PR as well

Rebased!

Merge? ;)

lib/reek/cli/options.rb
troessner: Not at all nitpicky, fixed here and below (there was another offender).

chastell:
There are three more cases of `:reek:TooManyStatements: : { max_statements:` – any chance to clean up those eyesores as well? :)

troessner:
@chastell absolutely!
But let's rather do this in separate PRs please in order to keep this PR small and contained ;)

chastell:
I meant ‘three more cases of the double colon’.

troessner:
>> I meant ‘three more cases of the double colon’.

Ah, of course ;)

Fixed - I also quickly grep'ed for this pattern so there shouldn't be more occurences in this PR.

troessner:
>> For final nitpicking, this should be part of the other commit

Good point, fixed!

chastell:
A @troessner PR with more than one commit? _Someone_ is growing s̶e̶n̶s̶ soft… Merging!

PR #697, opened 2015-09-09
Fix smells in source locator

martinciu:
I started with removing smells from Source::SourceLocator and did some more things:

- extracted SourcePath from SourceLocator
- made SourceLocator and SourcePath enumerables
- pass SourcePath instead of Pathname around code


lib/reek/source/source_code.rb
```diff
@@ -4,6 +4,7 @@
 end
 require_relative '../tree_dresser'
 require_relative '../ast/node'
+require_relative '../../../lib/reek/source/source_path'
```
troessner: You can just do 

```Ruby
require_relative 'source_path'
```

;)

lib/reek/cli/input.rb
```diff
@@ -31,11 +31,11 @@ def no_source_files_given?
       end
 
       def working_directory_as_source
-        Source::SourceLocator.new(['.']).sources
+        Source::SourceLocator.new(['.'])
```
troessner: Nice!

lib/reek/source/source_locator.rb
```diff
@@ -1,5 +1,5 @@
-require 'private_attr/everywhere'
-require 'find'
+require_relative '../source/source_path'
```
troessner: Same here, just

```Ruby
require_relative 'source_path'
```

should work.

lib/reek/source/source_locator.rb
```diff
@@ -8,64 +8,27 @@ module Source
...
+      def_delegator :sources, :each
       # Initialize with the paths we want to search.
       #
       # paths - a list of paths as Strings
       def initialize(paths, configuration: Configuration::AppConfiguration.default)
-        @paths = paths.flat_map do |string|
-          path = Pathname.new(string)
-          current_directory?(path) ? path.entries : path
+        @paths = paths.map do |string|
+          SourcePath.new(Pathname.new(string), configuration: configuration)
```
troessner: How about moving the 

```Ruby
Pathname.new(string)
```

into SourcePath's initializer to hide away this complexity?

lib/reek/source/source_locator.rb
```diff
@@ -8,64 +8,27 @@ module Source
     # Finds Ruby source files in a filesystem.
     #
     class SourceLocator
+      include Enumerable
+      extend Forwardable
+
+      def_delegator :sources, :each
       # Initialize with the paths we want to search.
       #
       # paths - a list of paths as Strings
```
troessner: Since we're using yarddoc this should be 

>> @param paths [Array<String>]

lib/reek/source/source_path.rb
```diff
@@ -0,0 +1,89 @@
...
+    class SourcePath
+      extend Forwardable
+      include Enumerable
+
+      def_delegators :pathname, :read, :to_s
+      def_delegator :traverse_children, :each
+
+      # Initialize with the pathname and configuration.
+      #
+      # pathname - a path as Pathname
```
troessner: Same here - yarddoc please.

lib/reek/source/source_path.rb
```diff
@@ -0,0 +1,89 @@
...
+      def initialize(pathname, configuration: Configuration::AppConfiguration.default)
+        @configuration = configuration
+        @pathname = pathname.cleanpath
+        ensure_file
+      end
+
+      # Checks is path is relevant
```
troessner: s/is/if/

lib/reek/source/source_path.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      def_delegators :pathname, :read, :to_s
+      def_delegator :traverse_children, :each
+
+      # Initialize with the pathname and configuration.
+      #
+      # pathname - a path as Pathname
+      def initialize(pathname, configuration: Configuration::AppConfiguration.default)
+        @configuration = configuration
+        @pathname = pathname.cleanpath
```
troessner: Can we indent that to the right most token please?

lib/reek/source/source_path.rb
```diff
@@ -0,0 +1,89 @@
...
+module Reek
+  module Source
+    #
+    # Source file path in a filesystem.
+    #
+    class SourcePath
+      extend Forwardable
+      include Enumerable
+
+      def_delegators :pathname, :read, :to_s
```
troessner: Ah, very nice and clean.

lib/reek/source/source_path.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      private
+
+      private_attr_reader :configuration, :pathname
+
+      # Traverses all paths under current path, finds
+      # all relevant Ruby files and returns them as a list.
+      #
+      # @return Enumerator - if no block given
+      def traverse_children
```
troessner: Looking awesome.

lib/reek/source/source_locator.rb
```diff
@@ -8,64 +8,27 @@ module Source
     # Finds Ruby source files in a filesystem.
     #
     class SourceLocator
+      include Enumerable
+      extend Forwardable
+
+      def_delegator :sources, :each
```
mvz: Where is `each` used?

lib/reek/source/source_path.rb
```diff
@@ -0,0 +1,89 @@
...
+        return enum_for(:traverse_children) unless block_given?
+        pathname.find do |pathname|
+          if (path = SourcePath.new(pathname, configuration: configuration)).relevant?
```
mvz: It seems SourcePath now can be two things:
* A path that may have children to examine.
* A single ruby file that may be examined.
I find this a little confusing.

lib/reek/source/source_path.rb
```diff
@@ -0,0 +1,89 @@
...
+            yield path
```
mvz: I think we should yield `pathname` here in order to not break the API for `Examiner`.

lib/reek/source/source_path.rb
```diff
@@ -0,0 +1,89 @@
...
+          elsif path.ignored?
```
mvz: I would check `ingnored?` first, since it's part of `relevant?`.

lib/reek/source/source_locator.rb
troessner: It's used implicitly down here via to_a: https://github.com/troessner/reek/pull/697/files#diff-7b9b794d247d6e79de6ee9446ea9a3feR31

Please also see: http://stackoverflow.com/questions/7384862/is-there-ruby-documentation-that-can-verify-that-enumerable-to-a-calls-each-inte

lib/reek/source/source_path.rb
```diff
@@ -0,0 +1,89 @@
...
+      def initialize(pathname, configuration: Configuration::AppConfiguration.default)
+        @configuration = configuration
+        @pathname = pathname.cleanpath
+        ensure_file
+      end
+
+      # Checks is path is relevant
+      #
+      # @return Boolean
+      def relevant?
```
mvz: Can you give this a more specific name? It is not immediately clear why any path that is not ignored would not be relevant.

lib/reek/source/source_locator.rb
troessner: And this is my biggest problem with proposed changes. A lot of implicitness here and it's partially obscure.
I can see why it makes sense that a SourcePath is enumerable, but making SourceLocator enumerable seems over-engineered.
I'd rather drop that.

lib/reek/source/source_code.rb
```diff
@@ -39,10 +40,10 @@ def initialize(code, origin, parser: Parser::Ruby22)
       # :reek:DuplicateMethodCall: { max_calls: 2 }
       def self.from(source)
         case source
-        when File     then new(source.read, source.path)
-        when IO       then new(source.readlines.join, IO_IDENTIFIER)
-        when Pathname then new(source.read, source.to_s)
-        when String   then new(source, STRING_IDENTIFIER)
+        when File       then new(source.read, source.path)
+        when IO         then new(source.readlines.join, IO_IDENTIFIER)
+        when SourcePath then new(source.read, source.to_s)
```
mvz: Examiner should still take a Pathname. Otherwise, we break the API. Also, I don't want to force people to have to wrap their Pathname in a SourcePath.

troessner:
Can you also please rebase against master?

lib/reek/source/source_locator.rb
mvz: No, I don't think that's where each is used. That's the `to_a` for SourcePath, not SourceLocator.

lib/reek/source/source_locator.rb
```diff
@@ -8,64 +8,27 @@ module Source
...
-        path_excluded?(path) || hidden_directory?(path)
-      end
-
-      def ruby_file?(path)
-        path.extname == '.rb'
-      end
-
-      def current_directory?(path)
-        [Pathname.new('.'), Pathname.new('./')].include?(path)
+        @paths.flat_map(&:to_a)
```
mvz: I find this line very confusing.

lib/reek/source/source_path.rb
troessner: :+1:

lib/reek/source/source_locator.rb
troessner: :+1:

lib/reek/source/source_locator.rb
troessner: >> No, I don't think that's where each is used. That's the to_a for SourcePath, not SourceLocator.

You're right, my bad. Talk about confusing ;)

troessner:
I do like the introduction of SourcePath but I feel that making all-the-things enumerable is more confusing than helpful. I'd rather drop it, especially since there is no real use case for it.

mvz:
Looking at the set of smells in SourceLocator, I think the better option is to move only the UtilityFunctions to a separate class. `SourceLocator#source_paths` should stay in SourceLocator and be refactored internally there. That gives a clearer separation of function between SourceLocator and SourcePath.

lib/reek/source/source_path.rb
martinciu: yep, it's confusing because it is using Pathname#find to traverse all children. SourcePath doesn't represent all possible paths, but only these provided from command line

lib/reek/source/source_path.rb
martinciu: Examiner is modified as well and accepts SourcePath

lib/reek/source/source_code.rb
martinciu: Why can't we change Examiner initializer attribute? Does anyone use Examiner outside of the reek? I think that sticking to Pathname and passing it everywhere is some kind of Primitive Obsession smell.

martinciu:
well Enumerable might be a little confusing but with it you don't have to call `.sources` or `.all` or anything like that on SourceLocator. You can just call `.each` on it. Also passing arrays IMO might be considered as primitive obsession smell. If you use your own "collection object" you have more control over it.

lib/reek/source/source_code.rb
mvz: > Does anyone use Examiner outside of the reek?

It's part of our public API, and yes, I know projects that use it directly (e.g., pronto-reek).

troessner:
I re-reviewed your pull request. Overall I'm afraid that's a "no" from my side.

The problem with this PR is that it adds a lot of complexity for something that:

- was pretty straight-forward before
- was not really a problem in terms of maintainability - In fact it was pretty maintainable already
- has a fixed set of requirements. Meaning that we only have a limited number of ways that we deal with "sources" and those ways won't change much - if at all - in the future since we already cover the basic ways how to do input. 

Your PR does have small classes with high cohesion - something that I always enjoy - but this doesn't change the fact that something that was rather simple conceptually now has an extremely sophisticated set up that feels over the top given the problem domain.

Sorry to tell you this, but I believe in this case it would have been better if you discussed your plan with us before implementing it.

mvz:
I'm afraid I have to agree with @troessner on this.

troessner:
Closing this one as explained above. @martinciu if you'd like to submit another PR -> awesome and highly welcome! But let's discuss it first ;)

Issue #681, opened 2015-09-01
UtilityFunction should not complain method that works inside an interator


isaiah:
The `UtilityFunction` smell could be misleading, e.g.

```ruby
# Just for Fun
class Fun
  attr_accessor :arr

  def bar
    arr.each do |a_variable|
      # This is an intermediate state, should not be communicated as
      # an instance variable, as reek suggests
      b_variable = process(a_variable)
      puts b_variable
    end
  end

  private

  def process(x_variable)
    x_variable + 1
  end
end
```

```shell
reek fun.rb
=> fun.rb -- 1 warning:
=>   [16]:Fun#process doesn't depend on instance state (UtilityFunction)
```

It is useful most of the time, but for method that process an intermediate state, create a instance variable is not the right way.

chastell:
I’d argue that `Fun#process` _is_ a UtilityFunction, but I think we should consider accepting _private_ UtilityFunctions; they’re usually added for readability.

isaiah:
It is, but instance state doesn't apply here. Actually it's huge NO.

troessner:
I'm open to changing this because I agree that the current set ip of UtilityFunction is too strict.
How about making it part of this smell's configuration if we alert on private UtilityFunction and disable it by default?
That way stricter settings are still possible.

mvz:
> ... for method that process an intermediate state, create a instance variable is not the right way.

I think here there is some confusion in the description reek uses: Reek is *not* trying to suggest changing the method to use an instance variable. In fact, the smell description doesn't try to suggest anything, but one way to resolve this is to move the implementation of process to `a_variable`'s class.

> How about making it part of this smell's configuration if we alert on private UtilityFunction and disable it by default?

I'm not too sure: I think UtilityFuntions are in fact often private and not flagging them means we miss out on opportunities to move methods to classes where they belong more naturally.

andyw8:
@mvz makes a good point. I'd like to see some real-world examples of where this warning would be considered invalid.

troessner:
@andyw8 what @chastell said above:

>> I’d argue that Fun#process is a UtilityFunction, but I think we should consider accepting private UtilityFunctions; they’re usually added for readability.

is something I see quite often unfortunately.

andyw8:
Well let's say that `process` was a more complex method, and `x_variable` was a fully-fledged object rather than just an count:

```ruby
  def process(x_variable)
    if x_variable.code == :foo
      "foo"
[...]

PR #698, opened 2015-09-11
Make UtilityFunction configurable for non-public methods.

troessner:
Related to #681 

CC'ing @isaiah @nTraum @lipanski @hwo411 @RaVbaker @tomash

After this is merged we should release 3.4. as well.

### DIFF
--- a/lib/reek/context/code_context.rb
+++ b/lib/reek/context/code_context.rb
@@ -9,7 +9,7 @@ module Reek
     # code element. CodeContexts form a tree in the same way the code does,
     # with each context holding a reference to a unique outer context.
     #
-    # :reek:TooManyMethods: { max_methods: 18 }
+    # :reek:TooManyMethods: { max_methods: 19 }
     # :reek:TooManyInstanceVariables: { max_instance_variables: 8 }
     class CodeContext
       attr_reader :exp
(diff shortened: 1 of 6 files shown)


lib/reek/smells/utility_function.rb
```diff
@@ -53,11 +56,13 @@ def contexts # :nodoc:
       # @return [Array<SmellWarning>]
       #
       # :reek:FeatureEnvy
+      # :reek:TooManyStatements: { max_statements: 6 }
```
troessner: Kind of sucks, but refactoring UtilityFunction is out of the scope for this PR and I'd rather address this separately.

lib/reek/context/code_context.rb
```diff
@@ -9,7 +9,7 @@ module Context
     # code element. CodeContexts form a tree in the same way the code does,
     # with each context holding a reference to a unique outer context.
     #
-    # :reek:TooManyMethods: { max_methods: 18 }
+    # :reek:TooManyMethods: { max_methods: 19 }
```
troessner: Kind of sucks, but refactoring CodeContext is out of the scope for this PR and I'd rather address this separately.

lib/reek/smells/utility_function.rb
troessner: Addendum: After thinking about it I came to the conclusion that the number of statements is **not** a smell here but indeed justified.

I'd leave it exactly the way it is since it is super readable.

PR #700, opened 2015-09-13
Fix smell_details handling in `reek_of` matcher.

### LINKED ISSUE #599: Fix reek_of matcher messages
The failure messages for `reek_of` are too general. For example, if the match fails due to the details being wrong, it just says it didn't smell.

troessner:
Fixes #599 
Actually this feature (properly checking and reacting to missing smell_details) was never implemented ;)

### DIFF
--- a/lib/reek/spec/should_reek_of.rb
+++ b/lib/reek/spec/should_reek_of.rb
@@ -7,6 +7,12 @@ module Reek
     # code smell.
     #
     class ShouldReekOf
+      attr_reader :failure_message, :failure_message_when_negated
+
+      private_attr_reader :configuration, :smell_category, :smell_details
+      private_attr_writer :failure_message, :failure_message_when_negated
+      private_attr_accessor :examiner
+
       def initialize(smell_category,
                      smell_details = {},
                      configuration = Configuration::AppConfiguration.default)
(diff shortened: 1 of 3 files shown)


spec/reek/spec/should_reek_of_spec.rb
```diff
@@ -3,61 +3,113 @@
 require_relative '../../../lib/reek/spec'
 
 RSpec.describe Reek::Spec::ShouldReekOf do
-  context 'rdoc demo example' do
-    let(:ruby) { 'def double_thing() @other.thing.foo + @other.thing.foo end' }
+  describe 'smell type check' do
```
troessner: I refactored the whole file so the diff is a little messy I'm afraid. Probably better if you check the whole file when reviewing, not the diff.

lib/reek/spec/should_reek_of.rb
```diff
@@ -17,22 +23,42 @@ def initialize(smell_category,
 
       def matches?(actual)
         self.examiner = Examiner.new(actual, configuration: configuration)
-        self.all_smells = examiner.smells
-        all_smells.any? { |warning| warning.matches?(smell_category, smell_details) }
+
+        set_failure_messages_for_smell_type && return unless matching_smell_types?
+        set_failure_messages_for_smell_details && return unless matching_smell_details?
+
+        true
```
chastell: After looking at it for some time I finally understood the logic behind this (two sets of messages, we set one for one failure, the other for the other failure and none + return `true` if there are no failures), but I still think it reads sub-optimally.

How about being a bit more verbose, but explicit:

```ruby
def matches?(actual)
  self.examiner = Examiner.new(actual, configuration: configuration)
  set_failure_messages
  matching_smell_types? && matching_smell_details?
end

private

def set_failure_messages
  set_failure_messages_for_smell_details unless matching_smell_details?
  set_failure_messages_for_smell_type unless matching_smell_types?
end
```

chastell:
One nitpick that kept me stumped every time I started reviewing this one and then LGTM. :)

lib/reek/spec/should_reek_of.rb
troessner: But the code you suggest is even more confusing, at least for me:

```Ruby
def set_failure_messages
  set_failure_messages_for_smell_details unless matching_smell_details?
  set_failure_messages_for_smell_type unless matching_smell_types?
end
```

This code gives the impression that both validations are independent from each other. But they are not. The second one is dependent on the first one because it can't stand on it's own.

With my implementation at least this is clear because I say "&& return". With your suggestion, it is (quoting a famous man) "baroque" ;)

troessner: But the code you suggest is even more confusing, at least for me:

```Ruby
def set_failure_messages
  set_failure_messages_for_smell_details unless matching_smell_details?
  set_failure_messages_for_smell_type unless matching_smell_types?
end
```

This code gives the impression that both validations are independent from each other. But they are not. The second one is dependent on the first one because it can't stand on it's own.

With my implementation at least this is clear because I say "&& return". With your suggestion, it is (quoting a famous man) "baroque" ;)

chastell: > This code gives the impression that both validations are independent from each other. But they are not. The second one is dependent on the first one because it can't stand on it's own.

Oh ok, this is a very good point. My initial proposition (which I refactored to the above) was the below `case` – maybe this is a good compromise between readability, the right mechanics and clearer `#matches?` return values?

```ruby
def matches?(actual)
  self.examiner = Examiner.new(actual, configuration: configuration)
  case
  when not matching_smell_types?
    set_failure_messages_for_smell_type
    false
  when not matching_smell_details?
    set_failure_messages_for_smell_details
    false
  else
    true
  end
end
```

(This all stems from `#matches?` doing two distinct things – setting the failure messages and returning a proper boolean. The above at least makes it clear what’s going on when.)

chastell: Alternatively split the responsibilities:

```ruby
def matches?(actual)
  self.examiner = Examiner.new(actual, configuration: configuration)
  case
  when not matching_smell_types?
    set_failure_messages_for_smell_type
  when not matching_smell_details?
    set_failure_messages_for_smell_details
  end
  matching_smell_types? && matching_smell_details?
end
```

(Not sure if clearer.)

troessner: But none of your suggestions above deal with:

````This code gives the impression that both validations are independent from each other. But they are not. The matching_smell_details? one is dependent on the matching_smell_types? one because it can't stand on it's own.````

Which is why I still believe that my initial solution is better.

How about a compromise with something like this:

```Ruby
def matches?(actual)
  if matching_smell_types?
    if matching_smell_details?
      return true
    else
      set_failure_messages_for_smell_details
    end
  else
    set_failure_messages_for_smell_types
  end
  false
end
```

WDYT?

chastell: OK, I see how the `case` does not show the strict dependency – but the two `&& return unless` lines didn’t either. :)

The clearest to me would probably be

```ruby
case
when matching_smell_types? and matching_smell_details?
  true
when matching_smell_types? and not matching_smell_details?
  set_failure_messages_for_smell_details
  false
when not matching_smell_types?
  set_failure_messages_for_smell_types
  false
end
```

…but I’m also OK with your `if`-`if`-`else`-`else` proposal.

troessner: >> …but I’m also OK with your if-if-else-else proposal.

Updated!

troessner:
Addressed all comments - merge? ;)

troessner:
Alright guys, I believe this pull request has been really discussed and changed long enough.

Can somebody merge this now please or are there bigger concerns left?

troessner:
Squashed and ready!

mvz:
Oh, GitHub, where did my comments go? :cry:

Issue #612, opened 2015-07-21
Instance variables vs getters / setters in reek


troessner:
I know we had already started a discussion about this but this was in an unrelated issue and I'd like to start clean.
So, what can we agree on?
I'd say we go with style suggested here: https://github.com/jswanner/private_attr
Meaning: No more instance variables at all.
That's a clear and easy rule to follow.:)

chastell:
It’s a tad u̶g̶l̶y̶ baroque and I have a slight preference for just using `@foo = foo` in constructors (and explicit setters everywhere else if mutation can’t be avoided), but I’m ok with it. Can’t question the simplicity of ‘no `@` at all’. ;)

mvz:
I'm afraid such a simple rule will kill my favourite pattern: memoization:

```ruby
def foo
  @foo ||= some calculation, for_foo
end
```

We use it about 33 times in reek. Creating differently named setters for each seems crazy to me. We'd get this:

```ruby
attr_accessor :foo_memoizer
def foo
  self.foo_memoizer ||= some calculation, for_foo
end
```

chastell:
Ah, right, memoization – so I _do_ use ivar setting outside of the constructor, I just disassociated it so much from state I forgot about this case. :)

troessner:
Ok, then how about "use instance vars in the constructor and for memoization and getters / setters for everything else"?

mvz:
> use instance vars in the constructor and for memoization and getters / setters for everything else

:+1:

chastell:
Oh very much so! I can do this (perfect task for my summer-fried in 30°C brain), but if not then please merge #610 first.

troessner:
Would it make sense to add a "reek styleguide" mentioning this to our contributing guidelines?

mvz:
@troessner good idea.

troessner:
I'll come up with a PR the next days.

chastell:
Oh, I think I have this done, just not pushed, sorry. Lemme check.

chastell:
Sorry, no, still working on this on the `ivar_clean-up` branch – it’s more work than anticipated. I should wrap it up by the end of next week.

troessner:
>> Sorry, no, still working on this on the ivar_clean-up branch – it’s more work than anticipated. I should wrap it up by the end of next week.

Then I'll just don't do anything and let you add that to our styleguide together with everything else in the ivar_clean-up branch, right?

chastell:
Right!

[...]

PR #701, opened 2015-09-13
Add first version of our style guide to contributing.md.

troessner:
Follow-up from our discussion in #612

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,6 +51,9 @@ Once you’re sure your copy of reek works create your own feature branch from o
 git checkout -b your_feature_or_fix_name
 ```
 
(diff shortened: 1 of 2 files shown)


mvz:
Maybe it's better to put this in a separate document and keep the Contributing guide short. I think this makes sense because GitHub links to it whenever someone wants to file a new ticket so it shouldn't be a lot of text.

troessner:
>> Maybe it's better to put this in a separate document and keep the Contributing guide short. I think this makes sense because GitHub links to it whenever someone wants to file a new ticket so it shouldn't be a lot of text.

Good idea - I put it into the docs folder so that we don't have more of this ALL-IN-CAPS-FILES-MADNESS-IN-OUR-ROOT-DIR, please re-review.

PR #703, opened 2015-09-15
Refactor UncommunicativeParameterName.

troessner:


lib/reek/smells/uncommunicative_parameter_name.rb
```diff
@@ -60,10 +56,9 @@ def examine_context(ctx)
 
       private
 
-      def bad_name?(ctx, name)
-        sanitized_name = sanitize name
-        return false if sanitized_name == '*' || accept_names(ctx).include?(sanitized_name)
-        reject_patterns(ctx).any? { |pattern| sanitized_name.match pattern }
```
troessner: I decided to NOT litter our code base with 

      # FIXME: switch to required kwargs when dropping Ruby 2.0 compatibility

We'll just have to grep for "raise" in lib once we drop support for 2.0.

mvz: Makes sense.

lib/reek/smells/uncommunicative_parameter_name.rb
```diff
@@ -60,10 +56,9 @@ def examine_context(ctx)
...
+      def acceptable_name?(context: raise, name: raise)
```
mvz: This looks identical to UncommunicativeMethodName. Perhaps we can merge these after these two refactorings?

lib/reek/smells/uncommunicative_parameter_name.rb
```diff
@@ -45,14 +40,15 @@ def self.default_config
...
       #
-      def examine_context(ctx)
-        context_expression = ctx.exp
-        context_expression.parameter_names.select do |name|
-          bad_name?(ctx, name) && ctx.uses_param?(name)
+      def examine_context(context)
+        expression = context.exp
+        expression.parameter_names.select do |name|
+          sanitized_name = sanitize name
+          !acceptable_name?(context: context, name: sanitized_name) && context.uses_param?(name)
```
mvz: How about breaking out this complex expression into a method with an expressive name?

troessner: Hmmm, I agree that it's a tad complex but at the same time I find this is very readable. "If name is not acceptable and the param is used -> smell".

I could move

```Ruby
!acceptable_name?(context: context, name: sanitized_name) && context.uses_param?(name)
```

to something like:

```Ruby
def uncommunicative_parameter_name(name: sanitized_name, context: context)
  !acceptable_name?(context: context, name: sanitized_name) && context.uses_param?(name)
end
```

What do you think about that?

lib/reek/smells/uncommunicative_parameter_name.rb
troessner: Yes, I thought about that as well since all those "UncommunicativeFoo"-smell detectors are structurally and functionally very similar. Let's do that in a follow-up PR.

lib/reek/smells/uncommunicative_parameter_name.rb
mvz: That's exactly the sort of thing I had in mind. :+1:

lib/reek/smells/uncommunicative_parameter_name.rb
mvz: Agreed.

lib/reek/smells/uncommunicative_parameter_name.rb
troessner: Updated!

troessner:

lib/reek/smells/uncommunicative_parameter_name.rb
chastell: I agree – I was for the `FIXME` when there was one or two of them, but now when there’s more we’ll remember to do a blanket fix (or stumble upon one of them and recall). :)

PR #704, opened 2015-09-15
Refactor UncommunicativeMethodName.

troessner:


spec/reek/smells/uncommunicative_method_name_spec.rb
```diff
@@ -12,7 +12,7 @@
 
   ['help', '+', '-', '/', '*'].each do |method_name|
     it "accepts the method name '#{method_name}'" do
-      src = "def #{method_name}(fred) basics(17) end"
+      src = "def #{method_name}(arg) basics(17) end"
```
troessner: Not really necessary but that "fred" thing really kills me.

lib/reek/smells/uncommunicative_method_name.rb
```diff
@@ -45,21 +39,31 @@ def self.default_config
...
-        reject_names = value(REJECT_KEY, ctx, DEFAULT_REJECT_SET)
-        accept_names = value(ACCEPT_KEY, ctx, DEFAULT_ACCEPT_SET)
-        name = ctx.name.to_s
-        return [] if accept_names.include?(ctx.full_name)
-        var = name.gsub(/^[@\*\&]*/, '')
-        return [] if accept_names.include?(var)
-        return [] unless reject_names.find { |patt| patt =~ var }
+      def examine_context(context)
+        name = context.name.to_s
+        return [] if acceptable_name?(context: context, name: name)
```
mvz: I would put the `name` argument first for clarity (not that it matters to Ruby, of course).

troessner: Good idea, I changed the call here and the signature as well.

PR #705, opened 2015-09-15
Make FeatureEnvy and UtilityFunction messages more helpful.

### LINKED ISSUE #702: Improve warnings for FeatureEnvy and UitilityFunction
Right now:

UtilityFunction: "doesn't depend on instance state"
FeatureEnvy: "refers to #{name} more than self"

I suggest to change this to:

UtilityFunction: "doesn't depend on instance state (maybe move it to another class?)"
FeatureEnvy: "refers to #{name} more than self (maybe move it to another class?)"

I can hear you scream "but the current message are correct!".
Yes, they are. But this is not academic software and our experience has clearly shown that users are very confused by the current messages, e.g. in #681 

And yes, we could keep the existing messages around and add more information to our -U flag or something similar, but no novice is going to read that. 

Let's make it as easy as possible for users.

WDYT?

troessner:
Fixes #702


mvz:
How did I miss that the build was red? I'm on it.

PR #707, opened 2015-09-17
Fix options hell in CLI namespace.

### LINKED ISSUE #605: Option hell in /cli
We should get rid of the option hell in /cli which always trips me up. There's the OptionInterpreter, the OptionParser and that options OpenStruct.
We should unify this and remove confusion.
WDYT?

troessner:
Fixes #605

Basically this removes the bad "Options#options" abstraction completely.
Now we're down to just using "Options" and "OptionInterpreter" which looks appropriate to me.
I also refactored application.rb and beefed up our specs a tad ;)

### DIFF
--- a/lib/reek/cli/application.rb
+++ b/lib/reek/cli/application.rb
@@ -1,5 +1,6 @@
 require_relative 'options'
 require_relative 'reek_command'
+require_relative 'option_interpreter'
 require_relative '../configuration/app_configuration'
 
 module Reek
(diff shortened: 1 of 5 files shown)


lib/reek/cli/options.rb
```diff
@@ -10,28 +8,40 @@ module CLI
     # Parses the command line
     #
     # See {file:docs/Command-Line-Options.md} for details.
+    #
+    # :reek:TooManyInstanceVariables: { max_instance_variables: 6 }
```
troessner: Those 2 smells were kind of there before but hidden since we were using the OpenStruct as "options" attribute.

lib/reek/cli/application.rb
```diff
@@ -48,12 +46,23 @@ def report_smells
...
+      def configure_options(argv)
+        Options.new(argv).parse
+      rescue OptionParser::InvalidOption => error
+        $stderr.puts "Error: #{error}"
+        exit STATUS_ERROR
+      end
+
+      def configure_app_configuration(config_file)
+        Configuration::AppConfiguration.from_path(config_file)
+      rescue Reek::Configuration::ConfigFileException => error
```
troessner: The rescue behaviour is similar to above but I'd keep it separate for clarity.

chastell:
Oh I like this one very much! Thanks!

PR #709, opened 2015-09-18
Bump deps and fix Aruba 0.9 deprecations

chastell:
(That thing where you want to get back to hacking on reek, think ‘I’ll just bring deps to the current upstreams’ and spend an hour fixing Aruba deprecations.)

### DIFF
--- a/features/step_definitions/reek_steps.rb
+++ b/features/step_definitions/reek_steps.rb
@@ -11,69 +11,72 @@ When /^I run rake (\w*) with:$/ do |name, task_def|
 end
 
 Then /^it reports nothing$/ do
-  assert_exact_output '', all_stdout
+  expect(last_command_started).to have_output_on_stdout('')
 end
 
 Then /^there is no output on stdout$/ do
-  assert_exact_output '', all_stdout
+  expect(last_command_started).to have_output_on_stdout('')
 end
 
 Then /^stdout includes "(.*)"$/ do |text|
-  assert_partial_output(text, all_stdout)
+  expect(last_command_started).to have_output_on_stdout(/#{Regexp.escape(text)}/)
 end
 
 Then /^it succeeds$/ do
-  assert_exit_status Reek::CLI::Application::STATUS_SUCCESS
+  success = Reek::CLI::Application::STATUS_SUCCESS
+  expect(last_command_started).to have_exit_status(success)
 end
 
 Then /^the exit status indicates an error$/ do
-  assert_exit_status Reek::CLI::Application::STATUS_ERROR
+  error = Reek::CLI::Application::STATUS_ERROR
+  expect(last_command_started).to have_exit_status(error)
 end
 
 Then /^the exit status indicates smells$/ do
-  assert_exit_status Reek::CLI::Application::STATUS_SMELLS
+  smells = Reek::CLI::Application::STATUS_SMELLS
+  expect(last_command_started).to have_exit_status(smells)
 end
 
 Then /^it reports:$/ do |report|
-  assert_exact_output "#{report}\n", all_stdout
+  expect(last_command_started).to have_output_on_stdout(report.gsub('\n', "\n"))
 end
 
 Then /^it reports this yaml:$/ do |expected_yaml|
   expected_warnings = YAML.load(expected_yaml.chomp)
-  actual_warnings = YAML.load(all_stdout)
+  actual_warnings = YAML.load(last_command_started.stdout)
   expect(actual_warnings).to eq expected_warnings
 end
 
 Then /^it reports this JSON:$/ do |expected_json|
   expected_warnings = JSON.parse(expected_json.chomp)
-  actual_warnings = JSON.parse(all_stdout)
+  actual_warnings = JSON.parse(last_command_started.stdout)
   expect(actual_warnings).to eq expected_warnings
 end
 
 Then /^stderr reports:$/ do |report|
-  assert_exact_output report, all_stderr
+  expect(last_command_started).to have_output_on_stderr(report.chomp)
 end
 
 Then /^it reports no errors$/ do
-  assert_exact_output '', all_stderr
+  expect(last_command_started).to have_output_on_stderr('')
 end
 
 Then /^it reports an error$/ do
-  expect(all_stderr).to_not be_empty
+  expect(last_command_started.stderr).to_not be_empty
 end
 
 Then /^it reports the error ['"](.*)['"]$/ do |string|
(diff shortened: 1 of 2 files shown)


features/step_definitions/reek_steps.rb
```diff
@@ -11,69 +11,72 @@
 end
 
 Then /^it reports nothing$/ do
-  assert_exact_output '', all_stdout
+  expect(last_command_started).to have_output_on_stdout('')
```
mvz: I'm trying to wrap my head around the code in Aruba starting at https://github.com/cucumber/aruba/blob/bfb67969567cf9d74140c1957e58cf65251adda7/lib/aruba/cucumber/command.rb#L134, but I *think* we need something like `an_output_string_being_eq` in here.

mvz: Or, maybe we should do what this line does: https://github.com/cucumber/aruba/blob/bfb67969567cf9d74140c1957e58cf65251adda7/lib/aruba/cucumber/command.rb#L312

chastell: Hm, the above switch was what [the deprecation message said we should use](https://github.com/cucumber/aruba/blob/v0.9.0/lib/aruba/api/deprecated.rb#L560) and seems to work (and read) right to me…

mvz: Yes, it is probably right. In Aruba's case, 'use the source' seems like a bad advice at the moment :smile:.

features/rake_task/rake_task.feature
```diff
@@ -36,6 +39,9 @@ Feature: Reek can be driven through its Task
         [4, 5]:Smelly#m calls @foo.bar 2 times (DuplicateMethodCall)
         [4, 5]:Smelly#m calls puts(@foo.bar) 2 times (DuplicateMethodCall)
         [3]:Smelly#m has the name 'm' (UncommunicativeMethodName)
+
+
+!!! `reek` has found smells - exiting!
```
mvz: What does this do?

chastell: We [do output this on smell finds](https://github.com/troessner/reek/blob/bump_deps/lib/reek/rake/task.rb#L108) and Cucumber examples can’t indent expected output properly if there are empty lines (regardless whether they’re indented or not).

We can try either switching the `it reports` step to not be an exact match or stop displaying this line on failures; I don’t have a better solution (given that we can’t make the scenario look nice).

features/step_definitions/reek_steps.rb
```diff
@@ -11,69 +11,72 @@
...
 end
 
 Then /^there is no output on stdout$/ do
-  assert_exact_output '', all_stdout
+  expect(last_command_started).to have_output_on_stdout('')
 end
 
 Then /^stdout includes "(.*)"$/ do |text|
-  assert_partial_output(text, all_stdout)
+  expect(last_command_started).to have_output(/#{Regexp.escape(text)}/)
```
mvz: You should use `have_output_on_stdout` here.

features/step_definitions/reek_steps.rb
```diff
@@ -11,69 +11,72 @@
...
 
 Then /^the exit status indicates smells$/ do
-  assert_exit_status Reek::CLI::Application::STATUS_SMELLS
+  smells = Reek::CLI::Application::STATUS_SMELLS
+  expect(last_command_started).to have_exit_status(smells)
 end
 
 Then /^it reports:$/ do |report|
-  assert_exact_output "#{report}\n", all_stdout
+  expect(last_command_started).to have_output(report.gsub('\n', "\n"))
```
mvz: You should use `have_output_on_stdout` here. That way, you don't have to add the extra line `!!! reek has found smells - exiting!` to the scenario's.

features/step_definitions/reek_steps.rb
```diff
@@ -11,69 +11,72 @@
...
 end
 
 Then /^it reports an error$/ do
-  expect(all_stderr).to_not be_empty
+  expect(all_commands.map(&:stderr).join("\n")).to_not be_empty
 end
 
 Then /^it reports the error ['"](.*)['"]$/ do |string|
-  assert_partial_output string, all_stderr
+  expect(last_command_started).to have_output(/#{Regexp.escape(string)}/)
```
mvz: -> `on_stderr`.

features/step_definitions/reek_steps.rb
```diff
@@ -11,69 +11,72 @@
...
 end
 
 Then /^it reports no errors$/ do
-  assert_exact_output '', all_stderr
+  expect(last_command_started).to have_output_on_stderr('')
 end
 
 Then /^it reports an error$/ do
-  expect(all_stderr).to_not be_empty
+  expect(all_commands.map(&:stderr).join("\n")).to_not be_empty
```
mvz: Mabe use `last_command_started.stderr`?

chastell:
Aaaah, thanks, @mvz, for the right pointers – all fixed and green.

features/step_definitions/reek_steps.rb
```diff
@@ -11,69 +11,72 @@
...
 
 Then /^it reports:$/ do |report|
-  assert_exact_output "#{report}\n", all_stdout
+  expect(last_command_started).to have_output_on_stdout(report.gsub('\n', "\n"))
 end
 
 Then /^it reports this yaml:$/ do |expected_yaml|
   expected_warnings = YAML.load(expected_yaml.chomp)
-  actual_warnings = YAML.load(all_stdout)
+  actual_warnings = YAML.load(all_commands.map(&:stdout).join("\n"))
```
mvz: Sorry to be so nitpicky, but ...
Why not just use `last_comand_started.stdout` here as well?

features/step_definitions/reek_steps.rb
```diff
@@ -11,69 +11,72 @@
...
   expect(actual_warnings).to eq expected_warnings
 end
 
 Then /^it reports this JSON:$/ do |expected_json|
   expected_warnings = JSON.parse(expected_json.chomp)
-  actual_warnings = JSON.parse(all_stdout)
+  actual_warnings = JSON.parse(all_commands.map(&:stdout).join("\n"))
```
mvz: ... and here?

chastell:
@mvz nitpicky is always good – and the only reason was that the deprecation warning was ‘stop using `all_stdout`, use `all_commands.map(&:stdout).join("\n")` instead’. Fixed, as `last_comand_started.stdout` is clearly better. :)

mvz:
Excellent! :shipit:

PR #711, opened 2015-09-19
LocationFormatters: sort lines in reports

chastell:
I recently noticed that our formatters don’t sort line numbers on output.

This makes the `DefaultLocationFormatter` sort lines on output, `SingleLineLocationFormatter` output the first line and turns all `.format` methods into `module_functions`.

### DIFF
--- a/spec/reek/report/location_formatter_spec.rb
+++ b/spec/reek/report/location_formatter_spec.rb
@@ -1,0 +1,30 @@
+require_relative '../../spec_helper'
+require_relative '../../../lib/reek/report/formatter'
+
+RSpec.describe 'location formatters' do
+  let(:warning) { build(:smell_warning, lines: [11, 9, 250, 100]) }
+
+  describe Reek::Report::BlankLocationFormatter do
+    describe '.format' do
+      it 'returns a blank String regardless of the warning' do
+        expect(described_class.format(warning)).to eq('')
+      end
+    end
+  end
+
+  describe Reek::Report::DefaultLocationFormatter do
+    describe '.format' do
+      it 'returns a prefix with sorted line numbers' do
+        expect(described_class.format(warning)).to eq('[9, 11, 100, 250]:')
+      end
+    end
+  end
+
+  describe Reek::Report::SingleLineLocationFormatter do
+    describe '.format' do
+      it 'returns the first line where the smell was found' do
+        expect(described_class.format(warning)).to eq('dummy_file:9: ')
+      end
+    end
+  end
+end
--- a/lib/reek/report/location_formatter.rb
+++ b/lib/reek/report/location_formatter.rb
@@ -4,7 +4,9 @@ module Reek
     # Formats the location of a warning as an empty string.
     #
     module BlankLocationFormatter
-      def self.format(_warning)
+      module_function
+
+      def format(_warning)
         ''
       end
     end
@@ -13,8 +15,10 @@ module Reek
     # Formats the location of a warning as an array of line numbers.
     #
     module DefaultLocationFormatter
-      def self.format(warning)
-        "#{warning.lines.inspect}:"
+      module_function
+
+      def format(warning)
+        "#{warning.lines.sort.inspect}:"
       end
     end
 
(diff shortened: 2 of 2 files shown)


spec/reek/report/location_formatter_spec.rb
```diff
@@ -0,0 +1,37 @@
+require_relative '../../spec_helper'
+require_relative '../../../lib/reek/report/formatter'
+require_relative '../../../lib/reek/smells/duplicate_method_call'
+require_relative '../../../lib/reek/smells/smell_warning'
```
troessner: Not something specific to this pull request but I believe the "../../../" madness in our specs highly distracting.
How about setting up a spec wide visible constant that points to $ROOT/lib/reek?
I believe this would clean up things quite a bit...

spec/reek/report/location_formatter_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+RSpec.describe 'location formatters' do
+  let(:warning) do
+    Reek::Smells::SmellWarning.new(Reek::Smells::DuplicateMethodCall.new,
+                                   lines:   [11, 9],
+                                   message: 'Foo#bar calls baz.qux 2 times',
+                                   source:  'string')
```
troessner: We already introduced FactoryGirl, how about using a proper factory for this?

spec/reek/report/location_formatter_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      it 'returns a blank String regardless of the warning' do
+        expect(described_class.format(warning)).to eq('')
+      end
+    end
+  end
+
+  describe Reek::Report::DefaultLocationFormatter do
+    describe '.format' do
+      it 'returns a prefix with sorted line numbers' do
+        expect(described_class.format(warning)).to eq('[9, 11]:')
```
troessner: Can we broaden our test cases to more than two line numbers? With just 2 it's a 50:50 chance that is sorted correctly already :laughing:

spec/reek/report/location_formatter_spec.rb
chastell: :+1: from me – either that or a `require_lib_reek` method that would work like this:

```ruby
require_lib_reek 'smells/smell_warning'
```

Using something like a `$LIB_REEK` constant path is a bit more obvious, `require_lib_reek` a bit prettier – make a call and I’ll prepare a PR. :)

spec/reek/report/location_formatter_spec.rb
chastell: Good point!

I never use FactoryGirl in my projects as I believe an elaborate test setup means there’s a missing factory method in the production code – which (a) would be then tested and kept working and (b) would also be helpful in irb sessions – but as we have FactoryGirl already I’ll switch to it later today. :)

spec/reek/report/location_formatter_spec.rb
troessner:
```Ruby
require_lib_reek 'smells/smell_warning'
```

feels a lot better than a constant, let's go for that!

chastell:
@troessner: addressed both comments. Not sure why Travis doesn’t seem to be building this, though. :)

spec/reek/report/location_formatter_spec.rb
```diff
@@ -0,0 +1,30 @@
+require_relative '../../spec_helper'
+require_relative '../../../lib/reek/report/formatter'
+
+RSpec.describe 'location formatters' do
+  let(:warning) { build(:smell_warning, lines: [11, 9, 250, 100]) }
```
troessner: Nice, and by using a factory we reduced the amount of code further -> like!

spec/reek/report/location_formatter_spec.rb
mvz: This is silly. Bundler will add ./lib to the load path, so we can just use `require` in this case.

troessner: >> This is silly. Bundler will add ./lib to the load path, so we can just use require in this case.

See our discussion in #712 ;)

PR #712, opened 2015-09-20
require_lib_reek

chastell:
As per [the recent discussion](https://github.com/troessner/reek/pull/711#discussion_r39927347) this introduces a private, top-level `require_lib_reek` for use in specs to shorten the `require_relative '../../..`… trainwrecks.

### DIFF
--- a/spec/reek/ast/node_spec.rb
+++ b/spec/reek/ast/node_spec.rb
@@ -1,5 +1,5 @@
 require_relative '../../spec_helper'
-require_relative '../../../lib/reek/ast/node'
+require_lib 'reek/ast/node'
 
 RSpec.describe Reek::AST::Node do
   context 'format' do
--- a/spec/reek/ast/object_refs_spec.rb
+++ b/spec/reek/ast/object_refs_spec.rb
@@ -1,5 +1,5 @@
 require_relative '../../spec_helper'
-require_relative '../../../lib/reek/ast/object_refs'
+require_lib 'reek/ast/object_refs'
 
 RSpec.describe Reek::AST::ObjectRefs do
   let(:refs) { Reek::AST::ObjectRefs.new }
--- a/spec/reek/ast/reference_collector_spec.rb
+++ b/spec/reek/ast/reference_collector_spec.rb
@@ -1,5 +1,5 @@
 require_relative '../../spec_helper'
-require_relative '../../../lib/reek/ast/reference_collector'
+require_lib 'reek/ast/reference_collector'
 
 RSpec.describe Reek::AST::ReferenceCollector do
   context 'counting refs to self' do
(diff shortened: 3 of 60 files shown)


andyw8:
I'm curious – since RSpec already appends `/lib` to the load path, why does reek use require_relative everywhere?

chastell:
Hm, we had a similar discussion in https://github.com/troessner/reek/pull/427 – we decided to use `require_relative` everywhere to side-step any chance for using the files from the currently-installed gem (rather than from local `lib`), but I’m open to simplifying things if we’re sure this would always use the local `lib` files (and blow up if they’re missing).

mvz:
I think `require_relative` makes sense inside `lib` itself, but if it becomes annoying in the specs we should just use `require` there.

It will blow up if you use Bundler to isolate yourself from any installed reek gem.

troessner:
>> I think require_relative makes sense inside lib itself, but if it becomes annoying in the specs we should just use require there.

I disagree with that. The point was to use "require_relative" in our code and since specs are also just ......code this would be inconsistent.
Additionally I believe it was a right and good decision to not use "require" where ever possible :)

chastell:
I don’t have a strong opinion either way (although I do have the _slightest_ of preference for `require_relative`, as it’s crystal-clear what it does, rather than ‘this line will require _something_ and _what exactly_ depends on whatever the user and other gems they might be running had the great idea of putting in `$LOAD_PATH`’).

On a slight tangent, I’m not a huge fan of RSpec adding `./lib` to `$LOAD_PATH` in the first place; it bit me more than once in the past (tests passing, code not working; same with RSpec’s implicit requires of some stdlibs, like `ostruct` – ‘look how nicely your code blows up, wish there was a way to replicate this in tests…’). ಠ_ರೃ

troessner:
>> chastell closed this a minute ago
chastell reopened this a minute ago

Nervous with your trigger finger? :laughing:

troessner:
>> I don’t have a strong opinion either way (although I do have the slightest of preferences for require_relative, as it’s crystal-clear what it does, rather than ‘this line will require something and what exactly depends on whatever the user and other gems they might be running had the great idea of putting in $LOAD_PATH’).

I do have a strong opinion about this thankfully. Ruby's "require" is broken. The Ruby core team silently agrees with that and hence invented require_relative. 
Your examples above are a picture perfect reason for me to still use require_relative, especially in our specs ;)

mvz:
Ok, I see your points. `require_relative` is indeed much more precise.

How about making the addition in this PR a little more generic then and creating a `require_lib`, that takes the full path from lib: `require_lib 'reek/examiner'`. I would prefer that tremendously over `require_lib_reek 'examiner'`.

troessner:
>> How about making the addition in this PR a little more generic then and creating a require_lib, that takes the full path from lib: require_lib 'reek/examiner'. I would prefer that tremendously over require_lib_reek 'examiner'.

:+1:

chastell:
> Nervous with your trigger finger? :laughing:

Yup – feeding the kid Haferbrei and clicking does not mix. :)

> `require_lib 'reek/examiner'`

Done and green!

PR #724, opened 2015-09-25
Report unused private methods.

troessner:
Just a stub for the smell detector right now.
I'd like to discuss the specs though.


chastell:
I like the specs, except the ‘does not report methods that are called on `self`’ can be dropped – you can’t call private methods on `self` in Ruby (as you can see by trying to use the code from the spec). :)

troessner:
>> I like the specs, except the ‘does not report methods that are called on self’ can be dropped – you can’t call private methods on self in Ruby (as you can see by trying to use the code from the spec). :)

Thanks for catching this, no idea what I was thinking (or rather: not thinking).

spec/reek/smells/unused_private_method_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    xit 'creates warnings correctly' do
+      source = <<-EOF
+        class Klazz
+          private
+          def foo; end
+        end
+      EOF
+
+      ctx = create_context(source)
```
mvz: Depending on the smell detector, you probably want to create different kinds of context (ModuleContext, MethodContext, ...).

mvz:
Specs are looking good, as far as I'm concerned.

spec/reek/smells/unused_private_method_spec.rb
troessner: Good point, I'll keep that in mind.

lib/reek/ast/node.rb
```diff
@@ -114,8 +115,6 @@ def look_for_types(target_types, ignoring = [], &blk)
 
       private
 
-      private_attr_reader :comments
```
troessner: Unrelated to this PR. Boyscout rule. It drivers me **cracy** when this is not where it belongs, which is, at the beginning of the class.

lib/reek/context/class_context.rb
```diff
@@ -0,0 +1,11 @@
+require_relative 'module_context'
+
+module Reek
+  module Context
+    #
+    # A context wrapper for any class found in a syntax tree.
+    #
+    class ClassContext < ModuleContext
```
troessner: We need an own context for this smell since it should only run on ....classes, not modules. 
Other than that, I believe we should do this anyway and not just alias class processing to module processing.

troessner:
Ready for ......, nope, wait, ready for nothing. Oh my. I really did underestimate the complexity of this feisty little honeybadger.
This PR is more or less done with the core functionality including "making it nice" however there are a couple of crucial things not working:

* TreeWalker spec is broken. All other specs are working including cucumber.
* Deactivating this smell on a method base does not work because I start out with the surrounding ClassContext and then work my way down to the methods. The comment however is associated to the method_context in question.
* Reported location is always the beginning of the class, not the method for the same reasons like above
* "rake test:quality" is failing because of [3] since now all the dynamic dispatches in TreeWalker a false alarms

I'll work on all of those points in the next days.
If anybody is interested for a preliminary review please go ahead.

lib/reek/tree_walker.rb
```diff
@@ -105,17 +111,19 @@ def process_default(exp)
...
-      end
-      if exp.attribute_writer?
+        element.track_visibility(method_name, exp.arg_names)
+      elsif exp.attribute_writer?
         exp.args.each do |arg|
           next unless arg.type == :sym
           new_context(Context::AttributeContext, arg, exp)
         end
+      else
+        new_context(Context::SendContext, exp, method_name)
```
mvz: Looks like these changes should not be part of the PR but came from some conflict resolution?

mvz: This may very well be the cause of the failing TreeWalker specs.

troessner: Not sure I follow - I need to create SendContexts so I can use them here: https://github.com/troessner/reek/pull/724/files#diff-18413dbca05d0a29829ff59ac73d71e6R22

mvz: Oh, I see now. Are you sure we need SendContext for that, and cannot just look for :send nodes using `Reek::Ast::Node#find_nodes`?

troessner: >> Oh, I see now. Are you sure we need SendContext for that, and cannot just look for :send nodes using Reek::Ast::Node#find_nodes?

Technically we could, but I believe we shouldn't. First of all "defined_private_methods" works only on contexts, and it has to work on context because we don't now their visibility otherwise. Since the whole point of "method_calls" at least in the scope of this PR is to be able to do something like [here](https://github.com/troessner/reek/pull/724/files#diff-13f69a71ba371f43053cba8fecf2d1e0R64) this would break consistency and work on different levels of abstraction.

Conceptually I believe we should put more and more logic into our own abstractions. So our contexts in this case.
Or in other words: TreeWalker >>> TreeDresser for me. 
Our adorned AST is a paper thin abstraction that provides you with a bare minimum of features.
Additionally we already started to go down this road heavily by putting the visibility tracking in there.

troessner:
Re: I started looking into 

>> [1] TreeWalker spec is broken. All other specs are working including cucumber.

and realized that **not** my code is the problem, but the specs.
1.) They are one big mess. Just look at the test set up. 
2.) The test are not testing anything that is public.
3.) The statement counting they test doesn't make any sense whatsoever. Maybe it did at one point and then lost it's meaning without anybody noticing. That's exactly the reason why **only** the TreeWalker test fail. If the code introduced in this PR had really broken the TreeWalker dozens of other tests would have failed as well.

Conclusion: I'll kill those tests and introduce new ones for the TreeWalker.

lib/reek/context/class_context.rb
mvz: If this is the case doesn't it make more sense to move `#method_calls` here?

(But perhaps this smell should also check modules?)

lib/reek/tree_walker.rb
mvz: I understand your reasoning, but I believe we should consider Contexts to be 'things that can have a smell', so currently these would be classes, modules, methods, and some class-level method calls like attribute definitions.

In contrast, all other additions to Nodes are in TreeDresser.

We could merge contexts and sexp extensions, but I think it makes sense to have two hierarchies, since the nodes often have intermediate hierarchy levels that we don't want to include in the contexts. This was one reason I moved the visibility tracking into contexts.

What do you think?

lib/reek/context/class_context.rb
troessner: >> If this is the case doesn't it make more sense to move #method_calls here?

I'd leave it here since the method's purpose is not tied to the SmellDetector.

>> (But perhaps this smell should also check modules?)

But you don't know if a private module method is called or not since it might get mixed into a class.....or not. And we don't know that. We *would* know it if Reek would slurp in all the sources and build up a call / class / whatever registry for later usage but with our current "one file == one examiner" paradigm this wouldn't work.

lib/reek/tree_walker.rb
troessner: >> I understand your reasoning, but I believe we should consider Contexts to be 'things that can have a smell'

Why? I can't see any good reasons for that but I can see quite a few compelling ones for having proper contexts. Not for all upfront, but for when we need it. E.g. now the SendContext for this very specific feature.

>> In contrast, all other additions to Nodes are in TreeDresser.

Which doesn't do visibility tracking which we need here. There is no way I could implement this SmellDetector properly **not** using contexts. Unless I'd re-implement it that on a TreeDresser base.

>> We could merge contexts and sexp extensions, but I think it makes sense to have two hierarchies, since the nodes often have intermediate hierarchy levels that we don't want to include in the contexts.

Agreed.

troessner:
After all my latest refactorings:

- #785 
- #783 
- #782 
- #781 
- #778 
- #777 

I restarted looking into this bad boy again.

Thankfully the refactorings made this problem:

>> [1] TreeWalker spec is broken. All other specs are working including cucumber.

 go away :laughing: 

Last problem left:

>>  [2] Deactivating this smell on a method base does not work because I start out with the surrounding ClassContext and then work my way down to the methods. The comment however is associated to the method_context in question.

-> wip.

mvz:
> I start out with the surrounding ClassContext

Have you tried starting with the MethodContext instead? Then, you can go up to the ClassContext easily while still getting the config from the method.

mvz:
> After all my latest refactorings:

Awesome work by the way! :+1:

troessner:
>> Have you tried starting with the MethodContext instead? Then, you can go up to the ClassContext easily while still getting the config from the method.

Yes, I thought about that as well. That would indeed solve the "configure via comment" problem but would lead to other problems.
The biggest one: Since a SmellDetector only gets passed the context he's interested in he has no knowledge outside of this context. Meaning that if I did it like that, I would have:

- traverse up until I hit a class scope
- which is not entirely simple since I can't rely on that `parent` is the class scope because I might have input code that looks like this:

```Ruby
class Holla
  def die
    def wald_fee; end
  end
end
```

or even

```Ruby
class Holla
  def die
    class << self
      def wald_fee; end
    end
  end
end
```

- then I'd have to traverse down and detect all method calls
- And here comes the big one: over and over again for every method context because a SmellDetector called for MethodContext B in Class C has no knowledge about what he has done with MethodContext A before

We might be able to come up with something that allows us to transfer this knowledge across SmellDetector calls but this would be hacky at best and error-prone and confusing as hell.

I believe the way I'm doing it right now is the right way to go.

mvz:
> ```
> class Holla
>  def die
>    class << self
>      def wald_fee; end
>    end
>  end
> end
> ```

I think the Attribute detector also fails in that case :frowning:.

> I believe the way I'm doing it right now is the right way to go. 

You're probably right.

Regarding the configuration comments: Our general rule is that the comments are added to the context that the detector starts from, so for this smell, the context would be the class, and that would be where the comments should go. What do you think of this solution?

troessner:
>> Regarding the configuration comments: Our general rule is that the comments are added to the context that the detector starts from, so for this smell, the context would be the class, and that would be where the comments should go. What do you think of this solution?

But then you can only enable or disable it on a class base. So if you have just one method in question you have to disable it for all. Seems like a no-go for me. 
I'd rather make an exception to the rule here.
WDYT?

mvz:
> But then you can only enable or disable it on a class base.

We would have to add a parameter like allowed_methods or something, similiar to how UncommunicateVariableName is configured.

mvz:
(If you've already implemented a way to get this working with comments on the methods themselves, please ignore my suggestion.)

troessner:
Ladies and Gentlemen, after what feels like **forever** I'd like to declare this little honeybadger ready for review!

@mvz regarding:

>> We would have to add a parameter like allowed_methods or something, similiar to how UncommunicateVariableName is configured.

I believe I solved it in an elegant fashion by using what we have already with the basic smell option `exclude` :)

@mvz and @chastell after this one has been merged I'd bump a minor version as well.

docs/Unused-Private-Method.md
```diff
@@ -0,0 +1,46 @@
...
+  def self.start_the_engine; end
+  def drive; end
+end
+```
+
+`Reek` would emit the following warning:
+
+```
+2 warnings:
+  [3]:Car has the unused private class method `start_the_engine` (UnusedPrivateMethod)
```
chastell: This warning is wrong, as `Car.start_the_engine` is not a private method (`private` does not apply to class methods, only the `private_class_method` does).

docs/Unused-Private-Method.md
```diff
@@ -0,0 +1,46 @@
...
+  [4]:Car has the unused private instance method `drive` (UnusedPrivateMethod)
+```
+
+As you can see above, `Unused Private Method` does differentiate between class
+and instance methods.
+
+Private methods that are called via `dynamic dispatch`
```
chastell: Let’s keep `<code> formatting` for actual code. I don’t think dynamic dispatch needs any quoting, but if you feel like it does just quote it with quotation marks. :)

docs/Unused-Private-Method.md
```diff
@@ -0,0 +1,46 @@
...
+will trigger a false alarm since detecting something like this is far out of
+scope for `Reek`. In this case you can disable this detector via the `exclude`
+configuration option (which is part of the [Basic Smell Options](Basic-Smell-Options.md))
+for instance like this (an example from `Reeks` own codebase):
```
chastell: Reek’s

docs/Unused-Private-Method.md
```diff
@@ -0,0 +1,46 @@
...
+
+```Ruby
+# :reek:UnusedPrivateMethod: { exclude: [ !ruby/regexp /process_/] }
```
chastell: Good example for regex config, kudos! I’d drop the space after the `[`, though, as it looks ugly. Or at least add a matching one before `]`. :)

lib/reek/context/code_context.rb
```diff
@@ -22,8 +22,8 @@ class CodeContext
       delegate %i(name type) => :exp
       delegate %i(visibility visibility= non_public_visibility?) => :visibility_tracker
 
-      attr_reader :exp, :children, :visibility_tracker, :statement_counter
-      private_attr_reader :context, :refs
+      attr_reader :context, :children, :exp, :visibility_tracker, :statement_counter
```
chastell: Trivial, but as long as we’re touching this line anyway: can we order the attrs alphabetically? It’d be easier to scan when checking for, say, `:statement_counter`.

lib/reek/context/module_context.rb
```diff
@@ -7,6 +9,19 @@ module Context
...
+
+      def defined_methods(visibility: :public)
+        select do |context|
+          context_is_some_kind_of_method?(context) &&
+            context.visibility == visibility
+        end
+      end
+
+      def method_calls
+        select { |context| context.is_a?(SendContext) }
```
chastell: `grep(SendContext)`? :)

lib/reek/context_builder.rb
```diff
@@ -14,7 +16,8 @@ module Reek
   # counting. Ideally `ContextBuilder` would only build up the context tree and leave the
   # statement and reference counting to the contexts.
   #
-  # :reek:TooManyMethods: { max_methods: 26 }
+  # :reek:TooManyMethods: { max_methods: 27 }
+  # :reek:UnusedPrivateMethod: { exclude: [ !ruby/regexp /process_/] }
```
chastell: Ditto: let’s either drop the space after `[` or add one before `]`.

lib/reek/smells/unused_private_method.rb
```diff
@@ -0,0 +1,128 @@
...
+  module Smells
+    # Class for storing `hits` which are unused private methods
+    # we found in the given context. `name`, `line` and `scope` are then used to
+    # construct SmellWarnings.
+    class Hit
+      attr_reader :name, :scope, :line
+
+      def initialize(context)
+        @name  = context.name
+        @scope = context.class == Context::SingletonMethodContext ? 'class' : 'instance'
```
chastell: How about making it a bit more elastic and check for potential `Context::SingletonMethodContext` descendants as well?

```ruby
@scope = context.class < Context::SingletonMethodContext ? 'class' : 'instance'
```

spec/reek/smells/unused_private_method_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    it 'reports instance methods' do
+      source = <<-EOF
+        class Car
+          private
+          def start_the_engine; end
+          def drive; end
+        end
+      EOF
+
+      expect(source).to reek_of(:UnusedPrivateMethod, name: :start_the_engine)
```
chastell: This is wrong, `Car.start_the_engine` is not a private method.

spec/reek/smells/unused_private_method_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+      expect(source).to reek_of(:UnusedPrivateMethod, name: :drive)
+    end
+
+    it 'reports class methods' do
+      source = <<-EOF
+        class Car
+          private
```
chastell: This should be `private_class_method`, otherwise `Car.start_the_engine` is actually public.

chastell:
I finally have a bit of time again. :)

Major problem: `private` does not make class methods private; `private_class_method` does.

Question: does this also cover prefix notations?

```ruby
class Car
  private_class_method self.start_the_engine
 end
  private def drive
  end
  def this_one_is_public
  end
end
```

lib/reek/context/module_context.rb
```diff
@@ -7,6 +9,19 @@ module Context
     # A context wrapper for any module found in a syntax tree.
     #
     class ModuleContext < CodeContext
+      include Enumerable
```
mvz: I think this will lead to confusing code elsewhere, and the fact that ModuleContext doesn't define `#each` itself is also confusing. I think we can just do `each.select` or `each.grep` below. I'd include Enumerable in ClassContext if it didn't also define other kinds of iterators.

mvz:
Any problems with how visibility is determined are general problems of the visibility logic I think.

docs/Unused-Private-Method.md
troessner: Good catch here and below. I'll fix this everywhere in the docs and the tests and then address the obvious issue in our visibility tracking.

docs/Unused-Private-Method.md
troessner: Fixed.

docs/Unused-Private-Method.md
troessner: Fixed.

docs/Unused-Private-Method.md
troessner: Added a matching one!

lib/reek/context/code_context.rb
troessner: I'm always up for the boyscout rule, fixed!

lib/reek/context/module_context.rb
troessner: I don't agree wholeheartedly but I can see why this might get confusing down the road. Changed.

lib/reek/context/module_context.rb
troessner: My brain just can't seem to put this into active memory :(

troessner: P.S.: Fixed.

spec/reek/smells/unused_private_method_spec.rb
troessner: Will address this separately, see first comment above.

lib/reek/context/module_context.rb
troessner: Fixed with some inspiration from http://blog.arkency.com/2014/01/ruby-to-enum-for-enumerator/

troessner:
Adressed all comments except the ones about the visibility problems - looking into that now.

troessner:
**Status report**

- I addressed all comments
- I fixed all specs that were using public class methods when they should have been private
- The reason that those specs were actually green was because of a bug in our visibility tracking that falsely labeled them private when they weren't: #792 
- After fixing those specs one is now failing with:

```
  
1) Reek::Smells::UnusedPrivateMethod unused private methods reports unused class methods even with equally named used instance method
     Failure/Error: expect(source).to reek_of :UnusedPrivateMethod,
       Expected string to reek of UnusedPrivateMethod, but it didn't
     # ./spec/reek/smells/unused_private_method_spec.rb:56:in `block (3 levels) in <top (required)>
```

- The reason for this is another bug I uncovered in the course of the pull request which relates to our context_builder: #793 

**So what now?**

Working on this is getting highly frustrating for me. This is like the never ending story in real life.

I suggest the following:

- We adapt this detector to only work on instance methods. We're doing this with other smell_detectors as well (undocumented because....nobody noticed or cared about it) and if we document it properly I don't see a problem in this
- We get this one merged asap after adapting it
- We fix #793 and #792 
- We can then return to this detector and make it work on class methods as well.

What do you guys think?

mvz:
> What do you guys think?

Yes, excellent idea! Private class methods are really a corner case so I have no problem handling them later.

troessner:
Closing this one for a clean slate and opened a subset of this one in #794

PR #725, opened 2015-09-25
Get rid of all deprecated methods in AST::Node.

### LINKED ISSUE #546: Get rid of all deprecated methods in AST::Node
Extracted from #542 
I'll look into this.

troessner:
Fixes #546

I believe "first" and "[]" were misleading, artificial and confusing abstractions that added complexity but no value.

Oh, and they **always** tripped me up.;)


chastell:
Oh yes, thanks!

PR #726, opened 2015-09-26
Ensure reek uses default settings during test runs

### LINKED ISSUE #674: errors on master due to ~/.reek file
So I imagine I've done something stupid, but `bundle exec rake` on master (commit hash: 034e209) gives me 800 examples, 23 failures, 2 pending.

I'm on OSX 10.10.5 
$ ruby -v
ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]

and I'm getting 23 errors, however I just tried removing my .reek file from my home directory and that fixed it.  That seems slightly unfortunate that ~/.reek should cause the tests to fail.  For reference here is my .reek file:

```
$ more ~/.reek 
IrresponsibleModule:
  enabled: false
Attribute:
  enabled: true
```

and the test failures it caused:

https://gist.github.com/tansaku/c71c0dc1e46c4a966454

mvz:
Fix #674.

* Pointing `$HOME` elsewhere won't work, since reek will *first* traverse the ancestor directories.
* To block access to the ancestor directories, I have moved defaults.reek to the repo root.
* This means *only* the scenario testing fetching the config from `$HOME` breaks.
* I've tried to move Aruba's working directory elsewhere, but it won't let me point outside the repo.
* Therefore I've removed the config finding scenario's in favor of specs. This exposed a bug in the finder which I've also fixed.


mvz:
Ready for review.

spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -7,18 +7,19 @@
 
 RSpec.describe Reek::Configuration::ConfigurationFileFinder do
   describe '.find' do
-    it 'returns the path if it’s set' do
+    it 'returns the file in :path if it’s set' do
```
chastell: Hm, why the change? `ConfigurationFileFinder#find` doesn’t return a `File`, it returns a `Pathname`.

mvz: I wanted to make it clearer that it would use the path argument. Maybe reword to `the argument in :path`?

chastell: Ah right, sorry, I’m a bit sick and not thinking clearly – it returns the `Pathname` to the file in the directory passed as the `path` argument. Maybe `it 'returns the configuration file found at the passed path'`? Technically, the `ConfigurationFileFinder` is also fuzzy (at it finds configuration files by returning `Pathnames`, not `Files`…).

chastell:
LGTM (other than the a bit fuzzy spec descriptions, but I don’t have a constructive feedback at the moment, sorry).

spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -42,6 +43,11 @@
       end
     end
 
+    it 'prefers the file in :current over one in :home' do
+      found = described_class.find(current: SAMPLES_PATH, home: SAMPLES_PATH.join('masked_by_dotfile'))
+      expect(found).to eq(SAMPLES_PATH.join('exceptions.reek'))
```
troessner: I'm confused. The example description says:

>> prefers the file in :current over one in :home

But in the "expect" you're checking equality with whatever is in "home", not "current"?

mvz: No, `SAMPLES_PATH.join('exceptions.reek')` is in current. From home we would get `SAMPLES_PATH.join('masked_by_dotfile/.reek')`

troessner: Ah, ok - can we make this clearer by using one-off variables?

troessner:
@mvz sorry, I had not realized that you had addressed my comments already. Maybe you can add a quick "Comments addressed" comment in the future ;)

PR consists of 4 commits but those seem to be fairly independent from each other, so I'm merging this bad boy as it is!

chastell:
> PR consists of 4 commits but those seem to be fairly independent from each other, so I'm merging this bad boy as it is!

:fireworks: :tada: :clap: ;)

PR #728, opened 2015-09-26
Unify config loading mechanisms

### LINKED ISSUE #694: Discrepancies between AppConfiguration.from_path and from_map
When you have a config file like this:

````
---
UncommunicativeModuleName:
  accept:
    - Inline::C
````

You end up with this in your AppConfiguration (which gets loaded via AppConfiguration.from_path):

```Ruby
@default_directive={Reek::Smells::UncommunicativeModuleName=>{"accept"=>["Inline::C"]}}
````

This should be equivalent to when you use AppConfiguration.from_map like this:

```Ruby
      default_directive_for_smell = {
        :default_directive => {
          "UncommunicativeModuleName" => {
            "accept" => ['Inline::C']
          }
        }
      }
      Reek::Configuration::AppConfiguration.from_map(default_directive_for_smell)
```

Unfortunately it isn't, since you end up with 

```Ruby
@default_directive={"UncommunicativeModuleName"=>{"accept"=>["Inline::C"]}}
```

in the latter case.

Both methods should produce the exact same result.

mvz:
Fixes #694.

I have some follow-up ideas for this but I'll make that a separate pull request.

### DIFF
--- a/lib/reek/configuration/app_configuration.rb
+++ b/lib/reek/configuration/app_configuration.rb
@@ -42,10 +42,9 @@ module Reek
       def self.from_map(map = {})
         allocate.tap do |instance|
           instance.instance_eval do
-            self.directory_directives = map.fetch(:directory_directives, {}).
-              extend(DirectoryDirectives)
-            self.default_directive = map.fetch(:default_directive, {}).extend(DefaultDirective)
-            self.excluded_paths = map.fetch(:excluded_paths, []).extend(ExcludedPaths)
+            load_values map.fetch(:directory_directives, {})
+            load_values map.fetch(:default_directive, {})
+            load_values EXCLUDE_PATHS_KEY => map.fetch(:excluded_paths, [])
           end
         end
       end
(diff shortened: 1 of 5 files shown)


lib/reek/configuration/app_configuration.rb
```diff
@@ -42,10 +42,9 @@ def self.from_path(path = nil)
       def self.from_map(map = {})
         allocate.tap do |instance|
           instance.instance_eval do
-            self.directory_directives = map.fetch(:directory_directives, {}).
-              extend(DirectoryDirectives)
-            self.default_directive = map.fetch(:default_directive, {}).extend(DefaultDirective)
```
troessner: Admit it: You never liked the extend / DCI thingie, did you? :laughing:

troessner:
Awesome job! Merged!

lib/reek/configuration/app_configuration.rb
mvz: True, but it's still there on lines 78, 82 and 86 :laughing:

chastell: :laughing: I sense another innocent PR coming, thinly veiled in the whole unify / clean-up agenda.

mvz: :innocent:

PR #729, opened 2015-09-26
Ignore iterators without block arguments for NestedIterators.

### LINKED ISSUE #685: Improve NestedIterators sensitivity
One example that will occur quite often in specs and ruby frameworks are "before / after" hooks:

```Ruby
    def self.require_user_uuid_param
      before do
        validate_params { param :user_uuid, String, required: true, format: UUID_REGEX }
        do_stuff
      end
    end
```
This example would be flagged with NestedIterators which doesn't make any sense.

In the issue where this [originated](https://github.com/troessner/reek/issues/676) we discussed ignoring further  common method names like "setup" and "teardown", which, after giving it some thought, seems like a dangerous road to go down due to its vagueness.

Why not exclude all method calls that do not pass an argument to the block, e.g. like the one above?
If it doesn't pass an argument, there's nothing to iterate over, so it should be ignored.
And yes, I'm aware that in theory every method in ruby that takes a block is called an iterator which is complete nonsense in my mind and I believe we shouldn't adhere to it.

We'd need to pay special attention to Symbol2Proc calls since there is no explicit argument but looking at the different S-expressions for all those cases I believe we can distinguish between them all.

What do you think?

cc @RaVbaker (I stumbled upon the example above in one of your PR's).

troessner:
Fixes #685 - ready for review!

### DIFF
--- a/docs/Nested-Iterators.md
+++ b/docs/Nested-Iterators.md
@@ -33,6 +33,21 @@ test.rb -- 1 warning:
 
 Nested Iterators reports failing methods only once.
 `Object#tap` is ignored by default and thus does not count as iterator.
+Furthermore iterators without block arguments are not counted, e.g.:
+
+
+```Ruby
+def foo
+  before do
+    item.each do |part|
+      puts part
+    end
+  end
+end
+```
+
+would not smell of NestedIterators (given a maximum allowed nesting of 1) since the
+`before` would not be counted (because it doesn't pass any arguments to the block).
 
 ## Configuration
 
--- a/lib/reek/smells/nested_iterators.rb
+++ b/lib/reek/smells/nested_iterators.rb
@@ -44,7 +44,6 @@ module Reek
         else
           []
         end
-        # BUG: no longer reports nesting outside methods (eg. in Optparse)
       end
 
       private
@@ -72,9 +71,10 @@ module Reek
         result
       end
 
+      # :reek:FeatureEnvy
       def ignored_iterator?(exp)
-        name = exp.call.method_name.to_s
-        ignore_iterators.any? { |pattern| /#{pattern}/ =~ name }
+        ignore_iterators.any? { |pattern| /#{pattern}/ =~ exp.call.method_name } ||
+          exp.without_block_arguments?
       end
     end
   end
--- a/spec/reek/smells/nested_iterators_spec.rb
+++ b/spec/reek/smells/nested_iterators_spec.rb
@@ -49,27 +49,9 @@ RSpec.describe Reek::Smells::NestedIterators do
   it 'detects an iterator with an empty block' do
     src = <<-EOS
       def foo
-        bar { baz { } }
-      end
-    EOS
-    expect(src).to reek_of(:NestedIterators)
-  end
-
-  it 'should report nested iterators only once per method' do
-    src = <<-EOS
-      def bad(fred)
-        @fred.each {|item| item.each {|part| @joe.send} }
-        @jim.each {|ting| ting.each {|piece| @hal.send} }
-      end
-    EOS
-    expect(src).to reek_of(:NestedIterators)
-  end
-
-  it 'reports nested iterators only once per method even if levels are different' do
-    src = <<-EOS
-      def bad(fred)
-        @fred.each {|item| item.each {|part| part.foo} }
-        @jim.each {|ting| ting.each {|piece| piece.each {|atom| atom.foo } } }
+        bar do |bar|
+          baz {|baz| }
+        end
       end
     EOS
     expect(src).to reek_of(:NestedIterators)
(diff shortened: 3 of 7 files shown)


spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -55,26 +55,6 @@ def foo
...
-  end
-
-  it 'reports nested iterators only once per method even if levels are different' do
-    src = <<-EOS
-      def bad(fred)
-        @fred.each {|item| item.each {|part| part.foo} }
-        @jim.each {|ting| ting.each {|piece| piece.each {|atom| atom.foo } } }
-      end
-    EOS
-    expect(src).to reek_of(:NestedIterators)
```
troessner: Those specs were completely useless in their current form since

>> expect(src).to reek_of(:NestedIterators)

is so generic that it does not fullfill the promise of the example description.

spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -103,7 +83,7 @@ def bad(fred)
     expect(src).to reek_of(:NestedIterators, count: 3)
   end
 
-  it 'handles the case where super recieves a block' do
+  it 'handles the case where super receives a block' do
```
troessner: Boyscout rule here and below.

spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -127,6 +107,72 @@ def super_call_with_block
...
+      warning = warnings.first
+
+      expect(warning.smell_category).to eq(Reek::Smells::NestedIterators.smell_category)
+      expect(warning.smell_type).to eq(Reek::Smells::NestedIterators.smell_type)
+      expect(warning.parameters[:name]).to eq('foo')
+      expect(warning.lines).to eq([2])
+    end
+
+    it 'should report nested iterators only once per method' do
+      source = <<-EOS
```
troessner: And here are those two specs from above again, this time fullfilling the promise.

lib/reek/smells/nested_iterators.rb
```diff
@@ -44,7 +44,6 @@ def examine_context(ctx)
         else
           []
         end
-        # BUG: no longer reports nesting outside methods (eg. in Optparse)
```
troessner: Boyscout rule. Stuff like that belongs into github issues and not in dangling comments.

lib/reek/smells/nested_iterators.rb
```diff
@@ -72,9 +71,10 @@ def find_iters_for_iter_node(exp, depth)
         result
       end
 
+      # :reek:FeatureEnvy
```
troessner: FeatureEnvy, but as far as I can see this is a good and clean solution, so I'd rather leave it like that. (we also might consider configuring FeatureEnvy to be a little bit more lenient, but that's another story).

spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -49,27 +49,9 @@ def chained
...
-      end
-    EOS
-    expect(src).to reek_of(:NestedIterators)
-  end
-
-  it 'reports nested iterators only once per method even if levels are different' do
-    src = <<-EOS
-      def bad(fred)
-        @fred.each {|item| item.each {|part| part.foo} }
-        @jim.each {|ting| ting.each {|piece| piece.each {|atom| atom.foo } } }
```
troessner: Those specs were useless in their former form. I moved them down and adjusted them.

docs/Nested-Iterators.md
```diff
@@ -33,6 +33,21 @@ test.rb -- 1 warning:
...
+def foo
+  before do
+    item.each do |part|
+      puts part
+    end
+  end
+end
+```
+
+would not smell of NestedIterators (given the default settings) since the
```
mvz: > given the default settings

I don't see any configuration to change this?

spec/spec_helper.rb
```diff
@@ -38,6 +38,21 @@ def sexp(type, *children)
...
+  # @param type [Symbol]
+  #   The type of context given as symbol. Can be :method, :code, :module and so on.
+  #
+  # @param source [String]
+  #   Code to analyze.
+  #
+  # @return [Reek::Context::*] One of the contexts in reek/contexts
+  def create_context(type: raise, source: raise)
+    syntax_tree = Reek::Source::SourceCode.from(source).syntax_tree
+    Reek::Context.const_get("#{type.to_s.capitalize}Context").new(nil, syntax_tree)
```
mvz: I'm suddenly thinking, why not use FactoryGirl for this?

```ruby
create :method_context, source: 'def foo; end'
```

docs/Nested-Iterators.md
troessner: That was misleading - I changed it to 

>> would not smell of NestedIterators (given a maximum allowed nesting of 1)

spec/spec_helper.rb
troessner: Good idea, updated!

troessner:

mvz:
One final thing: what version of reek should this be in? 3.6 or 4.0?

troessner:
>> One final thing: what version of reek should this be in? 3.6 or 4.0?

Do you mean 3.5 or 3.6?
Should be 3.5 in my mind.

mvz:
Well, I meant 3.5 or 4.0, but you answered my question anyway :-).

PR #731, opened 2015-09-28
Prepare 3.5.0 release.

troessner:


chastell:
:shipit:

troessner:
@chastell care to release as well? 

I think the person merging should be the one releasing so we have a clear workflow.
WDYT @chastell @mvz?

chastell:
Released! I agree that the ‘merge + release’ workflow is one less ping-pong.

mvz:
Yes, sounds reasonable.

Issue #589, opened 2015-07-10
writing one's own versions of reek messages


tansaku:
apologies for random question, but using reek in an educational context, is there a way to configure differences to the content of the reek mesages?

For example, say that for education purposes we wanted to be able to modify:

```
[13]:CoinManager#total refers to coin_array more than self (FeatureEnvy) [https://github.com/troessner/reek/blob/master/docs/Feature-Envy.md]
```

to

```
[13]:the method CoinManager#total refers to the coin_array parameter argument more times than it does directly to the CoinManager instance itself.  This is a code smell called "FeatureEnvy" which you can read more about here: https://github.com/troessner/reek/blob/master/docs/Feature-Envy.md
```

would there be a simple way to do this without modifying source code?

mvz:
Reek itself does not have any way to configure the messages. You could, however, have reek output in its JSON or YAML format, which both contain enough information to write any message you like.

That said, if you feel these messages are not clear to some of reek's target audience, perhaps we can improve them so you don't need to change them?

troessner:
>> That said, if you feel these messages are not clear to some of reek's target audience, perhaps we can improve them so you don't need to change them?

Yes, good idea!
Bordering on that, what do guys think about having the -U switch default to true?

tansaku:
I would love the -U switch default on personally :-)  

regarding updating the messages I guess I can fork and pull.  It's just that I think there are messages that are clear for experienced developers, that are less clear for novice developers.  I'd be happy to develop new messages for novices and share with the community, but then maybe have the type of message toggled with a switch, i.e. to move between terse, and verbose perhaps?

mvz:
> I'd be happy to develop new messages for novices and share with the community

Cool, maybe you can start with the message for one smell and then we can discuss if separate messages are needed or we can reach a middle ground. Talking about a concrete example makes sense here I think.

One thing you must realise, though, is that reek may not be a tool that's suited for novices, because acting on the smells requires some knowledge of OO design. What are your thoughts on this?

troessner:
Let's discuss that here with concrete suggestions. I also think that we can improve our messages for all users.

troessner:
@tansaku are you still interested in this topic? Otherwise I'd close since it doesnt seem that somebody will work on this any time soon.

tansaku:
hi @troessner - funny you should mention that as I was just checking out the reek source to start looking at this.  I've just been out of town for a month and am back now, and want to make working on reek my priority. 

[...]

PR #732, opened 2015-09-29
adding ability to check for smell message contents

tansaku:
So this is another tiny step towards https://github.com/troessner/reek/issues/589 getting set up so that we can actually test that a smell message has changed

### DIFF
--- a/lib/reek/smells/smell_warning.rb
+++ b/lib/reek/smells/smell_warning.rb
@@ -17,6 +17,8 @@ module Reek
       attr_reader :context, :lines, :message, :parameters, :smell_detector, :source
       def_delegators :smell_detector, :smell_category, :smell_type
 
+      COMPARABLE_ATTRIBUTES = %i(message lines context source parameters)
+
       # @note When using reek's public API, you should not create SmellWarning
       #   objects yourself. This is why the initializer is not part of the
       #   public API.
@@ -53,12 +55,17 @@ module Reek
         smell_classes.include?(klass.to_s)
       end
 
-      def matches_smell_details?(other_smell_details = {})
-        common_parameters_equal?(other_smell_details)
+      def matches_attributes?(other_attributes = {})
+        check_attributes_comparability(other_attributes)
+        return false unless common_parameters_equal?(other_attributes[:parameters])
+        other_attributes.delete(:parameters)
+        other_attributes.all? do |other_key, other_value|
+          send(other_key) == other_value
+        end
       end
 
-      def matches?(klass, other_smell_details = {})
-        matches_smell_type?(klass) && matches_smell_details?(other_smell_details)
+      def matches?(klass, other_attributes = {})
+        matches_smell_type?(klass) && matches_attributes?(other_attributes)
       end
 
       def report_on(listener)
@@ -88,25 +95,23 @@ module Reek
         [smell_detector.smell_category, smell_detector.smell_type]
       end
 
-      def common_parameters_equal?(other_parameters)
-        other_keys   = other_parameters.keys
-        other_values = other_parameters.values
+      def check_attributes_comparability(other_attributes)
+        other_attributes.keys.each do |attr|
+          unless COMPARABLE_ATTRIBUTES.include?(attr)
+            raise ArgumentError, "The attribute '#{attr}' is not available for comparison"
+          end
+        end
+      end
 
+      def common_parameters_equal?(other_parameters)
+        return true unless other_parameters
+        other_keys = other_parameters.keys
         other_keys.each do |key|
           unless parameters.key?(key)
             raise ArgumentError, "The parameter #{key} you want to check for doesn't exist"
           end
         end
-
-        # Why not check for strict parameter equality instead of just the common ones?
-        #
-        # In `self`, `parameters` might look like this:  {:name=>"@other.thing", :count=>2}
-        # Coming from specs, 'other_parameters' might look like this, e.g.:
-        # {:name=>"@other.thing"}
-        # So in this spec we are just specifying the "name" parameter but not the "count".
-        # In order to allow for this kind of leniency we just test for common parameter equality,
-        # not for a strict one.
-        parameters.values_at(*other_keys) == other_values
+        parameters.values_at(*other_keys) == other_parameters.values
       end
 
       def core_yaml_hash
--- a/spec/reek/context/module_context_spec.rb
+++ b/spec/reek/context/module_context_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe Reek::Context::ModuleContext do
       module Fred
         def simple(x) x + 1; end
       end
-    ').to reek_of(:UncommunicativeParameterName, name: 'x')
+    ').to reek_of(:UncommunicativeParameterName, parameters: { name: 'x' })
   end
 
   it 'should not report module with empty class' do
--- a/spec/reek/smells/smell_warning_spec.rb
+++ b/spec/reek/smells/smell_warning_spec.rb
@@ -96,12 +96,22 @@ RSpec.describe Reek::Smells::SmellWarning do
     end
 
     it 'matches on class symbol' do
-      expect(uncommunicative.matches?(:UncommunicativeVariableName)).to be_truthy
+      expect(uncommunicative.matches?(:UncommunicativeVariableName)).to\
+        be_truthy
     end
 
     it 'matches on class symbol and params' do
       expect(uncommunicative.matches?(:UncommunicativeVariableName,
-                                      test: 'something')).to be_truthy
+                                      parameters: {
+                                        test: 'something'
+                                      })).to be_truthy
+    end
+
+    it 'matches on class symbol, params and attributes' do
+      expect(uncommunicative.matches?(:UncommunicativeVariableName,
+                                      parameters: { test: 'something' },
+                                      message: "has the variable name '@s'"
+                                     )).to be_truthy
     end
 
     it 'does not match on different class symbol' do
@@ -110,7 +120,23 @@ RSpec.describe Reek::Smells::SmellWarning do
 
     it 'does not match on different params' do
       expect(uncommunicative.matches?(:UncommunicativeVariableName,
-                                      test: 'something else')).to be_falsy
+                                      parameters: {
+                                        test: 'something else'
+                                      })).to be_falsy
+    end
+
+    it 'does not match on different attributes' do
+      expect(uncommunicative.matches?(:UncommunicativeVariableName,
+                                      parameters: { test: 'something' },
+                                      message: 'nothing')).to be_falsy
+    end
+
+    it 'raises error on uncomparable attribute' do
+      expect do
+        uncommunicative.matches?(:UncommunicativeVariableName,
+                                 parameters: { test: 'something' },
+                                 random: 'nothing')
+      end.to raise_error("The attribute 'random' is not available for comparison")
     end
   end
 
--- a/spec/reek/spec/should_reek_of_spec.rb
+++ b/spec/reek/spec/should_reek_of_spec.rb
@@ -4,21 +4,21 @@ require_lib 'reek/spec'
 
 RSpec.describe Reek::Spec::ShouldReekOf do
   describe 'smell type check' do
-    let(:src) { 'def double_thing() @other.thing.foo + @other.thing.foo end' }
+    let(:ruby) { 'def double_thing() @other.thing.foo + @other.thing.foo end' }
 
     it 'reports duplicate calls' do
-      expect(src).to reek_of(:Duplication)
+      expect(ruby).to reek_of(:Duplication)
     end
 
-    it 'does not report any other smell types' do
-      expect(src).not_to reek_of(:FeatureEnvy)
+    it 'does not report any feature envy' do
+      expect(ruby).not_to reek_of(:FeatureEnvy)
     end
   end
 
   describe 'different sources of input' do
     context 'checking code in a string' do
       let(:clean_code) { 'def good() true; end' }
-      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, name: 'y') }
+      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, parameters: { name: 'y' }) }
       let(:smelly_code) { 'def x() y = 4; end' }
 
       it 'matches a smelly String' do
@@ -32,7 +32,7 @@ RSpec.describe Reek::Spec::ShouldReekOf do
 
     context 'checking code in a File' do
       let(:clean_file) { Pathname.glob("#{SAMPLES_PATH}/three_clean_files/*.rb").first }
-      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, name: '@s') }
+      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, parameters: { name: '@s' }) }
       let(:smelly_file) { Pathname.glob("#{SAMPLES_PATH}/two_smelly_files/*.rb").first }
 
       it 'matches a smelly file' do
@@ -47,7 +47,7 @@ RSpec.describe Reek::Spec::ShouldReekOf do
 
   describe 'smell types and smell details' do
     context 'passing in smell_details with unknown parameter name' do
-      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, foo: 'y') }
+      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, parameters: { foo: 'y' }) }
       let(:smelly_code) { 'def x() y = 4; end' }
 
       it 'raises ArgumentError' do
@@ -57,7 +57,7 @@ RSpec.describe Reek::Spec::ShouldReekOf do
 
     context 'both are matching' do
       let(:smelly_code) { 'def x() y = 4; end' }
-      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, name: 'y') }
+      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, parameters: { name: 'y' }) }
 
       it 'is truthy' do
         expect(matcher.matches?(smelly_code)).to be_truthy
@@ -66,8 +66,9 @@ RSpec.describe Reek::Spec::ShouldReekOf do
 
     context 'no smell_type is matching' do
       let(:smelly_code) { 'def dummy() y = 4; end' }
-      let(:falsey_matcher) { Reek::Spec::ShouldReekOf.new(:FeatureEnvy, name: 'y') }
-      let(:truthy_matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, name: 'y') }
+
+      let(:falsey_matcher) { Reek::Spec::ShouldReekOf.new(:FeatureEnvy, parameters: { name: 'y' }) }
+      let(:truthy_matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, parameters: { name: 'y' }) }
 
       it 'is falsey' do
         expect(falsey_matcher.matches?(smelly_code)).to be_falsey
@@ -90,7 +91,7 @@ RSpec.describe Reek::Spec::ShouldReekOf do
 
     context 'smell type is matching but smell details are not' do
       let(:smelly_code) { 'def dummy() y = 4; end' }
-      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, name: 'x') }
+      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, parameters: { name: 'x' }) }
 
       it 'is falsey' do
         expect(matcher.matches?(smelly_code)).to be_falsey
@@ -98,10 +99,9 @@ RSpec.describe Reek::Spec::ShouldReekOf do
 
       it 'sets the proper error message' do
         matcher.matches?(smelly_code)
-
         expect(matcher.failure_message).to\
           match('Expected string to reek of UncommunicativeVariableName (which it did) with '\
-                  'smell details {:name=>"x"}, which it didn\'t')
+                  'smell details {:parameters=>{:name=>"x"}}, which it didn\'t')
       end
 
       it 'sets the proper error message when negated' do
@@ -109,7 +109,7 @@ RSpec.describe Reek::Spec::ShouldReekOf do
 
         expect(matcher.failure_message_when_negated).to\
           match('Expected string not to reek of UncommunicativeVariableName with smell '\
-                  'details {:name=>"x"}, but it did')
+                  'details {:parameters=>{:name=>"x"}}, but it did')
       end
     end
   end
--- a/lib/reek/spec/should_reek_of.rb
+++ b/lib/reek/spec/should_reek_of.rb
@@ -48,7 +48,7 @@ module Reek
       end
 
       def matching_smell_details?
-        matching_smell_types.any? { |warning| warning.matches_smell_details?(smell_details) }
+        matching_smell_types.any? { |warning| warning.matches_attributes?(smell_details) }
       end
 
       def set_failure_messages_for_smell_type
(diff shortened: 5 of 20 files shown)


tansaku:
ah, okay, totally should have rebased first ...

lib/reek/smells/smell_warning.rb
```diff
@@ -109,6 +109,15 @@ def common_parameters_equal?(other_parameters)
         parameters.values_at(*other_keys) == other_values
       end
 
+      def vars_equal?(other_vars)
+        match = true
+        %i(message lines context source).each do |var|
+          other_var = other_vars[var]
+          match &= (send(var) == other_var) if other_var
```
troessner: Please check out "common_parameters_equal?" - this is rather the way to go.

lib/reek/smells/smell_warning.rb
```diff
@@ -57,8 +57,8 @@ def matches_smell_details?(other_smell_details = {})
         common_parameters_equal?(other_smell_details)
       end
 
-      def matches?(klass, other_smell_details = {})
-        matches_smell_type?(klass) && matches_smell_details?(other_smell_details)
+      def matches?(klass, parameters = {}, vars = {})
+        smell_classes.include?(klass.to_s) && common_parameters_equal?(parameters) && vars_equal?(vars)
```
troessner: That's is confusing now. The difference between "parameters" and "vars" is baroque.
I think we should rather beef up "matches_smell_details?" so that it deals with all "parameters" and "attributes" like this:

```Ruby
def matches?(klass, details = {})
  smell_classes.include?(klass.to_s) && matches_smell_details?(details)
end
```

with "details" looking like this:

```Ruby
{
  lines: [...], 
  message: ...,
  parameters: { name: ....}
}
```

And then in "matches_smell_details?" we compare those "details" in a similar, unified way similar to how we do it in "common_parameters_equal?" (which would then be obsolete).

tansaku:
thanks for the comments @troessner - we can do this - and I've done it, but it gives me 100+ errors through the rest of the code base where many other tests have to be updated stuff like: 

```ruby
    it 'reports one-letter variable name' do
      src = 'def simple(fred) x = jim(45) end'
      expect(src).to reek_of(:UncommunicativeVariableName, name: 'x' )
    end
```

to 

```
    it 'reports one-letter variable name' do
      src = 'def simple(fred) x = jim(45) end'
      expect(src).to reek_of(:UncommunicativeVariableName, parameters: { name: 'x' })
    end
```

I suspect it might take me an hour or so to work through all fixing those ... having made the change and got it basically working I'm keen to see it merged in, but there's a lot of overhead here :-)

In the meantime here's the relevant code:

```ruby
      def matches_smell_type?(klass)
        smell_classes.include?(klass.to_s)
      end

      def matches_smell_details?(details = {})
        details_equal?(details)
      end

      def matches?(klass, details = {})
        matches_smell_type?(klass) && matches_smell_details?(details)
      end

      def details_equal?(other_details)
        detail_keys = %i(message lines context source parameters)

        other_details.keys.each do |key|
          unless detail_keys.include?(key)
            raise ArgumentError, "The detail #{key} you want to check for doesn't exist"
          end
        end

        match = true
        other_details.each do |other_key, other_value|
          match &= (send(other_key) == other_value)
        end
        match
      end
```
I'm not so happy with the length of the details equal method, but well anyway, that's a stab at it - will see if I can get all the 100+ errors fixed up and give you a green build on this ...

troessner:
>>       def matches_smell_details?(details = {})
        details_equal?(details)
      end

      def matches?(klass, details = {})
        matches_smell_type?(klass) && matches_smell_details?(details)
      end

AFAIS the "details_equal?" method is now just an unnecessary redirection - we can kill that and just use "matches_smell_details?"

```Ruby
      def details_equal?(other_details)
        detail_keys = %i(message lines context source parameters)

        other_details.keys.each do |key|
          unless detail_keys.include?(key)
            raise ArgumentError, "The detail #{key} you want to check for doesn't exist"
          end
        end

        match = true
        other_details.each do |other_key, other_value|
          match &= (send(other_key) == other_value)
        end
        match
      end
```

Using "&=" always feels weird and is normally not necessary in Ruby.
You can just do:

```Ruby
other_details.all? do |other_key, other_value|
  send(other_key) == other_value
end
```

and get rid of the whole "match" / boolean thingie. You normally never need booleans in combination with Enumerable's in ruby.

Additionally, you need to cater to "parameters" in "details_equal?" which it doesn't seem to do anymore?

tansaku:
thanks for those tips @troessner - I've incorporated them into the lastest code update  - see what you think

lib/reek/smells/smell_warning.rb
```diff
@@ -17,6 +17,8 @@ class SmellWarning
       attr_reader :context, :lines, :message, :parameters, :smell_detector, :source
       def_delegators :smell_detector, :smell_category, :smell_type
 
+      DETAIL_KEYS = %i(message lines context source parameters)
```
troessner: "details_key" is not really expressive - how about COMPARABLE_ATTRIBUTES ?

lib/reek/smells/smell_warning.rb
```diff
@@ -88,25 +95,23 @@ def smell_classes
         [smell_detector.smell_category, smell_detector.smell_type]
       end
 
-      def common_parameters_equal?(other_parameters)
-        other_keys   = other_parameters.keys
-        other_values = other_parameters.values
+      def check_details_exist(other_details)
+        other_details.keys.each do |key|
+          unless DETAIL_KEYS.include?(key)
+            raise ArgumentError, "The detail #{key} you want to check for doesn't exist"
```
troessner: Same here, let's rather use "attribute".

lib/reek/smells/smell_warning.rb
troessner: We compare two things:

1.) "attributes": message lines context source parameters
and
2.) "parameters". This "attribute" has "keys". So let's use "keys" only in the context of "parameters", but not in the context of "attributes"

lib/reek/smells/smell_warning.rb
```diff
@@ -53,12 +55,17 @@ def matches_smell_type?(klass)
         smell_classes.include?(klass.to_s)
       end
 
-      def matches_smell_details?(other_smell_details = {})
-        common_parameters_equal?(other_smell_details)
+      def matches_smell_details?(other_details = {})
```
troessner: This method is very confusing because of the naming of the methods we use in there. 
I also think I made an error when I suggested using "_details". I believe we should rather name everything here and below "_attributes", because this is in fact what we're checking for - "attributes", not "details". The wording "details" made sense before but doesn't anymore.

How about doing something like this:

```Ruby
def matches_attributes(other_attributes = {})
  return false unless common_parameters_equal?(other_attributes[:parameters])
  other_attributes.delete(:parameters)
  other_attributes.all? do |other_key, other_value|
    send(other_key) == other_value
  end
end
```

And then use "check_details_exist" only in "common_parameters_equal?" like it was before. 

```Ruby
    send(other_key) == other_value
```

will raise on invalid attributes in any case.

And then we don't need "check_details_exist" at all and you can just put the method body at the beginning of "common_parameters_equal?"

lib/reek/smells/smell_warning.rb
troessner: Actually just see my comment directly below regarding "matches_smell_details?", that should make everthing clear ;)

lib/reek/smells/smell_warning.rb
troessner: Actually just see my comment directly above regarding "matches_smell_details?", that should make everthing clear ;)

tansaku:
thanks @troessner - have made some updates based on your suggestions.  Take a look.

I wasn't sure about putting the body of `check_details_exist` in with the common_parameters_equal? since that would make that method too long (and have it do two things).  Also if I remove that check from `matches_attributes?` reek flags it with feature envy ...

See how you like this latest variation

lib/reek/smells/smell_warning.rb
```diff
@@ -53,12 +55,17 @@ def matches_smell_type?(klass)
...
+        check_attributes_comparability(other_attributes)
+        return false unless common_parameters_equal?(other_attributes[:parameters])
+        other_attributes.delete(:parameters)
+        other_attributes.all? do |other_key, other_value|
+          send(other_key) == other_value
+        end
       end
 
-      def matches?(klass, other_smell_details = {})
-        matches_smell_type?(klass) && matches_smell_details?(other_smell_details)
```
troessner: Let's also rename that to "attributes" here and below.

lib/reek/smells/smell_warning.rb
```diff
@@ -88,25 +95,23 @@ def smell_classes
...
+      def check_attributes_comparability(other_attributes)
+        other_attributes.keys.each do |attr|
+          unless COMPARABLE_ATTRIBUTES.include?(attr)
+            raise ArgumentError, "The attribute #{attr} you want to check for doesn't exist"
+          end
+        end
+      end
 
+      def common_parameters_equal?(other_parameters)
+        return true unless other_parameters
```
troessner: This return is unnecessary (that's why it wasnt there before) and de facto a hidden type check, lets drop this line.

troessner:
Almost there - some small changes and then lets get this merged ;)

spec/reek/smells/smell_warning_spec.rb
```diff
@@ -87,30 +87,42 @@
   end
 
   context '#matches?' do
-    let(:uncommunicative) do
+    let(:uncomm) do
```
troessner: Why this change? "uncomm" is a really bad name, lets revert that.

troessner:
@tansaku please adress my latest comments and squash to one commit, then we can merge this bad boy ;)

spec/reek/smells/smell_warning_spec.rb
tansaku: this was to avoid the lines getting too long ...

lib/reek/smells/smell_warning.rb
tansaku: I get errors if I remove this.  We have situations where we are not defining any parameters to match against - note how we are calling it:

```ruby
      def matches_attributes?(other_attributes = {})
        check_attributes_comparability(other_attributes)
        return false unless common_parameters_equal?(other_attributes[:parameters])
```

so if other_attributes has no :parameters key (which is often the case) then `common_parameters_equal?` often gets called with a nil which we need to defend against

lib/reek/smells/smell_warning.rb
tansaku: done

tansaku:
@troessner 2 of 3 addressed - pushing back on one (see above) - will squash once your satisfied :-)

tansaku:
or should I squash in anticipation ...

tansaku:
hmm, squashing much harder than I thought - complex conflicts in many more files - have made my way through them, but now stuck on two fails:

```sh

  1) Reek::Spec::ShouldReekOf smell types and smell details smell type is matching but smell details are not sets the proper error message
     Failure/Error: expect(matcher.failure_message).to\
       expected "Expected string to reek of UncommunicativeVariableName, but it didn't" to match "Expected string to reek of UncommunicativeVariableName (which it did) with smell details {:parameters=>{:name=>\"x\"}}, which it didn't"
     # ./spec/reek/spec/should_reek_of_spec.rb:110:in `block (4 levels) in <top (required)>'

  2) Reek::Spec::ShouldReekOf smell types and smell details smell type is matching but smell details are not sets the proper error message when negated
     Failure/Error: expect(matcher.failure_message_when_negated).to\
       expected "Expected string not to reek of UncommunicativeVariableName, but it did" to match "Expected string not to reek of UncommunicativeVariableName with smell details {:parameters=>{:name=>\"x\"}}, but it did"
     # ./spec/reek/spec/should_reek_of_spec.rb:118:in `block (4 levels) in <top (required)>'

Finished in 2.11 seconds (files took 1.03 seconds to load)
804 examples, 2 failures

```

which seem to relate to changes in `lib/reek/spec/should_reek_of` where we've moved from:

```ruby
      def matches?(actual)
        self.examiner = Examiner.new(actual, configuration: configuration)
        set_failure_messages
        matching_smell_types? && matching_smell_details?
      end
```

to 

```
     def matches?(actual)
        self.examiner = Examiner.new(actual, configuration: configuration)
        self.all_smells = examiner.smells
        all_smells.any? { |warning| warning.matches?(smell_category, smell_details) }
      end
```

so the failure messages are not being set in quite the same way they were before ...?

So to fix this I'm not sure which direction others are going in - are those tests that I have failing at the moment being dropped since we are going with simpler error messages that don't distinguish between smell type and associated attribute matches?

Either way I'm confused as I thought I was rebasing to master, and master on github doesn't seem to have the same updated code as I've just pulled in ...

lib/reek/smells/smell_warning.rb
troessner: Ok, let's leave it like that.

troessner:
Code looks good to me! Now we only need to sort out your problems above ;)
I'm on a plane the whole day tomorrow - either you figure it out on your own or I can have a more detailed look at the problem on saturday / sunday ;)

tansaku:
thanks @troessner I've force pushed the single squashed commit so you can see precisely where we're up to.  I'm not going to be able to look at this again till Monday, so if you can see the resolution here that would be very helpful.

I believe I could resolve it either by removing the two failing tests, or by changing (reverting?) `lib/reek/spec/should_reek_of` to the state that supports those two tests, but I don't which the project wants and maybe I have misdiagnosed the issue and there is something more subtle going on. 

Your help very much appreciated

troessner:
>> This branch has conflicts that must be resolved

You need to rebase against origin/master (which is basically the master the from this very repository), fix the conflicts and then force push.
There's a high chance that the failing tests will then be magically green ;)
If not, I can take a look.

tansaku:
That's what I thought I did.  I'll have another look later today.  If I'm doing 'git rebase -i master' I need to do 'git fetch upstream' and git merge upstream/master first?

I thought I did do that already but will double check

tansaku:
okay, looks like the problem was that I forget the `git merge upstream/master` step - maybe I should just do `git rebase -i upstream/master` generally.  

Anyway, green now.  Hopefully this should be ready to merge.  

Ironically even after all this, this only completes clearing my throat to get started on smell message internationalisation for https://github.com/troessner/reek/issues/589 :-)

spec/reek/spec/should_reek_of_spec.rb
```diff
@@ -28,11 +40,16 @@
       it 'doesnt match a fragrant String' do
         expect(matcher.matches?(clean_code)).to be_falsey
       end
+
+      it 'reports the smells when should_not fails' do
+        expect(matcher.matches?(smelly_code)).to be_truthy
+        expect(matcher.failure_message_when_negated).to match('UncommunicativeVariableName')
```
troessner: Where is this example coming from all of a sudden? I'd rather remove this.

spec/reek/spec/should_reek_of_spec.rb
```diff
@@ -4,21 +4,33 @@
...
+
+    it 'reports duplicate calls to @other.thing' do
+      expect(ruby).to reek_of(:Duplication, parameters: { name: '@other.thing' })
+    end
+
+    it 'reports duplicate calls to @other.thing.foo' do
+      expect(ruby).to reek_of(:Duplication, parameters: { name: '@other.thing.foo' })
+    end
+
+    it 'does not report any feature envy' do
```
troessner: The example description is now misleading. It's not about FeatureEnvy, it's about "does not report any other smell types" like in the old example description -> let's change that back to what it was before.

spec/reek/spec/should_reek_of_spec.rb
```diff
@@ -4,21 +4,33 @@
...
-      expect(src).to reek_of(:Duplication)
+      expect(ruby).to reek_of(:Duplication)
     end
 
     it 'does not report any other smell types' do
-      expect(src).not_to reek_of(:FeatureEnvy)
+      expect(ruby).not_to reek_of(:FeatureEnvy)
+    end
+
+    it 'reports duplicate calls to @other.thing' do
```
troessner: Why are the next two examples introduced here? It seems like they are kind of copied from https://github.com/troessner/reek/blob/master/spec/reek/smells/duplicate_method_call_spec.rb ?
You can safely remove those.

spec/reek/smells/smell_warning_spec.rb
```diff
@@ -96,12 +96,22 @@
...
 
     it 'matches on class symbol and params' do
       expect(uncommunicative.matches?(:UncommunicativeVariableName,
-                                      test: 'something')).to be_truthy
+                                      parameters: {
+                                        test: 'something'
+                                      })).to be_truthy
+    end
+
+    it 'matches on class symbol, params and vars' do
```
troessner: "vars" is not a good name here and below (highly ambiguous). We introduced the "attributes" terminology above and should stick with it.

spec/reek/smells/smell_warning_spec.rb
```diff
@@ -110,7 +120,15 @@
...
+                                      parameters: {
+                                        test: 'something else'
+                                      })).to be_falsy
+    end
+
+    it 'does not match on different vars' do
+      expect(uncommunicative.matches?(:UncommunicativeVariableName,
+                                      parameters: { test: 'something' },
+                                      message: 'nothing')).to be_falsy
     end
```
troessner: SmellWarning#matches_attributes? from here https://github.com/troessner/reek/pull/732/files#diff-4c9efdbf77261fe0d68c190a1db4e3fcR58 is a very powerful method now that is not spec'ed at all it sees. Can you please add some specs?

troessner:
I'm afraid I still can't merge this since I found some other issues in my final review, see my latest comments. Please address those and then we can merge this.;)

tansaku:
thanks @troessner hoping to get to this soon - I already had some other additional tests for that long method that I was going to drop into a separate pull request - busy week - will try to get to this before the end of it :-)

spec/reek/spec/should_reek_of_spec.rb
tansaku: I think it might just be due to the funky rebase I did against an old master

spec/reek/spec/should_reek_of_spec.rb
tansaku: I think it might just be due to the funky rebase I did against an old master - sorry about that

spec/reek/context/module_context_spec.rb
```diff
@@ -8,7 +8,7 @@
       module Fred
         def simple(x) x + 1; end
       end
-    ').to reek_of(:UncommunicativeParameterName, name: 'x')
+    ').to reek_of(:UncommunicativeParameterName, parameters: { name: 'x' })
```
mvz: I think we should strive for a solution that doesn't need this extra key.

lib/reek/smells/smell_warning.rb
```diff
@@ -53,12 +55,17 @@ def matches_smell_type?(klass)
         smell_classes.include?(klass.to_s)
       end
 
-      def matches_smell_details?(other_smell_details = {})
-        common_parameters_equal?(other_smell_details)
+      def matches_attributes?(other_attributes = {})
+        check_attributes_comparability(other_attributes)
+        return false unless common_parameters_equal?(other_attributes[:parameters])
+        other_attributes.delete(:parameters)
```
mvz: Why not invert this logic, so you take out the values for the keys message, lines, context and source, and assume the rest are parameters? This avoids the extra indirection for the parameters.

lib/reek/smells/smell_warning.rb
```diff
@@ -88,25 +95,23 @@ def smell_classes
...
+          end
+        end
+      end
 
+      def common_parameters_equal?(other_parameters)
+        return true unless other_parameters
+        other_keys = other_parameters.keys
         other_keys.each do |key|
-          unless parameters.key?(key)
+          unless parameters.include?(key)
```
mvz: Why was this changed?

spec/reek/spec/should_reek_of_spec.rb
tansaku: investigating and I didn't write this.  It is an artifact of the funky rebase - this file had huge merge conflicts and I did my best to get it into the right shape, but because my local master was behind the upstream master it all went a bit wrong - have removed this test

spec/reek/spec/should_reek_of_spec.rb
tansaku: Here, this is a block of tests I didn't write.  I'm just making minor changes here to fit in with the other changes you suggested.  Again I think this may be down to the funky rebase.  I think what's happened here is that the old and new versions of a test are both present, so I'm going to delete what looks like the older one that I think I accidentally pulled in via the rebase to an old master.

spec/reek/spec/should_reek_of_spec.rb
tansaku: removing these as the diff certainly indicates I'm adding something here that isn't in the current master.  Again, I didn't write these - they were just what appeared in the file when I got massive merge conflicts with the old master

spec/reek/smells/smell_warning_spec.rb
tansaku: agreed - changed - that was an oversight not changing it retrospectively rather than an intentional insertion

spec/reek/smells/smell_warning_spec.rb
tansaku: I had a thought already about how we are not testing attribute match failure and had already written a test for that which I thought to include in a subsequent pull request.  I include that now, although that is still in the SmellWarning#matches? context.  Various of the code paths of SmellWarning#matches_attributes? are tested via the SmellWarning#matches? context.  Should they be broken out into a separate context? tested in both (no I guess) or should SmellWarning#matches_attributes? become a private method?

spec/reek/context/module_context_spec.rb
tansaku: That sounds like a noble goal, but this is a reflection of the instance variables we have on smell_warning:

```ruby
      def initialize(smell_detector, context: '', lines: raise, message: raise,
                                     source: raise, parameters: {})
        @smell_detector = smell_detector
        @source         = source
        @context        = context.to_s
        @lines          = lines
        @message        = message
        @parameters     = parameters
      end
```

parameters is a hash - the others are not.  If we want to match against both the instance variables where one is a hash, we need to distinguish them with keys.  Maybe all the parameters should become instance variables to avoid the need for the nested parameters so to speak, but getting involved in this debate I feel a strong sense of "Yak-shaving" :-) I'm all for improving reek, but I really want to get to adding the internationalization functionality for which having the ability to check the message associated with a smell is a pre-requisite.

Maybe when making these changes is a great opportunity for a larger refactoring, but I don't feel qualified to do it.  Particularly I feel I am so far away from having an end user feature to drive the decisions that I can't really argue for or against this.

lib/reek/smells/smell_warning.rb
tansaku: not sure - hadn't noticed the difference - have changed it back

lib/reek/smells/smell_warning.rb
tansaku: that's an interesting idea -  but wouldn't it be cleaner in the long run to just make all the parameter options into instance variables on smell_warning? and then we don't have what seems like an artificial distinction between some aspects of the smell_warning that are instance vars, and others that are collected into a open ended group called "parameters", but as I say that feels like a large refactoring that I don't feel qualified to do ... yet :-)

troessner: I'm with @tansaku on this one. I believe our current structure is not quite there but we're on a good way.
I'd rather see this PR finished and merged and then address this in a follow-up.

>> and others that are collected into a open ended group called "parameters"

Exactly. I believe a cleaner, more OOP solution would be to have a class called SmellDetails and then rename the "parameters" attribute above "details" and point that to an instance. But let's leave that for a separate discussion.

spec/reek/context/module_context_spec.rb
troessner: Agree, see above.

troessner:
I re-reviewed this one and I am of the opinion that we should merge it now.
This: https://github.com/troessner/reek/pull/732/files#diff-4c9efdbf77261fe0d68c190a1db4e3fcR61
would be a separate refactoring that we can address later when we feel like.
@mvz any objections?

lib/reek/smells/smell_warning.rb
mvz: Ok, let's address this in a follow-up. We just need to make sure we deal with this before the next release so we don't break the API.

mvz:
> @mvz any objections?

No objections. LGTM!

troessner:
Merged!
Thanks a lot @tansaku !

tansaku:
thanks @troessner and @mvz pleased to see that merged in - now onto the real work :-)

PR #735, opened 2015-10-03
Refactor NestedIterators.

troessner:
While working on #729 I figured that we have to refactor NestedIterators.

Why, you ask?

1.) It's spaghetti code at it's best. "find_iters" calls "find_iters_for_iter_node". Which then calls "find_iters". Which then calls "find_iters_for_iter_node". You see where I'm going with this, don't you? ;)
2.) It's not a real recursion (see [1]) which it should be, given that we are traversing trees
3.) It's very hard to read and understand. I'd call that "unmaintainable"
4.) It's not documented at all.

So, let's fix all the things!

Please find below my humble attempt to improve this.
I believe the refactoring below addresses all of the issues mentioned above and will allow us to extend NestedIterators in the future if necessary.

### DIFF
--- a/lib/reek/smells/nested_iterators.rb
+++ b/lib/reek/smells/nested_iterators.rb
@@ -10,6 +10,16 @@ module Reek
     #
     # See {file:docs/Nested-Iterators.md} for details.
     class NestedIterators < SmellDetector
+      # Struct for conveniently associating iterators with their depth (that is, their nesting).
+      Iterator = Struct.new :exp, :depth do
+        include Comparable
+        def <=>(other)
+          depth <=> other.depth
+        end
+      end
+
+      private_attr_accessor :ignore_iterators
+
       # The name of the config field that sets the maximum depth
       # of nested iterators to be permitted within any single method.
       MAX_ALLOWED_NESTING_KEY = 'max_allowed_nesting'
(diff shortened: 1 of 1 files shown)


lib/reek/smells/nested_iterators.rb
```diff
@@ -10,6 +10,16 @@ module Smells
     #
     # See {file:docs/Nested-Iterators.md} for details.
     class NestedIterators < SmellDetector
+      # Struct for conveniently associating iterators with their depth (that is, their nesting).
+      IteratorToDepth = Struct.new :iterator, :depth do
+        include Comparable
+        def <=>(other)
+          depth <=> other.depth
+        end
+      end
```
chastell: I really dislike the `IteratorToDepth` name. How about just `Iterator`, or `IteratorWithDepth` (bleh…)?

troessner: I renamed it to "Iterator" now ;)

chastell: OK, I feel really bad now, but how about renaming the `iterator` field to `exp`? `Iterator` being an `iterator` + `depth` sound trippy.

Also: have you considered moving `NestedIterators#ignored_iterator?` to `Iterator#ignored?` – this means `Iterator` would have to know about `NestedIterators#ignore_iterators`, but this seems doable. I’m OK if the naswer is ‘yes, let’s do it in a separate PR’. :)

troessner: >> OK, I feel really bad now, but how about renaming the iterator field to exp? Iterator being an iterator + depth sounds trippy.

Good point, changed!

>> Also: have you considered moving NestedIterators#ignored_iterator? to Iterator#ignored? – this means Iterator would have to know about NestedIterators#ignore_iterators, but this seems doable. I’m OK if the answer is ‘yes, let’s do it in a separate PR’. :)

I like the idea but let's do that in a follow up PR ;)

chastell:
OK, it’s a clear improvement (thanks!), merging!

PR #741, opened 2015-10-12
Bump Parser

chastell:
Parser 2.2.3.0 got released and it seems Reek can use it just fine.

I’m unhappy with this PR, though: some time ago we agreed that Parser’s versioning is at odds with SemVer and – since it’s crucial to Reek’s operations – we should be strict about this dependency. That said, making us depend on `parser ~> 2.2.3.0` means that Reek can’t be used alongside tools with a Parser 2.2.2.x dependency; one could argue that this is actually a breaking change.

We know that Reek works with Parser 2.2.2.5+ and assume it’ll work with any 2.2.3.x. Maybe we should do the dependency like this?

```Ruby
 s.add_runtime_dependency 'parser', '>= 2.2.2.5', '< 2.2.4'
```

### DIFF
--- a/reek.gemspec
+++ b/reek.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.0.0'
   s.summary = 'Code smell detector for Ruby'
 
-  s.add_runtime_dependency 'parser',       '~> 2.2.2.5'
+  s.add_runtime_dependency 'parser',       '~> 2.2', '>= 2.2.2.5'
   s.add_runtime_dependency 'private_attr', '~> 1.1'
   s.add_runtime_dependency 'rainbow',      '~> 2.0'
   s.add_runtime_dependency 'unparser',     '~> 0.2.2'


troessner:
>> Parser 2.2.3.0 got released and it seems Reek can use it just fine.

I see what you Did there!

chastell:
Do you have any Thoughts about the question in the PR, though? :)

(I don’t like the verbose dependency, but I think it’s better than being stuck with Parser 2.2.2.x until Reek 4.0 and bumping it with a `~> 2.2.3.0` is pushing a hard dep bump on others that is not necessary.)

troessner:
>>  s.add_runtime_dependency 'parser', '>= 2.2.2.5', '< 2.2.4'

If we're doing it like that then we can't rely on features from parser 2.2.3.0 in `reek` - why would we then upgrade at all?
Additionally what features in 2.2.3.0 prompted you to create this PR?

chastell:
> If we're doing it like that then we can't rely on features from parser 2.2.3.0 in reek - why would we then upgrade at all?

We don’t have to _upgrade_, but if we know that Reek works with both 2.2.2.x and 2.2.3.x it would be good that we allowed both versions – if, say, RuboCop upgrades to 2.2.3.x and we stick to 2.2.2.x then it won’t be possible to load the newest versions of RuboCop and Reek together.

>  Additionally what features in 2.2.3.0 prompted you to create this PR?

`bundle outdated` yelling at me that we have outdated dependencies. :) But, as I said, I’d rather allow both 2.2.2.5+ and 2.2.3.x than _require_ 2.2.3.x, so I’m open to altering this PR.

I also want to establish our default approach for these cases in the future, e.g. we don’t bump Ruby version dependency on minor releases, so we support Ruby 2.0 in Reek 3.x – the question is whether this also applies to other dependencies (I’d say: runtime deps yes, development deps no).

Cc: @mvz :)

mvz:
> the question is whether this also applies to other dependencies (I’d say: runtime deps yes, development deps no).

This is a crucial question I'd like to be answered for SemVer as a whole.

Some thoughts:

* We should be allowed to bumb dependencies by a point or even minor release if they fix bugs. We've done this before and it doesn't make sense not to, especially if those dependencies use SemVer themselves.
* Should we be allowed to upgrade to a new major release? I'm less sure here, since I can see scenarios where this breaks downstream projects.
* I haven't been careful about this **at all** in my projects, but I think it's a good thing to have a policy for.
* Ruby version support is a slightly different beast since rubygems is very bad at not trying to install gems that do not support the currently running version of Ruby.

> I’d rather allow both 2.2.2.5+ and 2.2.3.x than require 2.2.3.x, so I’m open to altering this PR.

Yes, so would I. We might even go so far as saying '>= 2.2.2.5, < 2.3'.

troessner:
>> Yes, so would I. We might even go so far as saying '>= 2.2.2.5, < 2.3'.

Then why not go the whole way and do 

>> ~> 2

?

mvz:
> Then why not go the whole way and do `~> 2`?

You mean as in greater than or equal to 2? Then we can just keep only `>= 2.2.2.5`. I'm not against that because we can assume reasonable compatibility of Ruby 3 syntax and up with Ruby 2, at least insofar as if Reek runs on it, it'll probably be ok. Also, the parser gem is [guaranteed not to make breaking API changes](https://github.com/whitequark/parser/blob/master/CONTRIBUTING.md).

chastell:
> We should be allowed to bumb dependencies by a point or even minor release if they fix bugs. We've done this before and it doesn't make sense not to, especially if those dependencies use SemVer themselves.

Ok, so our minor releases can bump our deps’ minor releases. I think I’m ok with this. :)

> Should we be allowed to upgrade to a new major release? I'm less sure here, since I can see scenarios where this breaks downstream projects.

I’d go with a tenative ‘no’ and come back when we have an actual case like this.

> We might even go so far as saying '>= 2.2.2.5, < 2.3'.

This would be `~> 2.2.2’, except know that the 2.2.2.2 versions is problematic.

> Then why not go the whole way and do `~> 2`

This means ‘anything greater than 2.0.0.0’, including 3, 4, etc.’ – we definitely don’t want that. If Parser was following SemVer we should just do `~> 2.2` (anything from 2.2.0.0 up until, but excluding, 3.0), but we had problems with Parser versioning in the past: https://github.com/troessner/reek/pull/523

I made the depenency be ‘2.2.2.5 or higher, up to – but excluding – 3.0’.

mvz:
> This would be `~> 2.2.2’, except know that the 2.2.2.2 versions is problematic.

Yes, we definitely don't want that so we will definitely need to exclude 2.2.2.4 and lower.

> I made the depenency be ‘2.2.2.5 or higher, up to – but excluding – 3.0’.

I'm ok with this. We should be fine with that strategy until Ruby 3.0 comes out.

chastell:
…sooo this is green and ready to merge, then. :)

Issue #738, opened 2015-10-11
Unify SmellWarning parameters


mvz:
As a follow-up to #732, unify SmellWarning parameters and the method for testing them using `#reek_of`.

troessner:
Closing this one since #742 got merged.

PR #742, opened 2015-10-14
Unified smell warning interface

mvz:
See #738.

The ultimate goal is for all of SmellWarning's attributes to be on equal footing. This was already the case for the json/yaml hash output.

This restores the `.reek_of` interface.

### DIFF
--- a/lib/reek/spec/smell_matcher.rb
+++ b/lib/reek/spec/smell_matcher.rb
@@ -1,0 +1,59 @@
+module Reek
+  module Spec
+    #
+    # Matches a +SmellWarning+ object agains a smell type and hash of attributes.
+    #
+    class SmellMatcher
+      attr_reader :smell_warning
+
+      COMPARABLE_ATTRIBUTES = %i(message lines context source)
+
+      def initialize(smell_warning)
+        @smell_warning = smell_warning
+      end
+
+      def matches?(klass, attributes = {})
+        matches_smell_type?(klass) && matches_attributes?(attributes)
+      end
+
+      def matches_smell_type?(klass)
+        smell_classes.include?(klass.to_s)
+      end
+
+      def matches_attributes?(attributes)
+        check_attributes_comparability(attributes)
+
+        # FIXME: Use Array#to_h when dropping Ruby 2.0 compatibility.
+        fields, params = attributes.
+          partition { |key, _| COMPARABLE_ATTRIBUTES.include? key }.
+          map { |arr| Hash[arr] }
+
+        common_parameters_equal?(params) &&
+          common_attributes_equal?(fields)
+      end
+
+      private
+
+      def smell_classes
+        [smell_warning.smell_category, smell_warning.smell_type]
+      end
+
+      def check_attributes_comparability(other_attributes)
+        parameter_keys = other_attributes.keys - COMPARABLE_ATTRIBUTES
+        extra_keys = parameter_keys - smell_warning.parameters.keys
+        return if extra_keys.empty?
+        raise ArgumentError, "The attribute '#{extra_keys.first}' is not available for comparison"
+      end
+
+      def common_parameters_equal?(other_parameters)
+        smell_warning.parameters.slice(*other_parameters.keys) == other_parameters
+      end
+
+      def common_attributes_equal?(attributes)
+        attributes.all? do |other_key, other_value|
+          smell_warning.send(other_key) == other_value
+        end
+      end
+    end
+  end
+end
--- a/lib/reek/smells/smell_warning.rb
+++ b/lib/reek/smells/smell_warning.rb
@@ -9,6 +9,7 @@ module Reek
     # @public
     #
     # :reek:TooManyInstanceVariables: { max_instance_variables: 6 }
+    # :reek:TooManyStatements: { max_statements: 6 }
     class SmellWarning
       include Comparable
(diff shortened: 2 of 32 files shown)


lib/reek/spec/smell_matcher.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      def matches_smell_type?(klass)
+        smell_classes.include?(klass.to_s)
+      end
+
+      def matches_attributes?(attributes)
+        check_attributes_comparability(attributes)
+
+        main_attributes = attributes.slice(*COMPARABLE_ATTRIBUTES)
+        parameter_attributes = attributes.except(*COMPARABLE_ATTRIBUTES)
```
chastell: I wonder whether the below wouldn’t be a tad clearer:

```ruby
comparables, params = attributes.partition { |attr| COMPARABLE_ATTRIBUTES.include?(attr) }
```

chastell:
Wow, this was a lot of work, but I like the resulting API. I don’t have enough experience to dive deeper into the implementation (other than a minor nitpick), but LGTM. @troessner?

lib/reek/spec/smell_matcher.rb
mvz: Yes, I like!

mvz: But it returns an array of arrays :crying_cat_face:. Also, activesupport must die: This whole thing accidentally won't work if the downstream user doesn't use activesupport!

chastell: Wat? I would _never_ suggest to infest a codebase with Active Support, [`Enumerable#partition` is core Ruby](http://www.rubydoc.info/stdlib/core/Enumerable#partition-instance_method).

I’m also not sure what you mean by ‘it returns an array of arrays’ – it does, that’s why I suggested a destructuring assign `comparables, params = …`.

chastell: Oh, I read the fixing commit, this was a hash and we need hashes… Sadly `Array#to_h` was added in Ruby 2.1, so we can’t use it till Reek 4.0, so I think `Hash#partition#map` is the best we can do for now.

chastell:
Can you fix the syntax and do something to the commits so that we don’t have known-red ones? :)

lib/reek/spec/smell_matcher.rb
mvz: Yes, it was the original implementation that accidentally depended on activesupport.

mvz:
Done!

lib/reek/spec/smell_matcher.rb
troessner: Y U all hatin' on activesupport???

mvz: 'cause it spanks too many monkeys!

But seriously, having it in our tests proves a bit dangerous because we might accidentally use it in our library code.

chastell: > Y U all hatin' on activesupport???

Because it monkey-patches core classes.

It’s also very popular and many projects use antiquated versions (matching the rest of their Rails stack), so if we were depending on its newest version then Reek couldn’t be used with, say, Rails 3 projects (which I just upgraded from Rails 2.3!).

troessner: Ok, we need a sarcasm emoji for our github discussions, I wasn't serious, I'm also not the biggest fan of AS for the very reasons you list above ;)

mvz: Yeah I wasn't sure so I gave a serious answer anyway :laughing:.

There's always: :interrobang:

troessner: >> There's always: :interrobang:

Ah, I'll keep that in mind! :laughing:

chastell:
Sorry, two more nitpicks:

1. Can you alter the commit that failed the build so that we don’t merge known-red commits? I might be a bit over-zealous here, but I really like to be able to git-bisect assuming every commit builds.
1. Can you move the `FIXME` from a comment to an issue (‘port all `Hash.[]` calls to `Array#to_h`)? I think I agree with Timo that it’s better to track them this way (or, in this case, just fix it on the `reek-4` branch).

mvz:
> Can you move the FIXME from a comment to an issue (‘port all Hash.[] calls to Array#to_h)?

I'm not sure I like having those things in separate tickets. These FIXMEs are highly localized and otherwise we have to start grepping for `Hash[` and things like that. Having these things marked makes it easier to find, and avoids having countless trivial tickets hanging around.

mvz:
> I might be a bit over-zealous here, but I really like to be able to git-bisect assuming every commit builds.

Well, it is the quality check that fails, but OK :-).

chastell:
> These FIXMEs are highly localized and otherwise we have to start grepping for `Hash[` and things like that.

Well, I’d say that’s a point for having a blanket issue for this – we want to switch all instances of `Hash.[]` to `Array#to_h`, not just the one(s) marked with `FIXME`. :)

> Well, it is the quality check that fails, but OK :-).

Aren’t all tests quality checks? ;)

Thanks, merging!

mvz:
> switch all instances of `Hash.[]` to `Array#to_h`.

Yes, but how many are there :stuck_out_tongue_closed_eyes:.

Issue #662, opened 2015-08-22
Make AppConfiguration.from_map take a simple Hash


troessner:
Follow-up issue from https://github.com/troessner/reek/pull/623#discussion-diff-37694166R113

Let's change `AppConfiguration.from_map` so that we can do this: 

```Ruby
hash = YAML.load_file(Pathname.new('config.reek'))
configuration = Reek::Configuration::AppConfiguration.from_map hash
```

troessner:
#743 merged -> closing!

troessner:
>> #743 merged -> closing!

Seems like I actually forgot to close it :laughing: 
Done!

PR #743, opened 2015-10-14
Unify config loading

mvz:
#662.

I decided to create a new method `.from_hash` in the end because I think this is a nicer name and we can more simply deprecate the entire `.from_map` method. We can remove `.from_map` in Reek 4.

Internally, `.from_map` is still used at the moment.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -73,44 +73,25 @@ IrresponsibleModule:
   enabled: false
 ```
 
+Reek will load this file automatically by default. If you want to load the
+configuration explicitely, you can use one of the methods below.
+
 You can now use either
 
 ```Ruby
 Reek::Configuration::AppConfiguration.from_path Pathname.new('config.reek')
 ```
 
-but you can also pass a hash via `Reek::Configuration::AppConfiguration.from_map`.
+but you can also pass a hash with the contents of the `config.reek` YAML file
+to `Reek::Configuration::AppConfiguration.from_hash`.
 
-This hash can have the following 3 keys:
-
-1.) directory_directives [Hash] for instance:
-
-```Ruby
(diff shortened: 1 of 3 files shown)


docs/API.md
```diff
@@ -73,44 +73,25 @@ IrresponsibleModule:
...
-be the default configuration for smell types for which there is
-no "directory directive" (so a directory-specific configuration):
+Given the example above you should load that as follows:
 
 ```Ruby
 require 'reek'
 
-default_directive = { Reek::Smells::IrresponsibleModule => { 'enabled' => false } }
-configuration = Reek::Configuration::AppConfiguration.from_map default_directive: default_directive
+config_hash = { 'IrresponsibleModule' => { 'enabled' => false } }
```
troessner: How about adding a "directory directive" and an "exclude path" to the config_hash here so the example is actually comprehensive?

troessner:
One comment, rest looks good and LGTM.

mvz:
I decided to add a full example separately, since I like the flow of the story when we can refer to the earlier config and show how that particular one can be set programmatically.

mvz:
I'll squash once everything is ok and green.

troessner:
>> I decided to add a full example separately, since I like the flow of the story when we can refer to the earlier config and show how that particular one can be set programmatically.

Ok, understood.

troessner:
>>I'll squash once everything is ok and green.

Greeeeeen ;)

mvz:
@troessner, @chastell Does either of you have an idea why Ataru keeps failing on Ruby 2.2?

mvz:
:green_apple: :green_book: :green_heart:

PR #745, opened 2015-10-18
[WIP] Bury dead code and introduce `debride`.

troessner:


troessner:
Unfortunately the amount of false-positives given by debride is very high. Of course everything that is dynamically assembled at runtime but also a couple of other false-positives I **think** debride could detect better. But people probably say the same thing about `reek` as well ;)

troessner:
Back to WIP, I'll add a whitelist for the false positives and then we can integrate this bad boy into our CI!

troessner:
On hold until we clarified: https://github.com/seattlerb/debride/issues/23

mvz:
Since we have a rather good test coverage, why not use that instead to find dead code? In Ruby, I believe dynamic analysis tends to trump static analysis.

troessner:
>>  In Ruby, I believe dynamic analysis tends to trump static analysis.

says the guy working on a *static* code analysis tool :laughing:  

>> Since we have a rather good test coverage, why not use that instead to find dead code? 

I don't follow - how do you detect the absence of something via test coverage? If a public method is never used, how do you test that?

Looking at (this)[https://github.com/seattlerb/debride/issues/23] it seems like my whitelisting doesn't work out properly. I'll close this one and cherrypick just the unused method changes into a separate PR

mvz:
> says the guy working on a static code analysis tool :laughing:

Ha ha! True. I'm still dreaming of full call stack analysis based on test runs.

> I don't follow - how do you detect the absence of something via test coverage? If a public method is never used, how do you test that?

SimpleCov can tell you which parts of your source code are never run. If you test a public method, then it will have coverage, of course, so it won't catch that. But I wonder if debride would find that? Maybe I misunderstand what it does.

troessner:
Closing this one in favor of #747

PR #747, opened 2015-10-21
Remove unused code.

troessner:
Coming from #745

### DIFF
--- a/spec/reek/cli/options_spec.rb
+++ b/spec/reek/cli/options_spec.rb
@@ -24,8 +24,7 @@ RSpec.describe Reek::CLI::Options do
 
   describe 'parse' do
     it 'raises on invalid argument in ARGV' do
-      options = described_class.new
-      options.argv = ['-z']
+      options = described_class.new ['-z']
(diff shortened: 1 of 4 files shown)


spec/reek/cli/options_spec.rb
```diff
@@ -25,7 +25,7 @@
   describe 'parse' do
     it 'raises on invalid argument in ARGV' do
       options = described_class.new
-      options.argv = ['-z']
+      options.instance_variable_set :@argv, ['-z']
```
mvz: Can we solve this some other way? I can look into options later ...

troessner: >> Can we solve this some other way? I can look into options later ...

You don't like instance_variable_set? The madness!

chastell: Why can’t we just pass this to `described_class.new` above?

troessner: Ah yeah OR that . :laughing: 
Updated!

chastell:
:raised_hands:

PR #754, opened 2015-10-28
Fix Ataru travis failures.

troessner:


troessner:
It seems like rvm, which is used by travis, does something else than rbenv which is why it didn't fail for me locally.

troessner:
@mvz your comment is gone because (i think) you commented on the commit itself (which is gone after squashing), not in the pull request. I solved the problem differently, please re-review ;)

chastell:
Gaah, a top-level, not-hidden file. But let’s go with it for now. :)

PR #759, opened 2015-11-05
Explain that smell messages are not part of our API.

troessner:
Hopefully this will save us some trouble in the future...

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -50,6 +50,33 @@ string -- 5 warnings:
 
 Note that `Reek::Examiner.new` can take `source` as `String`, `Pathname`, `File` or `IO`.
 
+## API stability
+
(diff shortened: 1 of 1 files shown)


mvz:
Nice, that's a good explanation. One possible addition: You could also mention the structured formats (YAML, JSON) as alternatives if people need to access reeks results in a more detailed way.

troessner:
>> Nice, that's a good explanation. One possible addition: You could also mention the structured formats (YAML, JSON) as alternatives if people need to access reeks results in a more detailed way.

Good idea, I added a paragraph at the end of my changes.

troessner:
Anybody merging this bad boy?

chastell:
:+1: merging!

PR #760, opened 2015-11-05
Refactor SmellRepository and friends.

troessner:


lib/reek/smells/smell_detector.rb
```diff
@@ -15,6 +15,7 @@ module Smells
     # :reek:TooManyMethods: { max_methods: 19 }
     # :reek:TooManyInstanceVariables: { max_instance_variables: 5 }
     class SmellDetector
+      private_attr_reader :config
```
troessner: Let's put calls to class macro's always at the top. I find it highly confusing that they are sometimes scattered across the file.

lib/reek/smells/smell_detector.rb
```diff
@@ -60,41 +38,20 @@ def smell_type
         self.class.smell_type
       end
 
-      class << self
```
troessner: We opened up "self" for the second time here which wasn't only unnecessary but also confusing and bad for readability.

lib/reek/smells/smell_detector.rb
```diff
@@ -60,41 +38,20 @@ def smell_type
...
-      end
-
       def register(hooks)
         return unless config.enabled?
         self.class.contexts.each { |ctx| hooks[ctx] << self }
       end
 
-      def examine(context)
-        return unless enabled_for? context
+      def run_for(context)
```
troessner: Changed the naming. I think "examine" was a pretty bad and confusing name since the detectors itself were using "examine_context" and the difference was unclear.

lib/reek/smells/smell_detector.rb
```diff
@@ -60,41 +38,20 @@ def smell_type
...
         return if exception?(context)
 
-        sm = examine_context(context)
-        self.smells_found += sm
+        self.smells_found += examine_context(context)
       end
 
-      def report_on(report)
-        smells_found.each { |smell| smell.report_on(report) }
+      def report_on(collector)
```
troessner: Again, bad naming. We should use "collector" throughout the application.

lib/reek/smells/smell_detector.rb
```diff
@@ -103,7 +60,6 @@ def exception?(context)
 
       protected
 
-      # NOTE: Needs to be protected so += works for Ruby < 2.2
```
troessner: AFAIS this comment is not correct.

lib/reek/smells/smell_detector.rb
```diff
@@ -120,20 +76,45 @@ def config_for(ctx)
...
-        context = options.fetch(:context)
-        exp = context.exp
-        ctx_source = exp.loc.expression.source_buffer.name
-        SmellWarning.new(self,
-                         source: ctx_source,
-                         context: context.full_name,
-                         lines: options.fetch(:lines),
-                         message: options.fetch(:message),
-                         parameters: options.fetch(:parameters, {}))
+        SmellWarning.build self, options
```
troessner: This logic didn't here. Building smell warnings should be the responsibility of ....SmellWarning ;)

lib/reek/smells/smell_detector.rb
```diff
@@ -120,20 +76,45 @@ def config_for(ctx)
...
       end
 
-      private_attr_reader :config
+      public
+
+      class << self
```
troessner: I put all class methods at the end of the file, makes it easier on the eyes.

lib/reek/smells/smell_detector.rb
```diff
@@ -24,32 +25,9 @@ class SmellDetector
...
-          subclasses << subclass
-        end
-
-        def descendants
-          subclasses
-        end
-
-        private
-
-        def subclasses
```
troessner: Hard to see in the diff, but I got completely rid of `subclasses`. Now we just use `descendants` everywhere.

lib/reek/smells/smell_repository.rb
```diff
@@ -23,37 +25,33 @@ def initialize(smell_types: self.class.smell_types,
...
         @smell_types   = smell_types
+
+        @detectors     = smell_types.map do |klass|
+          { klass => klass.new(configuration_for(klass)) }
+        end.reduce({}, :merge)
       end
 
-      def report_on(listener)
-        detectors.each_value { |detector| detector.report_on(listener) }
+      def report_on(collector)
```
troessner: See above, only that `listener` is even more confusing that `report` was.

lib/reek/smells/smell_repository.rb
```diff
@@ -23,37 +25,32 @@ def initialize(smell_types: self.class.smell_types,
                      configuration: {})
         @configuration = configuration
         @smell_types   = smell_types
+        @detectors     = smell_types.map do |klass|
```
troessner: I'm sorry to say this, but sometimes we are really overusing the memoization "||=" pattern in `Reek`.
This kind of initialization belongs in the ..... initializer, and not in a method IMHO.

lib/reek/smells/smell_repository.rb
```diff
@@ -23,37 +25,32 @@ def initialize(smell_types: self.class.smell_types,
...
+        configuration.fetch klass, {}
       end
 
-      # TODO: Make a method smell_detectors_for(scope)
-      def smell_listeners
-        @smell_listeners ||= Hash.new { |hash, key| hash[key] = [] }.tap do |listeners|
-          detectors.each_value { |detector| detector.register(listeners) }
+      def smell_detectors_for(type)
+        @smell_detectors_for ||= Hash.new { |hash, key| hash[key] = [] }.tap do |hooks|
+          detectors.each_value { |detector| detector.register(hooks) }
```
troessner: "hooks" is more fitting than "listener".

lib/reek/tree_walker.rb
```diff
@@ -25,7 +25,7 @@ def initialize(smell_repository, exp)
     end
 
     def walk
-      result.each do |element|
+      context_tree.each do |element|
```
troessner: I guess `result` is the most annoying variable name **ever** ;)

troessner:
Ready for review!

lib/reek/smells/smell_detector.rb
mvz: I don't think I agree. The contents of options is specific to the context of the smell detector.

With your change, we have two constructors: SmellWarning.build and SmellWarning.new, with similar but different parameters. I put this logic specifically outside of SmellWarning so that class doesn't need to know about the internals of Context.

lib/reek/smells/smell_detector.rb
mvz: It doesn't need to be protected, or it needs to be protected but for a different reason?

lib/reek/smells/smell_detector.rb
mvz: :+1:

lib/reek/smells/smell_repository.rb
mvz: I find it confusing either way. What it *is* is a map from context to an array of smell detectors. Let's name it accordingly.

Also, we only ever use the values of detectors, so let's make it an Array instead of a Hash.

Finally, perhaps we can remove the indirection by moving the logic of SmellDetector#register her. It is not very complex.

lib/reek/smells/smell_detector.rb
mvz: But `run_for` is more vague. Perhaps we can rename the other method instead? Or rename this to `examine_if_relevant` or something? I'm not sure yet, but `run_for` is too vague I think.

Also, we should provide a default implementation of `examine_context` that raises.

lib/reek/smells/smell_detector.rb
mvz: Yes, that is better.

lib/reek/smells/smell_repository.rb
mvz: :+1:

spec/reek/smells/control_parameter_spec.rb
```diff
@@ -270,7 +270,7 @@ def things(arg)
         end
       EOS
       ctx = Reek::Context::MethodContext.new(nil, Reek::Source::SourceCode.from(src).syntax_tree)
-      smells = detector.examine(ctx)
+      smells = detector.run_for(ctx)
```
mvz: Now that I look at this, it seems rather odd that `#run_for` returns detected smells, but we also call `#report_on` to have the smells passed to the reporter. Perhaps removing that indirection will be something to do in a next refactoring.

troessner: >> Perhaps removing that indirection will be something to do in a next refactoring.

Absolutely. I have this on my todo list already, but I wanted to keep this PR manageable :)

lib/reek/smells/smell_detector.rb
troessner: >> Or rename this to examine_if_relevant or something?

That' s even more confusing for me :)

Hmmm, how about renaming "run_for" to "examine" and "examine_context" to "inspect"?

lib/reek/smells/smell_detector.rb
troessner: That was my bad, it was indeed needed. However I solved it differently so we still get rid of the comment, please see my second commit.

lib/reek/smells/smell_detector.rb
troessner: Ok, good points. I reverted this change, see my second commit.

lib/reek/smells/smell_repository.rb
troessner: Done, done and done. Please see my second commit :)

troessner:
Addressed all your comments @mvz !
Probably a little hard to see due to github's shitty comment system, but you can just check out my second commit.

lib/reek/smells/smell_repository.rb
```diff
@@ -23,37 +25,40 @@ def initialize(smell_types: self.class.smell_types,
...
+      def configuration_for(klass)
+        configuration.fetch klass, {}
       end
 
-      # TODO: Make a method smell_detectors_for(scope)
-      def smell_listeners
-        @smell_listeners ||= Hash.new { |hash, key| hash[key] = [] }.tap do |listeners|
-          detectors.each_value { |detector| detector.register(listeners) }
+      # :reek:NestedIterators: { max_allowed_nesting: 2 }
+      # :reek:TooManyStatements: { max_statements: 6 }
```
troessner: I know. I know. Two warnings. However, I find the code below pretty readable and straight forward so I'd vote for leaving it like it is.

lib/reek/smells/smell_repository.rb
```diff
@@ -23,38 +25,34 @@ def initialize(smell_types: self.class.smell_types,
...
+      def configuration_for(klass)
+        configuration.fetch klass, {}
       end
 
-      # TODO: Make a method smell_detectors_for(scope)
-      def smell_listeners
-        @smell_listeners ||= Hash.new { |hash, key| hash[key] = [] }.tap do |listeners|
-          detectors.each_value { |detector| detector.register(listeners) }
+      def smell_detectors_for(type)
+        enabled_detectors.select do |detector|
```
troessner: Actually I realized we can get rid of the hash alltogether :tada:

troessner:
I guess the only thing that is left to discuss is how to rename 

SmellDetector#examine and #examine_context 

I'd vote for:

SmellDetector#examine -> examine (so leaving it like it is)
and 
SmellDetector#examine_context -> inspect OR detect

WDYT?

mvz:
> SmellDetector#examine -> examine (so leaving it like it is)
and 
SmellDetector#examine_context -> inspect OR detect

Yes, and I prefer `inspect`.

mvz:
Looking good! Let's get the method names sorted and then this can be merged.

troessner:
>> Looking good! Let's get the method names sorted and then this can be merged.

Done! I kept it as two separate commits though for convenience.
LGTM?

mvz:
Done!

PR #773, opened 2015-11-27
Reverse merge default directives into directory directives.

### LINKED ISSUE #772: TooManyStatements top-level config ignored
I have this in my config.reek

```yaml
TooManyStatements:
  enabled: true
  max_statements: 15
```

But when I run `bundle exec reek` I get the following output

```
...
[9]:CartsController#destroy has approx 8 statements (TooManyStatements)

### PR DESCRIPTION
**Pull request by troessner**
Fixes #772

### DIFF
--- a/lib/reek/configuration/app_configuration.rb
+++ b/lib/reek/configuration/app_configuration.rb
@@ -90,10 +90,11 @@ module Reek
       #
       # @param source_via [String] - the source of the code inspected
       #
-      # @return [Hash] the directory directive for the source or, if there is
-      # none, the default directive
+      # @return [Hash] the directory directive for the source with the default directive
+      #                reverse-merged into it.
       def directive_for(source_via)
-        directory_directives.directive_for(source_via) || default_directive
+        hit = directory_directives.directive_for(source_via)
+        hit ? default_directive.merge(hit) : default_directive
       end
 
       def path_excluded?(path)
--- a/README.md
+++ b/README.md
@@ -258,7 +258,29 @@ exclude_paths:
   - lib/rake/legacy_tasks
 ```
 
-Note you do not need a configuration file at all. If you're fine with all the [defaults](defaults.reek) we set you can skip this completely.
+If you have a directory directive for which a default directive exists, the more specific
+one (which is the directory directive) will take precedence.
+
+This configuration for instance:
+
+```yaml
+---
+IrresponsibleModule:
+  enabled: false
+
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by troessner** [comment]
Ready for review @chastell && @mvz !

File: lib/reek/configuration/app_configuration.rb [hunk 1]
```diff
@@ -90,10 +90,11 @@ def self.new(*)
...
       # @param source_via [String] - the source of the code inspected
       #
-      # @return [Hash] the directory directive for the source or, if there is
-      # none, the default directive
+      # @return [Hash] the directory directive for the source with the default directive
+      #                reverse-merged into it.
       def directive_for(source_via)
-        directory_directives.directive_for(source_via) || default_directive
+        hit = directory_directives.directive_for(source_via)
+        hit ? default_directive.merge(hit) : default_directive
```
chastell: Can we make `directory_directives.directive_for` return an empty hash (`{}`) and then just merge that unconditionally?

chastell: …but I’ll be ok if we did that in a follow-up PR, as to keep this one simple.

chastell: Your call, @troessner! :)

troessner: >> Can we make directory_directives.directive_for return an empty hash ({}) and then just merge that unconditionally?

I thought about this as well but it kind of hurts my feelings. 
From a type perspective this seems wrong, very wrong. Think about it: I ask "Give me a directive for this directory". 
If there is none, I'd expect something that signals absence. So, nil. An empty hash doesn't mean anything and would be confusing in a lot of context. We'd do this **only** for convenience and only here. But we would kill semantical sense for that.
WDYT?

chastell: Well, this is only because we have a primitive obsession and our directives are hashes. :)

If this method actually returned a directive for a directory, then there could be a null directive for a directory that has no directives. But it doesn’t return a directive for a directory, it returns a hash (or a nil).

Refactoring this to actually use directory directives is definitely out of scope of this PR (and I’m not saying we should do this; hashes served us good enough so far) – and if we are to stick to hashes then I agree with you; so, merging!

jfelchner:
@troessner this was the source of 30 minutes of banging my head against a wall.  Thanks so much for doing this. :+1:

troessner:
Your welcome @jfelchner ;)

Issue #775, opened 2015-11-30
Object oriented clusterfuck pattern in Reek


troessner:
We have done a **lot** of refactorings in Reek over the last couple of months and years.

One antipattern though is still used heavily in our code base. I call it "Object oriented clusterfuck pattern" in lack of a better name and it refers to doing **everything** by setting state implicitly in X to be used in Y.

Sounds vage? Ok, let me give you an example by looking at our [Examiner](https://github.com/troessner/reek/blob/master/lib/reek/examiner.rb#L83):

- We create a SmellRepository
- Get a syntax tree
- pass both to the initializer of the TreeWalker and then call "walk" on it. But that doesnt return anything. That's the first WTF.
- That  SmellRepository is then heavily worked on the TreeWalker. That's the second big WTF.
- This SmellRepository then reports to an collector

A creates B and passes B to C. C then works on B so that later A can make B do D.
That's pretty bad spaghetti code. 

Unfortunately there are more instances where we do this. Let's kill this deamon with fire everywhere.
When A creates B then B should create C and pass the results back explicitly.

Or in other words: Method should return values and not operate on foreign entities and with a gazillion of sideeffects.

What do you guys think?

chastell:
Very much yes! It also really hinders inferring how Reek actually works (in detail).

mvz:
:+1:, and :+1: for the name 'Object Oriented Clusterfuck Pattern'.

troessner:
Cleaning up our issues: Was just for discussion and since we all agree we can close this one.

PR #776, opened 2015-12-01
Spec TreeWalker properly.

troessner:
Ready for review!

This is a prerequisite for me in order to start refactoring TreeWalker and CodeContexts to address #775 and quite a few other issues we're having there (obscurity being the biggest one).
The diff is messy by nature unfortunately, probably makes more sense to look at the file without a diff.

What I did:

1.) Introduced specs for #initialize (un-speced before)
2.) Restructured, simplified and cleaned up our existing "statement counting" specs
3.) Introduced specs for #walk (un-speced before)

I'd appreciate a quick review @chastell && @mvz since this is kind of blocking me from further refactoring tasks.

### DIFF
--- a/lib/reek/tree_walker.rb
+++ b/lib/reek/tree_walker.rb
@@ -22,6 +22,7 @@ module Reek
       @smell_repository = smell_repository
       @exp = exp
       @element = Context::RootContext.new(exp)
+      @context_tree = process(exp)
     end
 
     def walk
@@ -33,7 +34,7 @@ module Reek
     private
 
     private_attr_accessor :element
-    private_attr_reader :exp, :smell_repository
+    private_attr_reader :exp, :smell_repository, :context_tree
 
     # Processes the given AST, memoizes it and returns a tree of nested
     # contexts.
@@ -53,10 +54,6 @@ module Reek
     # RootContext -> children: 1 ModuleContext -> children: 1 MethodContext
     #
     # @return [Reek::Context::RootContext] tree of nested contexts
-    def context_tree
-      @context_tree ||= process(exp)
-    end
-
     def process(exp)
       context_processor = "process_#{exp.type}"
       if context_processor_exists?(context_processor)
--- a/spec/reek/tree_walker_spec.rb
+++ b/spec/reek/tree_walker_spec.rb
@@ -1,237 +1,238 @@
 require_relative '../spec_helper'
 require_lib 'reek/tree_walker'
-require_lib 'reek/source/source_code'
 
-# Dummy repository to inject into TreeWalker in order to count statements in
-# all contexts.
-class TestSmellRepository
-  attr_accessor :num_statements
-  def examine(context)
-    self.num_statements = context.num_statements
-  end
-end
+RSpec.describe Reek::TreeWalker do
+  describe '#initialize' do
+    describe 'the structure of the context_tree' do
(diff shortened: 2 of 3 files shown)


lib/reek/tree_walker.rb
```diff
@@ -22,6 +22,7 @@ def initialize(smell_repository, exp)
       @smell_repository = smell_repository
       @exp = exp
       @element = Context::RootContext.new(exp)
+      @context_tree = process(exp)
```
troessner: Ok, I smuggled a minor refactoring in there, so sue me ;)

spec/reek/tree_walker_spec.rb
```diff
@@ -1,237 +1,238 @@
 require_relative '../spec_helper'
 require_lib 'reek/tree_walker'
-require_lib 'reek/source/source_code'
-
-# Dummy repository to inject into TreeWalker in order to count statements in
-# all contexts.
-class TestSmellRepository
```
troessner: That was the worst part about our current tests.

spec/reek/tree_walker_spec.rb
```diff
@@ -1,237 +1,238 @@
...
+        expect(number_of_statements_for(code)).to eq(1)
+      end
     end
+  end
 
-    it 'counts 2 statement in an iterator' do
-      method = process_method('def one() fred.each do; callee(); end; end')
-      expect(method.num_statements).to eq(2)
-    end
+  describe '#walk' do
```
troessner: The specs here are not the best ones. A lot of "send" and reaching deep down into objects. I know.
Let me address #775 and then we can revisit this.
As it is right now our code is so full of side effects that are specs naturally reflect that unfortunately.

spec/reek/tree_walker_spec.rb
```diff
@@ -1,237 +1,238 @@
...
-      klass = process_method(source)
-      expect(klass.num_statements).to eq(0)
+      end
     end
+  end
 
-    it 'does not count empty conditional expression' do
-      method = process_method('def one() if val == 4; ; end; end')
-      expect(method.num_statements).to eq(0)
+  describe 'statement counting' do
```
troessner: I more or less left those specs unchanged. I just simplified them in regards to their test set up and also removed a couple of more or less duplicates.

lib/reek/tree_walker.rb
```diff
@@ -53,10 +54,6 @@ def walk
     # RootContext -> children: 1 ModuleContext -> children: 1 MethodContext
     #
     # @return [Reek::Context::RootContext] tree of nested contexts
-    def context_tree
-      @context_tree ||= process(exp)
```
troessner: I said it before and I'll say it again: We are heavily overusing this pattern in Reek and so it was here.

mvz: I may be mostly to blame for this overuse, as I tend to favour memoization over initialization in the constructor. In this case I don't see a problem moving it since generally you will want to call `#walk` on the TreeWalker anyway.

spec/reek/tree_walker_spec.rb
mvz: Agreed on revisiting this later. See also the SMELL and TODO comments at the top of TreeWalker. Ideally, we can expose `#process` as a public method and have TreeWalker just produce context trees that we can walk elsewhere.

mvz:
Nice!

PR #777, opened 2015-12-02
Let TreeWalker return a SmellRepository.

troessner:
First part of addressing #775

### DIFF
--- a/lib/reek/tree_walker.rb
+++ b/lib/reek/tree_walker.rb
@@ -18,23 +18,35 @@ module Reek
   #
   # :reek:TooManyMethods: { max_methods: 29 }
   class TreeWalker
-    def initialize(smell_repository, exp)
-      @smell_repository = smell_repository
-      @exp = exp
+    private_attr_accessor :element
+    private_attr_reader :exp, :context_tree
+
+    def initialize(syntax_tree)
+      @exp = syntax_tree
       @element = Context::RootContext.new(exp)
       @context_tree = process(exp)
     end
 
-    def walk
-      context_tree.each do |element|
-        smell_repository.examine(element)
+    # Walks the syntax tree, runs the configured SmellDetectors and returns
+    # a SmellRepository that contains the reported SmellWarnings.
+    #
+    # @param smell_types [Array<Reek::Smells::SmellDetector>] The smell types we want
+    #        to apply when walking the tree.
+    # @param configuration [Configuration::AppConfiguration] The corresponding configuration.
+
+    #
+    # @return [Reek::Smells::SmellRepository] SmellRepository that contains all used
+    #         SmellDetectors including all reported SmellWarnings.
+    def walk(smell_types: raise, configuration: raise)
+      Smells::SmellRepository.new(smell_types: smell_types,
+                                  configuration: configuration).tap do |smell_repository|
+        context_tree.each do |element|
+          smell_repository.examine(element)
+        end
       end
     end
 
     private
-
-    private_attr_accessor :element
-    private_attr_reader :exp, :smell_repository, :context_tree
 
     # Processes the given AST, memoizes it and returns a tree of nested
     # contexts.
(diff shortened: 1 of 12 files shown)


lib/reek/tree_walker.rb
```diff
@@ -18,24 +18,35 @@ module Reek
   #
   # :reek:TooManyMethods: { max_methods: 29 }
   class TreeWalker
-    def initialize(smell_repository, exp)
-      @smell_repository = smell_repository
-      @exp = exp
+    private_attr_accessor :element
```
troessner: I will continue doing that until I slayed every last attribute declaration that is NOT at the top of the file ;)

lib/reek/tree_walker.rb
```diff
@@ -18,24 +18,35 @@ module Reek
...
+    # Walks the syntax tree, runs the configured SmellDetectors and returns
+    # a SmellRepository that contains the reported SmellWarnings.
+    #
+    # @param smell_types [Array] The smell types we want to apply when walking the tree.
+    # @param configuration [Configuration::AppConfiguration] The corresponding configuration.
+
+    #
+    # @return [Reek::Smells::SmellRepository] SmellRepository that contains all used
+    #         SmellDetectors including all reported SmellWarnings.
+    def walk(smell_types: raise, configuration: raise)
```
troessner: Yes, I did change the signature of "walk" as well. I always hated those parameter-less methods, feels like 1965.
I believe this change is a step forward since it allows us for more flexible uses of the TreeWalker in the future.
E.g. Walking the same tree but using different configurations and then comparing the results and doing other awesome stuff like this ;)
And it just feels more fluent to me.

troessner:
Same here:
I'd appreciate a speedy review @mvz && @chastell since this is part of my ongoing for refactoring TreeWalker, his family, his friends and his colleagues.

lib/reek/tree_walker.rb
```diff
@@ -18,24 +18,36 @@ module Reek
...
+    # @param smell_types [Array<Reek::Smells::SmellDetector>] The smell types we want
+    #        to apply when walking the tree.
+    # @param configuration [Configuration::AppConfiguration] The corresponding configuration.
+
+    #
+    # @return [Reek::Smells::SmellRepository] SmellRepository that contains all used
+    #         SmellDetectors including all reported SmellWarnings.
+    def walk(smell_types: raise, configuration: raise)
+      Smells::SmellRepository.new(smell_types: smell_types,
+                                  configuration: configuration).tap do |smell_repository|
```
mvz: Why not pass in the SmellRepository itself instead and avoid this extra fixed dependency?

troessner: Because then we're back to "A creates B, passes that to C, C magically works on this" with the only difference that there is an explicit return which would make it even more confusing in my mind.

Conceptually it also makes sense that a TreeWalker not only knows about it, but also creates it - after all, the TreeWalker is the one working on it.

mvz: Yes, I see your point about the magic workings.

Perhaps for now we can inline the entire `#walk` method into `Examiner#run`? I realize this sort of makes the name `TreeWalker` meaningless, but we can rename it to match the actual function it then thas, which is setting up the contexts.

This will then allow refactoring the `smell_repository.examine(...); smell_repository.report_on(...)` combination, which is another instance of the OOCP.

WDYT?

troessner:
>>

That's an awesome idea! Please check out my second commit: https://github.com/troessner/reek/commit/6554563914a3aa6dd822474a5351ac4023647163
Moving this back to wip now since I need to fix the last Examiner spec (which I can't now since I need to run).

WDYT @mvz ?

mvz:
> WDYT about our new and shiny ContextBuilder @mvz ?

:+1:

troessner:
>> Moving this back to wip now since I need to fix the last Examiner spec (which I can't now since I need to run).

Fixed!

I left it as two commits just this time since they very different from each other ;)

Ready for final review!

mvz:
A couple of nits, then LGTM!

troessner:
>> A couple of nits, then LGTM!

Adressed all of your comments, merge? :)

mvz:
RuboCop :cry:

troessner:
Fixed!

PR #778, opened 2015-12-02
Refactor ObjectRefs.

troessner:
I'd appreciate a speedy review @mvz && @chastell since this is part of my ongoing for refactoring TreeWalker, his family, his friends and his colleagues.

### DIFF
--- a/lib/reek/ast/object_refs.rb
+++ b/lib/reek/ast/object_refs.rb
@@ -3,23 +3,46 @@ require 'private_attr/everywhere'
 module Reek
   # Represents functionality related to an Abstract Syntax Tree.
   module AST
-    # Responsible for holding one specific object reference.
-    ObjectRef = Struct.new(:name, :line)
     #
-    # Manages and counts the references out of a method to other objects.
+    # ObjectRefs is used in CodeContexts.
+    # It manages and counts the references out of a method to other objects and to `self`.
+    #
+    # E.g. this code:
+    #   def foo(thing)
+    #     bar.call_me
+    #     bar.maybe(thing.wat)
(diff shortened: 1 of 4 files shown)


lib/reek/ast/object_refs.rb
```diff
@@ -3,25 +3,48 @@
 module Reek
   # Represents functionality related to an Abstract Syntax Tree.
   module AST
-    # Responsible for holding one specific object reference.
-    ObjectRef = Struct.new(:name, :line)
```
troessner: One constant less in our name system! Everybody wins!
Apart from that, this thing was super confusing since the `name` was completely redundant.

lib/reek/ast/object_refs.rb
```diff
@@ -3,25 +3,48 @@
...
       end
 
+      # Records the references a given method in a CodeContext has including
+      # `self` (see the example at the beginning of this file).
+      #
+      # @param name [Symbol] The name of the object that the method references or `self`.
+      # @param line [Int] The line number where this reference occurs.
+      #
+      # @return [Int|nil] The line number that was added (which might be nil).
+      def record_reference_to(name: raise, line: nil)
```
mvz: With the change to named arguments, we can drop the `_to` suffix I think.

lib/reek/ast/object_refs.rb
mvz: :+1:

mvz:
One small nit, otherwise OK for me.

troessner:
>> With the change to named arguments, we can drop the _to suffix I think.

Good point, updated!

PR #781, opened 2015-12-05
Refactor code context and friends

troessner:
Ready for review as 2 separate commits!
But depending on #777 - so I'd appreciate a review there first so I can rebase this PR against an updated master.

A speedy review would be awesome since this this one and #777 are blocking me in regards to the next goal: Refactoring the ContextBuilder.


.rubocop.yml
```diff
@@ -59,3 +59,7 @@ Style/Documentation:
     - 'lib/reek/ast/sexp_extensions/send.rb'
     - 'lib/reek/ast/sexp_extensions/super.rb'
     - 'lib/reek/ast/sexp_extensions/variables.rb'
+
+Style/AccessorMethodName:
+  Exclude:
+    - 'lib/reek/context/visibility_tracker.rb'
```
troessner: I believe the exception here is legit.

lib/reek/context/singleton_method_context.rb
```diff
@@ -6,9 +6,6 @@ module Context
     # A context wrapper for any singleton method definition found in a syntax tree.
     #
     class SingletonMethodContext < MethodContext
-      def envious_receivers
```
troessner: This wasn't used anywhere except for in its own spec.

troessner:
And with #777 merged, this one is ready for review!

lib/reek/context/code_context.rb
```diff
@@ -12,10 +16,13 @@ module Context
     # :reek:TooManyMethods: { max_methods: 19 }
     # :reek:TooManyInstanceVariables: { max_instance_variables: 8 }
     class CodeContext
-      attr_reader :exp
-      attr_reader :num_statements
-      attr_reader :children
-      attr_reader :visibility
+      extend Forwardable
+      def_delegators :exp, :name, :type
+      def_delegators :visibility_tracker, :visibility, :visibility=, :non_public_visibility?
```
chastell: `def_delegators`’ first argument always trips me up, so I have a slight preference for the more explicit `delegate` syntax:

```ruby
      delegate %i(visibility visibility= non_public_visibility?) => :visibility_tracker
```

WDYT?

lib/reek/context/visibility_tracker.rb
```diff
@@ -0,0 +1,54 @@
...
+      #
+      def track_visibility(children, visibility, names)
+        if names.any?
+          children.each do |child|
+            child.visibility = visibility if names.include?(child.name)
+          end
+        else
+          self.tracked_visibility = visibility
+        end
+      end
```
chastell: Hm, this method is called exactly once (in `CodeContext#track_visibility`), and the caller knows whether `names` is empty or not (and could just as well filter `children` based on `names`), so this smells to me of control parameter.

This also means that the second half of this method (when `names` is not `#empty?`) ignores ⅔ of the method’s arguments and uses only the middle one. I’d at least make them keyword arguments (sadly, we can’t make them required without the `foo: raise` hack while we support Ruby 2.0).

chastell:
(Finally coming back after being sick for two weeks…)

I like the idea of factoring out the `VisibilityTracker`, but I have a couple nitpicks. :)

lib/reek/context/code_context.rb
troessner: I like the idea a lot, updated.

lib/reek/context/visibility_tracker.rb
troessner: >> Hm, this method is called exactly once (in CodeContext#track_visibility), and the caller knows whether names is empty or not (and could just as well filter children based on names on the caller side), so this smells to me of control parameter.
This also means that the second half of this method (when names is #empty?) ignores ⅔ of the method’s arguments and uses only the middle one. 

I agree that this method is not the best one! Let's improve on that in subsequent pull requests, I'm rampaging over CodeContexts and friends right now heavily anyways!

>>  I’d at least make them keyword arguments (sadly, we can’t make them required without the foo: raise hack while we support Ruby 2.0).

Updated.

troessner:

chastell:
:+1: merging!

PR #783, opened 2015-12-08
Move statement counting into a proper object.

troessner:

### DIFF
--- a/lib/reek/context/statement_counter.rb
+++ b/lib/reek/context/statement_counter.rb
@@ -1,0 +1,32 @@
+require_relative '../ast/node'
+require 'private_attr/everywhere'
+
+module Reek
+  module Context
+    # Responsible for counting the statements in a `CodeContext`.
+    class StatementCounter
+      attr_reader :value
+      private_attr_writer :value
+
+      def initialize
+        @value = 0
+      end
+
+      def increase_by(sexp)
+        return unless sexp
+        case sexp
+        when Reek::AST::Node
+          self.value = value + 1
+        when Array
+          self.value = value + sexp.length
+        else
+          raise ArgumentError, "Invalid type #{sexp} given"
+        end
+      end
+
+      def decrease_by(number)
+        self.value = value - number
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


lib/reek/context/statement_counter.rb
```diff
@@ -0,0 +1,33 @@
...
+    # Responsible for counting the statements in a `CodeContext`.
+    class StatementCounter
+      attr_reader :value
+      private_attr_writer :value
+
+      def initialize
+        @value = 0
+      end
+
+      # :reek:DuplicateMethodCall: { max_calls: 2 }
```
troessner: Technically it's correct, but we still shouldn't report DuplicateMethodCall here. Something we should fix in the future.

lib/reek/context/statement_counter.rb
```diff
@@ -0,0 +1,32 @@
...
+        return unless sexp
+        case sexp
+        when Reek::AST::Node
+          self.value = value + 1
+        when Array
+          self.value = value + sexp.length
+        else
+          raise ArgumentError, "Invalid type #{sexp} given"
+        end
+      end
```
chastell: I’d probably go with the shorthand notation (like `self.value += sexp.length` and `self.value -= number` below.

Also, `rspec spec/quality/reek_source_spec.rb` ran locally says this method smells of DuplicateMethodCall (`self.value`); doesn’t our CI run that check?

troessner: With that

>> I’d probably go with the shorthand notation (like self.value += sexp.length and self.value -= number below.

I get:

>>        NoMethodError:
         private method `value=' called for #<Reek::Context::StatementCounter:0x007fed2c80b468 @value=0>
       # ./lib/reek/context/statement_counter.rb:21:in `increase_by'

under ruby 2.1 and 2.0.

>> Also, rspec spec/quality/reek_source_spec.rb ran locally says this method smells of DuplicateMethodCall (self.value); doesn’t our CI run that check?

It does! And I don't get this error locally as well. Are you sure you've checked out the latest version of the code?

chastell: Aaah, right, I actually assumed that private `self.value` would blow up, but then tried it and was surprised it works – turns out it works only on Ruby 2.2.

As for the smell, it seems to show up only once the code is changed to `+=` and `-=`. Strange, but irrelevant for this one – so I’ll merge it. :)

PR #785, opened 2015-12-12
Refactor context building.

troessner:
Our context building was pretty confusing.
So basically we were using the chain

inside_new_context -> new_context -> push

but we were also using `new_context` standalone (e.g. for AttributeContext). In this case however not for a new context itself, but for the sideeffect of appending a child context.
Additonally we were passing down a block 2 methods deep, something that always smells for me.

This PR cleans this up and separates concerns better which results in better readability.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -63,3 +63,6 @@ Style/Documentation:
 Style/AccessorMethodName:
   Exclude:
     - 'lib/reek/context/visibility_tracker.rb'
+
+Style/ParallelAssignment:
+  Enabled: false
--- a/lib/reek/context/visibility_tracker.rb
+++ b/lib/reek/context/visibility_tracker.rb
@@ -39,7 +39,7 @@ module Reek
         child.visibility = tracked_visibility
       end
 
-      # return [Boolean] If the visibility is public or not.
+      # @return [Boolean] If the visibility is public or not.
       def non_public_visibility?
         visibility != :public
       end
--- a/lib/reek/context_builder.rb
+++ b/lib/reek/context_builder.rb
@@ -102,7 +102,7 @@ module Reek
       if exp.attribute_writer?
         exp.args.each do |arg|
           next unless arg.type == :sym
-          new_context(Context::AttributeContext, arg, exp)
+          append_new_context(Context::AttributeContext, arg, exp)
         end
       end
       element.record_call_to(exp)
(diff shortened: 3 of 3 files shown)


.rubocop.yml
```diff
@@ -63,3 +63,6 @@ Style/Documentation:
 Style/AccessorMethodName:
   Exclude:
     - 'lib/reek/context/visibility_tracker.rb'
+
+Style/ParallelAssignment:
```
troessner: I believe we shouldn't follow this cop.
WDYT?

lib/reek/context/visibility_tracker.rb
```diff
@@ -39,7 +39,7 @@ def set_child_visibility(child)
         child.visibility = tracked_visibility
       end
 
-      # return [Boolean] If the visibility is public or not.
+      # r@eturn [Boolean] If the visibility is public or not.
```
troessner: boyscout rule ;)

lib/reek/context_builder.rb
```diff
@@ -102,7 +102,7 @@ def process_send(exp)
       if exp.attribute_writer?
         exp.args.each do |arg|
           next unless arg.type == :sym
-          new_context(Context::AttributeContext, arg, exp)
+          append_new_context(Context::AttributeContext, arg, exp)
```
troessner: The new name makes it actually clear that we calling this for the side effect, and not for the return value.

.rubocop.yml
chastell: Yup, let’s disable it.

lib/reek/context/visibility_tracker.rb
chastell: Is it some l33t boy scout or can we move the `@` a tad to the left? ;)

chastell:
I like the PR, but let’s turn that `r@eturn` into `@return`. :)

lib/reek/context/visibility_tracker.rb
troessner: Whoopsie! My daughter was crying, so I was pushing the last change in a hurry :laughing:

troessner: Fixed!

troessner:
>> I like the PR, but let’s turn that r@eturn into @return. :)

Done. Merge? ;)

PR #787, opened 2015-12-13
Refactor and document ContextBuilder.

troessner:


lib/reek/context_builder.rb
```diff
@@ -10,10 +10,11 @@ module Reek
...
   # specific node types.
   #
-  # SMELL: This class is responsible for counting statements and for feeding
-  # each context to the smell repository.
+  # TODO: This class is responsible for statements and reference
+  # counting. Ideally `ContextBuilder` would only build up the context tree and leave the
+  # statement and reference counting to the contexts.
   #
-  # :reek:TooManyMethods: { max_methods: 29 }
+  # :reek:TooManyMethods: { max_methods: 26 }
```
troessner: Minus 3 methods! Only 500 more to go!

lib/reek/context_builder.rb
```diff
@@ -22,7 +23,7 @@ class ContextBuilder
     def initialize(syntax_tree)
       @exp = syntax_tree
       @element = Context::RootContext.new(exp)
-      @context_tree = process(exp)
+      @context_tree = build(exp)
```
troessner: The naming here has been annoying me for sooooo long. "process" never really processed anything. It started to "build" the tree ;)

lib/reek/context_builder.rb
```diff
@@ -45,54 +50,88 @@ def initialize(syntax_tree)
...
       element
     end
 
+    # Handles every node for which we have no context_processor.
+    #
+    def process(exp)
+      exp.children.each do |child|
+        build(child) if child.is_a?(AST::Node)
+      end
+    end
```
chastell: Somewhat tempted to suggest going all the way here:

```ruby
def process(exp)
  exp.children.grep(AST::Node).each(&method(:build))
end
```

…but I understand it might be considered too far. ;)

lib/reek/context_builder.rb
```diff
@@ -45,54 +50,88 @@ def initialize(syntax_tree)
...
 
-    def process_args(_) end
-
+    # Handles `send` nodes a.k.a. method calls.
+    #
+    # An input example that would trigger this method would be:
+    #
+    #   call_me()
+    #
+    # Besides for checking if its a visibility modifier or an attribute writer
```
chastell: `s/Besides for checking if its/Besides checking if it's`

lib/reek/context_builder.rb
```diff
@@ -101,93 +140,278 @@ def process_send(exp)
...
+      process(exp)
     end
 
     alias_method :process_kwbegin, :process_begin
 
+    # Handles `if` nodes.
+    #
+    # An input example that would trigger this method would be:
+    #
+    # if(a > 5 && b < 3)
```
chastell: I’d try to keep our examples RuboCop-clean and drop the parentheses here.

lib/reek/context_builder.rb
```diff
@@ -101,93 +140,278 @@ def process_send(exp)
...
       decrease_statement_count
-      process_default(exp)
+      process(exp)
     end
 
+    # Handles `while` and `until` nodes.
+    #
+    # An input example that would trigger this method would be:
+    #
+    # while(x < 5)
```
chastell: Ditto: I’d drop the parentheses.

lib/reek/context_builder.rb
```diff
@@ -101,93 +140,278 @@ def process_send(exp)
...
+    #   end
+    # end
+    #
+    # Counts the `rescue` body as one statement.
+    #
+    # At the end we subtract one statement because the surrounding context was already counted
+    # as one (e.g. via `process_def`).
+    #
+    # `exp.children.first` below refers to the everything that is covered by the `rescue`
+    # node body (which is everything from `begin` to `rescue` in the example above).
```
chastell: Isn’t it everything from `rescue` to `end`?

chastell: Oh, maybe not, maybe that’s a `resbody` – I’d add an explicit note here about this, though.

lib/reek/context_builder.rb
```diff
@@ -101,93 +140,278 @@ def process_send(exp)
...
+    #
+    # An input example that would trigger this method would be:
+    #
+    # foo = 5
+    # case foo
+    #   when (1..100)
+    #     puts 'In between'
+    #   else
+    #     puts 'Not sure what I got here'
+    # end
```
chastell: I’d indent `when` and `else` the same as `case` and `end`, and maybe drop the parentheses around `1..100`.

lib/reek/context_builder.rb
```diff
@@ -101,93 +140,278 @@ def process_send(exp)
...
+    # Handles `when` nodes.
+    #
+    # An input example that would trigger this method would be:
+    #
+    # foo = 5
+    # case foo
+    #   when (1..100)
+    #     puts 'In between'
+    #   else
+    #     puts 'Not sure what I got here'
```
chastell: Same as above.

lib/reek/context_builder.rb
```diff
@@ -101,93 +140,278 @@ def process_send(exp)
...
+    # end
+    #
+    # Counts the `when` body.
```
chastell: Should we add a note whether this is triggered by `if foo then`?

chastell:
Awesome job, especially on all the extra docs. I have a few minor suggestions and then it’s good to go.

lib/reek/context_builder.rb
troessner: I like it!
Additionally, I've come around with the 

>> &method(:..)

syntax, I updated it.

lib/reek/context_builder.rb
troessner: Fixed.

lib/reek/context_builder.rb
troessner: I updated the PR regarding this comment and all other ones.

lib/reek/context_builder.rb
troessner: Updated!

lib/reek/context_builder.rb
troessner: Updated both method comments, should be a lot clearer now!

troessner:
Addressed all comments! Merge?

chastell:
:ok_hand:

PR #791, opened 2015-12-23
Refactor CodeComment.

troessner:
The initial implementation was a little smelly (magic number!), a little obscure (gsub ftw) and a little messy.

### DIFF
--- a/lib/reek/code_comment.rb
+++ b/lib/reek/code_comment.rb
@@ -1,39 +1,45 @@
 require 'yaml'
 require 'private_attr/everywhere'
 
-# NOTE: Work-around for https://github.com/tenderlove/psych/issues/223
-require 'psych.rb' if Object.const_defined?(:Psych)
-
 module Reek
   #
   # A comment header from an abstract syntax tree; found directly above
   # module, class and method definitions.
   #
   class CodeComment
-    CONFIG_REGEX = /:reek:(\w+)(:\s*\{.*?\})?/
-
-    def initialize(text)
-      @config = Hash.new { |hash, key| hash[key] = {} }
-      @text = text.gsub(CONFIG_REGEX) do
-        @config.merge! add_to_config($1, $2)
-        ''
-      end.gsub(/#/, '').gsub(/\n/, '').strip
-    end
+    CONFIGURATION_REGEX            = /:reek:(\w+)(:\s*\{.*?\})?/
+    SANITIZE_REGEX                 = /(#|\n|\s)+/ # Matches '#', newlines and > 1 whitespaces.
+    DISABLE_DETECTOR_CONFIGURATION = ': { enabled: false }'
+    MINIMUM_CONTENT_LENGTH         = 2
 
     attr_reader :config
+    private_attr_reader :original_comment
+
+    #
+    # @param comment [String] - the original comment as found in the source code
+    #   E.g.:
+    #   "\n        # :reek:Duplication: { enabled: false }\n      "
+    #
+    def initialize(comment)
+      @original_comment  = comment
+      @config            = Hash.new { |hash, key| hash[key] = {} }
+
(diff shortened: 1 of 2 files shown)


lib/reek/code_comment.rb
```diff
@@ -1,39 +1,55 @@
 require 'yaml'
 require 'private_attr/everywhere'
 
-# NOTE: Work-around for https://github.com/tenderlove/psych/issues/223
-require 'psych.rb' if Object.const_defined?(:Psych)
```
troessner: Issue is still open but the problem itself is fixed.

lib/reek/code_comment.rb
```diff
@@ -1,39 +1,55 @@
...
+    # @param comment [String] - the original comment.
+    #   E.g.:
+    #   "\n        # :reek:Duplication: { enabled: false }\n      "
+    #
+    def initialize(comment)
+      @original_comment  = comment
+      @config            = Hash.new { |hash, key| hash[key] = {} }
+      @sanitized_comment = sanitize
+
+      @original_comment.scan(CONFIGURATION_REGEX) do |smell_type, configuration|
```
troessner: Gets rid of the obscure "gsub" usage from before.

lib/reek/code_comment.rb
mvz: Yes, it is fixed in latest rubygems.

lib/reek/code_comment.rb
```diff
@@ -1,39 +1,44 @@
...
     private
 
-    # :reek:UtilityFunction
-    def add_to_config(smell, options)
-      options ||= ': { enabled: false }'
-      YAML.load(smell + options)
+    def sanitized_comment
+      @sanitized_comment ||= original_comment.
+        gsub(CONFIGURATION_REGEX, '').
+        gsub(/#|\n|(\s{2,})/, '').
```
mvz: This regexp is a little obscure. What does it do?

spec/reek/code_comment_spec.rb
```diff
@@ -81,5 +81,15 @@
...
     end
+
+    it 'removes the configuration options from the comment' do
+      subject = described_class.new('
+        # Actual
+        # :reek:Duplication: { enabled: false }
+        # :reek:NestedIterators: { enabled: true }
+        # comment
+      ')
+      expect(subject.send(:sanitized_comment)).to eq('Actual comment')
```
mvz: I'm not sure we need to remove the newlines from the sanitized comment.

lib/reek/code_comment.rb
troessner: I cleaned it up a little including a comment, hopefully this makes it clearer.

spec/reek/code_comment_spec.rb
troessner: We don't!
However it makes debugging a lot easier and it's way more convenient to play around with stuff like this in the console.
Let's call it a productivity enhancement and I believe that we can't add enough of those ;)

mvz: :+1:

lib/reek/code_comment.rb
```diff
@@ -1,39 +1,45 @@
...
     private
 
-    # :reek:UtilityFunction
-    def add_to_config(smell, options)
-      options ||= ': { enabled: false }'
-      YAML.load(smell + options)
+    def sanitized_comment
+      @sanitized_comment ||= original_comment.
+        gsub(CONFIGURATION_REGEX, '').
+        gsub(SANITIZE_REGEX, '').
```
mvz: I think this will convert `foo  bar` (with two spaces) to `foobar`.

Perhaps `gsub(/(#|\n|\s)+/, ' ') will to the right thing?

troessner: Good catch, fixed!

troessner:

Issue #792, opened 2015-12-24
Bug in our visibility tracking for non-instance methods


troessner:
#724 uncovered a bug in our visibility tracking in regards to non-instance methods.

The following spec illustrates this bug:

```Ruby
    it 'reports the visibility of class methods properly' do
      src = <<-EOS
        class Car
          private
          def self.drive; end
        end
      EOS

      syntax_tree = Reek::Source::SourceCode.from(src).syntax_tree

      context_tree = Reek::ContextBuilder.new(syntax_tree).send(:context_tree)

      singleton_method = context_tree.children.first.children.first
      expect(singleton_method.visibility).to eq(:public)
    end
```

The output of this:

```
  1) Reek::Context::CodeContext enumerating syntax elements reports the visibility of class methods properly
     Failure/Error: expect(singleton_method.visibility).to eq(:public)

       expected: :public
            got: :private
```

@mvz || @chastell any of you got time to look into this?

mvz:
I may be able to look at it tomorrow, as a fun Christmas activity :smile:.

mvz:
Fixed in #796.

PR #794, opened 2015-12-24
Report unused private instance methods.

### LINKED ISSUE #793: Bug in our context building for non-instance methods
#724 uncovered a bug in our context building in regards to non-instance methods.

The following spec illustrates this bug:

```Ruby
    it "properly recognizes singleton_methods correctly" do
      src = <<-EOS
        class Car
          class << self
            def drive; end
          end

          def self.start; end
        end
      EOS

      syntax_tree = Reek::Source::SourceCode.from(src).syntax_tree

      context_tree = Reek::ContextBuilder.new(syntax_tree).send(:context_tree)

      class_node = context_tree.children.first
      drive_method = class_node.children.first
      start_method = class_node.children.last

      expect(start_method.class).to eq(Reek::Context::SingletonMethodContext)
      expect(drive_method.class).to eq(Reek::Context::SingletonMethodContext)
    end
```

Both methods should be recognized as SingletonMethod, but only the first one is:

```
Failures:

  1) Reek::ContextBuilder#initialize properly recognizes singleton_methods correctly
     Failure/Error: expect(drive_method.class).to eq(Reek::Context::SingletonMethodContext)

       expected: Reek::Context::SingletonMethodContext
            got: Reek::Context::MethodContext

       (compared using ==)
```

I thought we had fix this bug a looooong time ago but apparently not.

troessner:
This is a subset of #724 with the ability to discern between instance and class methods removed.

The "plan" is so far:

* We get this one merged 
* We fix #793 and #792
* We can then return to this detector and make it work on class methods as well.

Review please - most of the code you should be already familiar with.


lib/reek/context_builder.rb
```diff
@@ -74,7 +77,11 @@ def process_module(exp)
       end
     end
 
-    alias_method :process_class, :process_module
+    def process_class(exp)
+      inside_new_context(Context::ClassContext, exp) do
```
mvz: What is the purpose of having a different class here?

lib/reek/smells/unused_private_method.rb
```diff
@@ -0,0 +1,82 @@
+require_relative 'smell_detector'
+require_relative 'smell_warning'
+require_relative '../context/method_context'
+
+module Reek
+  module Smells
+    # Class for storing `hits` which are unused private methods
+    # we found in the given context. `name` and `line` are then used to
+    # construct SmellWarnings.
+    class Hit
```
mvz: Maybe this class should be moved into the smell detector class, since it is only used there?

lib/reek/smells/unused_private_method.rb
```diff
@@ -0,0 +1,82 @@
...
+        defined_private_methods.select do |defined_method|
+          !called_method_names.include?(defined_method.name)
+        end
+      end
+
+      #
+      # @param ctx [Context::ClassContext]
+      # @return [Boolean]
+      #
+      def ignore_method?(ctx, method)
```
mvz: Maybe you can add some specs for the logic in this method?

lib/reek/smells/unused_private_method.rb
```diff
@@ -0,0 +1,82 @@
...
+        ignore_methods = value(EXCLUDE_KEY, ctx, DEFAULT_EXCLUDE_SET)
+        ignore_methods.any? { |ignore_method| method.name[ignore_method] }
```
mvz: I didn't know about this use of `String#[]`. Nifty :smile:.

mvz:
A nice compact pull request due to all the hard refactoring work you did earlier!

Just a few nits, then LGTM!

lib/reek/context_builder.rb
troessner: I was on the make-all-a-context bandwagon but now on second thought after your comment I realized that it doesn't improve our architecture by just adding this by default. I restore the initial

>> alias_method :process_class, :process_module

lib/reek/smells/unused_private_method.rb
troessner: Good idea, updated!

lib/reek/smells/unused_private_method.rb
troessner: Added a spec that tests the overall expected behaviour instead of spec'ing this very method: https://github.com/troessner/reek/pull/794/files#diff-03c7dc164fe6c989d619625ce17dd00cR50

troessner:

PR #795, opened 2015-12-25
Add bin/console.

troessner:
I find little things like this super useful and productivity increasing.

### DIFF
--- a/tasks/console.rake
+++ b/tasks/console.rake
@@ -1,0 +1,5 @@
+desc 'Starts the interactive console'
+task :console do
+  require 'pry'
+  Pry.start
+end


mvz:
Perhaps it makes sense to namespace this in `scripts/` rather than lumping it in `bin/`?

(Actually, Bundler's gem generator has started to put a gem's executables in `exe/`, while putting binstubs in `bin/`, which I find both annoying because it reminds me of Windows, and funny because Rails originally put its scripts in `scripts/`, then moved them to `bin/` as binstubs, causing actual distributable executables (that have nothing to do with Rails) to be moved in turn.)

mvz:
Alternatively, how about a rake task for this?

I'm just posting my thoughts, don't have a preference yet.

troessner:
>> Perhaps it makes sense to namespace this in scripts/ rather than lumping it in bin/?

I guess that's depending on one's history. Personally I find it more intuitive in "bin" :)

>> Alternatively, how about a rake task for this?

Good idea! I changed the pull request accordingly, please re-review.

mvz:
Looks good, but the commit comment does not make sense any more :smile:.

troessner:
>> Looks good, but the commit comment does not make sense any more :smile:.

Yep, and the branch name as well ;)
Opened the successor for this via #797 -> closing this one.

PR #800, opened 2015-12-27
Don't raise on missing directory for directive.

### LINKED ISSUE #799: Directory Directive blows up if the directory doesn't exist
This seems reasonable at first glance, but I would really like to be able to have a common `.reek` file that I use everywhere (Rails, non-Rails, legacy, greenfield) that I can keep up-to-date with a simple script to copy my most recent config version to all my projects.  Unfortunately this issue keeps me from doing that.  For example:

```
'app/helpers':
  FeatureEnvy:
    enabled:                false
  UtilityFunction:
    enabled:                false
```

blows up on non-Rails projects (or Rails projects where the `app/helpers` directory has been removed).

Thoughts?

troessner:
Fixes #799

### DIFF
--- a/features/configuration_files/directory_specific_directives.feature
+++ b/features/configuration_files/directory_specific_directives.feature
@@ -170,22 +170,6 @@ Feature: Directory directives
     2 total warnings
     """
 
-  Scenario: Abort on non-existent directory
-    Given a file named "config.reek" with:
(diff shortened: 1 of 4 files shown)


chastell:
`+1 ­-48`? Merging right away! :heart_eyes:

PR #801, opened 2015-12-28
Parser 2.3 support

chastell:
I want to get my pet projects that would benefit from the squiggly heredoc over to Ruby 2.3 ASAP, and for this we need to make sure Reek supports it.

To the best of my quick checks the current blockers are:

1. (the newest) parser 2.3.0.pre.6 still not supporting the new syntax,
1. (the newest) unparser 0.4.2 depending on parser `'~> 2.2.2'`.

Once the above are resolved I want to make sure Reek at least doesn’t blow up on the new syntax, but ideally addresses those parts of the new syntax it should address. ;)

### DIFF
--- a/features/command_line_interface/stdin.feature
+++ b/features/command_line_interface/stdin.feature
@@ -33,7 +33,7 @@ Feature: Reek reads from $stdin when no files are given
       """
 
   Scenario: syntax error causes the source to be ignored
-    When I pass "def incomplete" to reek
+    When I pass "= invalid syntax =" to reek
     Then it reports a parsing error
     Then it succeeds
     And it reports nothing
--- a/reek.gemspec
+++ b/reek.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.summary = 'Code smell detector for Ruby'
 
   s.add_runtime_dependency 'codeclimate-engine-rb', '~> 0.1.0'
-  s.add_runtime_dependency 'parser',                '~> 2.2.2', '>= 2.2.2.5'
+  s.add_runtime_dependency 'parser',                '~> 2.3'
   s.add_runtime_dependency 'private_attr',          '~> 1.1'
   s.add_runtime_dependency 'rainbow',               '~> 2.0'
 end


chastell:
Oooh, the lovely issue → PR transformation API still works. ;)

troessner:
Aaaaaand parser 2.3 is out! One problem though:

```
  ( $ (master)) bundle update
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/..
Resolving dependencies...
Bundler could not find compatible versions for gem "parser":
  In Gemfile:
    reek (>= 0) ruby depends on
      parser (~> 2.3) ruby

    rubocop (~> 0.36.0) ruby depends on
      parser (< 3.0, >= 2.3.0.0) ruby

    reek (>= 0) ruby depends on
      unparser (~> 0.2.4) ruby depends on
        parser (~> 2.2.2) ruby
```

and that is the latest `unparser` version. Seems like we need to get #815 going first.

troessner:
I'll start with #815 right away.
@chastell do you have time to handle this issue or are you still sick? If yes, I can start to deal with this.
Sorry for being pushy, but I think we should act quick here. Ruby 2.3 is out for a while now and I believe we have a high probability to alienate our users.
Let's get this going.

mvz:
> I'll start with #815 right away.

You're on a roll! Awesome :+1:

chastell:
I’m sick _again_ (as in, third time in a row), but I’ll be on this this weekend. :)

(And I agree we should get it fixed ASAP, I have projects I want to switch to the squiggly heredoc!) ;)

chastell:
This is green now and (patiently!) waits for https://github.com/mbj/unparser/pull/60.

features/command_line_interface/stdin.feature
```diff
@@ -33,7 +33,7 @@ Feature: Reek reads from $stdin when no files are given
       """
 
   Scenario: syntax error causes the source to be ignored
-    When I pass "def incomplete" to reek
+    When I pass "=" to reek
```
troessner: Why this change?

troessner:
>> This is green now and (patiently!) waits for mbj/unparser#60.

Are those really the only changes we need to do? 
I mean

>> Parser::CurrentRuby.parse("user.&profile")

looks vastly different from

>> Parser::CurrentRuby.parse("user.profile")

....

features/command_line_interface/stdin.feature
chastell: Parser 2.3.0.1 [blows up](https://travis-ci.org/troessner/reek/jobs/102731465) on parsing `def incomplete`:

```
lib/parser/diagnostic.rb:121:in `[]=': 1...1 out of range (RangeError)
```

I tried to make Parser raise the expected `Parser::SyntaxError` and `=` was the first thing that worked. I can try coming up with something less crypting that doesn’t make Parser blow up (but raise what we expect).

(We should probably guard Reek users from Parser blowing up, but that’s a task for another PR…)

chastell:
> Are those really the only changes we need to do?

You’re right, these are by far not the only changes we need to do to support the full Ruby 2.3 syntax; this is just the base to actually make Reek being able to depend on Parser 2.3. Still, I’d rather we get this in ASAP (probably by getting #827 in first) so that people can at least upgrade their RuboCops.

I’ll work on supporting what Parser calls on the new syntax ASAP, unless you’d rather I worked on #827 first. :)

troessner:
>> I’ll work on supporting what Parser calls on the new syntax ASAP, unless you’d rather I worked on #827 first. :)

No, please go ahead by all means. I'm afraid removing `unparser` is more effort than I had thought. Right now I'm thinking more about actually fixing this in `unparser`, but let's see).

features/command_line_interface/stdin.feature
mvz: Perhaps this should be reported as a bug in parser?

chastell:
> I'm thinking more about actually fixing this in unparser

I [did fix](https://github.com/mbj/unparser/pull/60) the dependency issue in unparser, but if you read that PR you’ll see that it’s unlikely to be merged soon, and I doubt unparser will support the new Ruby 2.3 syntax anytime soon either.

I think we should push on with #827; we can’t release Reek with a [runtime dependency on a GitHub repo](https://github.com/troessner/reek/commit/d9e0a7f8191b64b6fc54b85ea362207c83474457). Let me know if I can help with #827 – I really want a Reek version that does not block Parser upgrade ASAP (even before we fully support Ruby 2.3), as people currently can’t upgrade to the newest RuboCop if they happen to use Reek as well (and I really don’t want them to stop using Reek).

mvz:
> I think we should push on with #827;

I agree. I can help with that one as well if needed.

troessner:
>> I agree. I can help with that one as well if needed.

That would be highly welcome! #827 is really just a super-early wip branch, nothing more. If you'd like to take over, I'd just close this one. Would you like to tackle this?

mvz:
I can just take over the branch from #827. It seems to at least be heading in the right direction.

troessner:
Ok, sounds great, then I'll just leave it like that ;)

troessner:
Now that #827 is merged lets rebase this one so we can make Reek 2.3 compatible!

troessner:
@chastell let's get this going, you just need to remove [this line](https://github.com/troessner/reek/pull/801/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R5) and then we can merge this one ;)

chastell:
Yes, sorry! This week totally blew up for me. On it!

chastell:
@troessner I limited this PR (and issue) to relaxing our dependency on Parser so that Reek can be installed alonside Parser 2.3. I also made the invalid syntax feature a bit more self-explanatory.

I think we should get this merged and released, and I’ll work on actual Ruby 2.3 support (handling what Parser emits for the new syntax) in a separate PR.

troessner:
>>I limited this PR (and issue) to relaxing our dependency on Parser so that Reek can be installed alonside Parser 2.3

I dont get it, why not 

>> gem 'parser',  ~> '2.3'

right away? Whats the benefit? Parser is fully backwards compatible so i'd rather update now...

chastell:
My reasons for allowing Parser 2.2.2.5 and 2.2.2.6 are:
* Reek _does_ work with these versions and I don’t think it will stop working with them anytime soon,
* if there is another tool that depends on unparser or Parser < 2.3 then it won’t be usable alongside Reek 3.8.4+ (just as Reek 3.8.3 is not installable alongside current RuboCop),
* I’m reluctant to tighten runtime dependencies unless necessary, and I’d say such tightening requires at least a minor version bump.

But if you’re feeling strong about `'~> 2.3'` then I’ll do it right away so we can release ASAP. :)

troessner:
But that way Reek + ruby 2.3 might work for a user depending on randomness. That is, if he has 2.2.2 already installed this doesnt work. On fresh install it would work. Which is really really weird.

>> Reek does work with these versions and I don’t think it will stop working with them anytime soon,

Conjecture. Maybe, maybe not. Donald Knuth is screaming "premature optimization" in my ear as we speak.

>> if there is another tool that depends on unparser or Parser < 2.3 then it won’t be usable alongside Reek 3.8.4+ (just as Reek 3.8.3 is not installable alongside current RuboCop),

He is still screaming. 

And my closing argument: This is not enterprise java beans but ruby. I believe we should rather take swift and bold actions and not live the live of a "make it right for every user on this planet" slave.

Wdyt?

troessner:
Addendum: i see the rubocop problem but do we really want to limit Reek to some other tool, no matter how popular it is?

chastell:
> But that way Reek + Ruby 2.3 might work for a user depending on randomness.

Ok, this is a good point. I bumped this to `'~> 2.3'`.

> I see the RuboCop problem but do we really want to limit Reek to some other tool, no matter how popular it is?

I’m not sure I got my point across: with `'~> 2.3'` Reek is behaving towards other tools like RuboCop 0.36 behaved towards Reek: if Tool X depends on Parser < 2.3 _or_ depends on unparser then it won’t be installable along current version of Reek.

Meanwhile, this built. ;)

troessner:
Merged! Release incoming!

troessner:
>> : if Tool X depends on Parser < 2.3 or depends on unparser then it won’t be installable along current version of Reek.

I see your point, but i beliebe being tentative about this makes dependency hell even worse. We should update aggressively and tool x then just *has* to update as well.

chastell:
> and tool x then just _has_ to update as well

I KNOW, RIGHT? :laughing: 

Thanks so much for the merge!

PR #808, opened 2016-01-01
Make exit codes configurable.

### LINKED ISSUE #790: CodeClimate engine - exit code
Currently, Reek exits with a status of 2 when it detect smells.

When running Reek as a Code Climate engine (see #763), the engine needs to exit with a code of 0 when there are smells.

I'm trying to think how this can be elegantly handled in Reek.

The most direct way would be to add a conditional in `Application#report_smells` which reads the `options.report_format`, but that's essentially a type check smell.

I was thinking a better approach might be to set the status codes on a per-formatter basis, so each report would have methods for `#success_exit_code`, `#error_exit_code` and `#smells_exit_code`.

This has the advantage that any future engines can control the exit codes, but it's perhaps a premature optimisation since this is currently only needed for Code Climate.

Any thoughts?

troessner:
Fixes #790

### DIFF
--- a/features/step_definitions/sample_file_steps.rb
+++ b/features/step_definitions/sample_file_steps.rb
@@ -1,3 +1,22 @@
+Given(/^the smelly file 'smelly.rb'$/) do
+  contents = <<-EOS.strip_heredoc
+    class Smelly
+      def x; end
+    end
+  EOS
+  write_file('smelly.rb', contents)
+end
+
+Given(/^the clean file 'clean.rb'$/) do
+  contents = <<-EOS.strip_heredoc
+    # Explanatory comment
+    class Clean
+      def foo; end
+    end
+  EOS
+  write_file('clean.rb', contents)
+end
+
 Given(/^the smelly file 'demo.rb' from the example in the README$/) do
   contents = <<-EOS.strip_heredoc
     class Dirty
--- a/lib/reek/cli/application.rb
+++ b/lib/reek/cli/application.rb
@@ -11,53 +11,45 @@ module Reek
     # command line.
     #
     class Application
-      STATUS_SUCCESS = 0
-      STATUS_ERROR   = 1
-      STATUS_SMELLS  = 2
       attr_reader :configuration
 
       private_attr_accessor :status
       private_attr_reader :command, :options
 
       def initialize(argv)
-        @status = STATUS_SUCCESS
         @options = configure_options(argv)
+        @status = options.success_exit_code
         @configuration = configure_app_configuration(options.config_file)
         @command = ReekCommand.new(OptionInterpreter.new(options))
       end
 
       def execute
-        return status if error_occured?
         command.execute self
         status
       end
 
       def report_success
-        self.status = STATUS_SUCCESS
+        self.status = options.success_exit_code
       end
 
       def report_smells
-        self.status = STATUS_SMELLS
+        self.status = options.failure_exit_code
       end
 
       private
-
-      def error_occured?
-        status == STATUS_ERROR
-      end
 
       def configure_options(argv)
         Options.new(argv).parse
       rescue OptionParser::InvalidOption => error
         $stderr.puts "Error: #{error}"
-        exit STATUS_ERROR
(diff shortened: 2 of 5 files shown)


features/step_definitions/sample_file_steps.rb
```diff
@@ -1,3 +1,22 @@
...
+
+Given(/^the clean file 'clean.rb'$/) do
+  contents = <<-EOS.strip_heredoc
+    # Explanatory comment
+    class Clean
+      def foo; end
+    end
+  EOS
+  write_file('clean.rb', contents)
+end
```
troessner: I believe we could and should reduce and unify all of our used code samples since they are quite a few that overlap. But that would be a separate pull request.

lib/reek/cli/application.rb
```diff
@@ -11,53 +11,45 @@ module CLI
...
       def report_smells
-        self.status = STATUS_SMELLS
+        self.status = options.failure_exit_code
       end
 
       private
 
-      def error_occured?
-        status == STATUS_ERROR
-      end
```
troessner: That method could never be called at runtime.

lib/reek/cli/application.rb
```diff
@@ -11,53 +11,45 @@ module CLI
...
-      STATUS_SMELLS  = 2
       attr_reader :configuration
 
       private_attr_accessor :status
       private_attr_reader :command, :options
 
       def initialize(argv)
-        @status = STATUS_SUCCESS
         @options = configure_options(argv)
+        @status = options.success_exit_code
```
troessner: `status` now needs to be initialized after `options`, hence the moving of the lines.

lib/reek/cli/options.rb
```diff
@@ -9,9 +9,14 @@ module CLI
     #
     # See {file:docs/Command-Line-Options.md} for details.
     #
-    # :reek:TooManyInstanceVariables: { max_instance_variables: 7 }
+    # :reek:TooManyInstanceVariables: { max_instance_variables: 9 }
     # :reek:Attribute: { enabled: false }
+    #
     class Options
+      DEFAULT_SUCCESS_EXIT_CODE  = 0
+      DEFAULT_ERROR_EXIT_CODE    = 1
```
troessner: DEFAULT_ERROR_EXIT_CODE is **not** configurable at the moment. I just put it here for consistency reasons. 
We exit the hard way using this code only at 2 places:
1.) https://github.com/troessner/reek/pull/808/files#diff-8cb50151ad3a458ecb02fe222bb330eaR52
2.) https://github.com/troessner/reek/pull/808/files#diff-8cb50151ad3a458ecb02fe222bb330eaR45

and that's where we're actually parsing options. So making them configurable right now would be completely pointless.
If we start to introduce this error code across then we should make it configurable as well.
And we should probably. Thinking about #809

lib/reek/cli/options.rb
```diff
@@ -19,16 +24,20 @@ class Options
...
-        @show_links       = true
-        @smells_to_detect = []
-        @colored          = color_support?
+        @argv              = argv
+        @parser            = OptionParser.new
+        @report_format     = :text
+        @location_format   = :numbers
+        @show_links        = true
+        @smells_to_detect  = []
+        @colored           = color_support?
```
troessner: Changes here and above are just indentation changes.

chastell:
LGTM, but @mvz didn’t comment on this approach in #790, so I’ll defer merging to him.

mvz:
I'm not sure all this is needed, but technically I have no problems with this approach. Merging.

troessner:
What do you mean with "all this is needed"? It's a prerequisite for the codeclimate integration.

mvz:
My thoughts were:
* We could have used just one hard-code exit code for the code climate reporter (less flexible, but enough for that use case).
* We may not need to customize the success exit code.
On the plus side:
* This is a flexible and consistent solution that will allow people to customize codes to their heart's content.

troessner:
>> This is a flexible and consistent solution that will allow people to customize codes to their heart's content.

:+1: ;)

PR #811, opened 2016-01-02
Switching to keyword arguments for SourceCode initializer

tuexss:
Improves readability

### DIFF
--- a/lib/reek/source/source_code.rb
+++ b/lib/reek/source/source_code.rb
@@ -21,7 +21,7 @@ module Reek
       # code   - Ruby code as String
       # origin - 'STDIN', 'string' or a filepath as String
       # parser - the parser to use for generating AST's out of the given source
-      def initialize(code, origin, parser: Parser::Ruby22)
+      def initialize(code: raise, origin: raise, parser: Parser::Ruby22)
         @source = code
         @origin = origin
         @parser = parser
@@ -39,10 +39,10 @@ module Reek
       # :reek:DuplicateMethodCall: { max_calls: 2 }
       def self.from(source)
         case source
-        when File     then new(source.read, source.path)
-        when IO       then new(source.readlines.join, IO_IDENTIFIER)
-        when Pathname then new(source.read, source.to_s)
-        when String   then new(source, STRING_IDENTIFIER)
+        when File     then new(code: source.read,           origin: source.path)
+        when IO       then new(code: source.readlines.join, origin: IO_IDENTIFIER)
+        when Pathname then new(code: source.read,           origin: source.to_s)
+        when String   then new(code: source,                origin: STRING_IDENTIFIER)
         end
       end
 
--- a/spec/reek/source/source_code_spec.rb
+++ b/spec/reek/source/source_code_spec.rb
@@ -6,20 +6,20 @@ RSpec.describe Reek::Source::SourceCode do
   describe '#syntax_tree' do
     it 'associates comments with the AST' do
       source = "# this is\n# a comment\ndef foo; end"
-      source_code = Reek::Source::SourceCode.new(source, '(string)')
+      source_code = Reek::Source::SourceCode.new(code: source, origin: '(string)')
       result = source_code.syntax_tree
       expect(result.leading_comment).to eq "# this is\n# a comment"
     end
 
     it 'cleanly processes empty source' do
-      source_code = Reek::Source::SourceCode.new('', '(string)')
+      source_code = Reek::Source::SourceCode.new(code: '', origin: '(string)')
       result = source_code.syntax_tree
       expect(result).to be_nil
     end
 
     it 'cleanly processes empty source with comments' do
       source = "# this is\n# a comment\n"
-      source_code = Reek::Source::SourceCode.new(source, '(string)')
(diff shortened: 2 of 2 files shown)


lib/reek/source/source_code.rb
```diff
@@ -21,10 +21,10 @@ class SourceCode
...
       # origin - 'STDIN', 'string' or a filepath as String
       # parser - the parser to use for generating AST's out of the given source
-      def initialize(code, origin, parser: Parser::Ruby22)
-        @source = code
-        @origin = origin
-        @parser = parser
+      def initialize(args)
+        @source = args[:code]
+        @origin = args[:origin]
+        @parser = args[:parser] || Parser::Ruby22
```
troessner: That's actually a step backwards - using Hashes for just slurping in all arguments is an anti-pattern since it decreases the expressiveness of the method signature (or even obliterates it, like in this case). Back in the day using the "hash-for-all-the-things" was idiomatic ruby since ruby didn't have keywords arguments, which it now does. So if we want to improve this method we should rather switch all arguments to keyword arguments (and keep the default value for parser in the signature as well) and not the other way round ;)

lib/reek/source/source_code.rb
```diff
@@ -36,14 +36,14 @@ def initialize(code, origin, parser: Parser::Ruby22)
       # @param source [File|IO|String] - the given source
       #
       # @return an instance of SourceCode
-      # :reek:DuplicateMethodCall: { max_calls: 2 }
```
troessner: The DuplicateMethodCall refers to the `source.read` below and is still valid.

lib/reek/source/source_code.rb
```diff
@@ -36,14 +36,14 @@ def initialize(code, origin, parser: Parser::Ruby22)
...
-        when Pathname then new(source.read, source.to_s)
-        when String   then new(source, STRING_IDENTIFIER)
-        end
+        params = case source
+                 when File     then { code: source.read,           origin: source.path }
+                 when IO       then { code: source.readlines.join, origin: IO_IDENTIFIER }
+                 when Pathname then { code: source.read,           origin: source.to_s }
+                 when String   then { code: source,                origin: STRING_IDENTIFIER }
+                 end
+        new(params)
```
troessner: The way it was before was very straightforward and easy to read - this change actually decreases readability.without apparent benefit.

spec/reek/source/source_code_spec.rb
```diff
@@ -6,20 +6,20 @@
   describe '#syntax_tree' do
     it 'associates comments with the AST' do
       source = "# this is\n# a comment\ndef foo; end"
-      source_code = Reek::Source::SourceCode.new(source, '(string)')
+      source_code = Reek::Source::SourceCode.new(code: source, origin: '(string)')
```
troessner: This **is** actually more readable than before, but like explained above you can switch to keyword arguments and then keep the exact same syntax here ;)

lib/reek/source/source_code.rb
troessner: And we should keep `source.read` for readability reasons and since it has no significant negative impact (actually this whole SmellDetector should ignore duplicate calls when they happen in mutually exclusive branches, but that's another topic).

lib/reek/source/source_code.rb
troessner: Addendum: Just to be clear: Changing the initializer above to use keyword arguments will improve readability here - using the "params" indirection rather not.

lib/reek/source/source_code.rb
tuexss: TIL: keyword arguments :+1:

lib/reek/source/source_code.rb
tuexss: Exactly. One branch shouldn't be held accountable for what another branch does.

lib/reek/source/source_code.rb
tuexss: Well, imagine you'd have to do a second thing based on the switch. This is a boilerplate code spawn kitchen :D

tuexss:
Amended. Please have another look.

troessner:
Yes, this looks good now!

2 things though:

1.) We need to change [this](https://github.com/troessner/reek/pull/811/files#diff-f9dae46a54ebb72c2c734ff15d80d26eR24) to 

```Ruby
def initialize(code: raise, origin: raise, parser: Parser::Ruby22)
```

Reason for this is that `Reek` still support Ruby 2.0 where keyword arguments are handled differently than in Ruby >= 2.1. 
Quoting from [here](https://robots.thoughtbot.com/ruby-2-keyword-arguments):

>> Unfortunately, Ruby 2.0 doesn’t have built-in support for required keyword arguments. Luckily, Ruby 2.1 introduced required keyword arguments, which are defined with a trailing colon

With that nice little hack from above we kind enforce similar behaviour and use this throughout our codebase (until ruby 2.0 dies and we will remove this from our codebase again)

2.) Travis fails but should be green after [1]

tuexss:
all green now :)

troessner:
Nice! Merged, thanks!

chastell:
Oh nice! Thanks, @tuexss!

PR #812, opened 2016-01-02
Move typebased source patterns to lookup table

tuexss:
Second attempt at eliminating the smell, even if it was a false alarm.

This was more of an academic exercise, as I wanted to find out how to get rid off the switch statement and find a better separation for data and logic. I'm not sure if this PR has any value, as the original version was more readable imho. If the concept of the naming of the origin wasn't mixing different levels of abstraction, it would be much easier to duck type our way out of having this switch statement/lookup table altogether.

Anyway, I learned a lot about Hashes and Closures on the way, so discarding it is also fine with me :)

wdyt?

### DIFF
--- a/lib/reek/source/source_code.rb
+++ b/lib/reek/source/source_code.rb
@@ -13,6 +13,10 @@ module Reek
     class SourceCode
       IO_IDENTIFIER     = 'STDIN'
       STRING_IDENTIFIER = 'string'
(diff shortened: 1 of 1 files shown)


mvz:
I agree that this is an interesting solution, and that the original version is more readable :smile:.

troessner:
>>  I'm not sure if this PR has any value, as the original version was more readable imho.

Exactly ;)
Closing this one.

PR #818, opened 2016-01-06
Improve documentation for UnusedPrivateMethod.

troessner:


mvz:
The second 'limitation' is actually by intent. Although my initial idea was to avoid the reverse-template-anti-pattern (i.e., providing some private methods to be used by methods in subclasses), I also think using private methods to fill out a template in a superclass is not a good idea.

docs/Unused-Private-Method.md
```diff
@@ -45,3 +45,14 @@ class ContextBuilder
 end
 ```
 
+Note that disabling this detector via comment works on a `class` scope, not
+a method scope (like you can see above).
+
+## Known limitations
+
+* Method calls via dynamic dispatch (e.g. via `send`) is something `Reek` (or any other
```
mvz: You could add the filters from Rails as an example here as well.

troessner: Yes, good idea, updated!

troessner:
>> The second 'limitation' is actually by intent.

I'd say by design. Even if we wanted to not alert in this case, we couldn't. 

>> I also think using private methods to fill out a template in a superclass is not a good idea.

Agreed, I updated the text to make this clearer.

troessner:
Addressed all comments!

mvz:
:+1:

chastell:
> using private methods to fill out a template in a superclass is not a good idea

Oh interesting! Why?

troessner:
The "private" is the problem here, I guess @chastell. For me the template pattern is about adapting "behaviour". But private methods are by definition an implementation detail, not an interface.

chastell:
> For me the template pattern is about adapting "behaviour". But private methods are by definition an implementation detail, not an interface.

Hm, but that equates the template algorithm’s steps with public interface of the classes and leaves no place for distinction between how the whole family can be called and how it operates.

Let me specify the question (and feel free to ignore it, of course!). In Lovely Rufus (a text wrapping library/CLI tool) I have middleware layes massaging text. In one of them I want to strip comment delimiters from code comment paragraphs, in another quotation marks from email paragraphs (and – in both cases – add them back, cleaned up, when the rewrapped text makes its way back up the middleware stack). [`CodeCommentStripper`](https://github.com/chastell/lovely_rufus/blob/b5e4936c55575c52e1f4e8c8a787d6f17c9d608d/lib/lovely_rufus/layers/code_comment_stripper.rb) and [`EmailQuoteStripper`](https://github.com/chastell/lovely_rufus/blob/b5e4936c55575c52e1f4e8c8a787d6f17c9d608d/lib/lovely_rufus/layers/email_quote_stripper.rb) differ only by what constitutes the quotation marks and their clean-up, with everything else extracted to a common [`QuoteStripper`](https://github.com/chastell/lovely_rufus/blob/b5e4936c55575c52e1f4e8c8a787d6f17c9d608d/lib/lovely_rufus/layers/quote_stripper.rb) superclass. I _definitely_ don’t want to expose publicly anything else than `#call` (which is the common `Layer` protocol). Is this an impure template pattern that can be refactored (how?), or is it just an edge case outside the general rule?

And to make this all Reek-worthwhile: as can be seen above, we don’t seem to consider an uncalled private `alias_method` to be an UnusedPrivateMethod. Should we?

troessner:
>> Hm, but that equates the template algorithm’s steps with public interface of the classes and leaves no place for distinction between how the whole family can be called and how it operates.

But then I could argue that you're not talking about a family but a group of unrelated families ;)

>> CodeCommentStripper and EmailQuoteStripper differ only by what constitutes the quotation marks and their clean-up, with everything else extracted to a common QuoteStripper superclass. 

And I could argue that CodeCommentStripper and EmailQuoteStripper only seem similar on the outside, but aren't really. I don't know lovely rufus enough to make **any** claims here of course. I'm just saying that whenever I have seen the template pattern where private methods were changing those classes were actually not related on second inspection.

>> And to make this all Reek-worthwhile: as can be seen above, we don’t seem to consider an uncalled private alias_method to be an UnusedPrivateMethod. Should we?

Good point, I think we should.

mvz:
> Is this an impure template pattern that can be refactored (how?), or is it just an edge case outside the general rule?

I would probably choose to move the logic in CodeCommentStripper and EmailQuoteStripper to different classes that would function as collaborators for QuoteStripper.

I'll read up on the template pattern again and see what it says about considerations like this. Testability considerations would urge the classes filling out the template to do so using public methods.

troessner:
>> I would probably choose to move the logic in CodeCommentStripper and EmailQuoteStripper to different classes that would function as collaborators for QuoteStripper.
Testability considerations would urge the classes filling out the template to do so using public methods.

Couldn't agree more ;)

mvz:
The GoF book does indeed suggest using protected (in Ruby: private) methods to fill out the template.

I may need to write a blog post on this :smile:.

troessner:
W00t?! :laughing:

PR #822, opened 2016-01-12
Add CodeClimate Docker integration.

troessner:
References: 

* [Engine example](http://blog.codeclimate.com/blog/2015/07/07/build-your-own-codeclimate-engine/)
* [Spec](https://github.com/codeclimate/spec)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+# CodeClimate specification: https://github.com/codeclimate/spec/blob/master/SPEC.md
+#
+# Build and run via:
+#   docker build -t codeclimate/codeclimate-reek . && docker run codeclimate/codeclimate-reek
+
+FROM codeclimate/alpine-ruby:b38
+
+MAINTAINER The Reek core team
+
+ENV code_dir /code
+ENV app_dir /usr/src/app
+ENV user app
+
+RUN apk --update add git
+
+ADD . ${app_dir}
+
+WORKDIR ${app_dir}
+
+RUN bundle install --without debugging development
+RUN adduser -u 9000 -D ${user}
+RUN chown -R ${user}:${user} ${app_dir}
+
+VOLUME ${code_dir}
+WORKDIR ${code_dir}
+USER ${user}
+
+CMD [ "/usr/src/app/bin/code_climate_reek" ]
--- a/bin/code_climate_reek
+++ b/bin/code_climate_reek
@@ -1,0 +1,57 @@
+#!/usr/bin/env ruby
+#
+# Wrapper for the CodeClimate integration.
+
+require_relative '../lib/reek'
+require_relative '../lib/reek/cli/application'
+
+# Map input coming from CodeClimate to Reek.
+class CodeClimateToReek
+  # Following the spec (https://github.com/codeclimate/spec/blob/master/SPEC.md)
+  # we have to exit with a zero for both failure and success.
+  ENGINE_CONFIGURATION = [
+    '-f', 'code_climate',
+    '--failure-exit-code', '0',
+    '--success-exit-code', '0'
+  ]
+
+  attr_reader :configuration_file_path, :include_paths_key, :include_paths_default
+
+  def initialize(configuration_file_path: '/config.json',
+                 include_paths_key: 'include_paths',
+                 include_paths_default: [])
+    @configuration_file_path = configuration_file_path
+    @include_paths_key       = include_paths_key
+    @include_paths_default   = include_paths_default
+  end
+
+  def cli_arguments
+    include_paths + ENGINE_CONFIGURATION
+  end
+
+  private
+
+  def configuration_file_exists?
+    Pathname.new(configuration_file_path).exist?
+  end
+
+  # The config.json file we try to read below might look like this:
+  # {
+  #   "include_paths":[
+  #     "lib",
+  #     "spec"
+  #   ]
+  # }
+  def include_paths
+    if configuration_file_exists?
+      config = JSON.parse File.read(configuration_file_path)
+      config.fetch include_paths_key, include_paths_default
+    else
+      include_paths_default
+    end
+  end
+end
+
+application = Reek::CLI::Application.new(CodeClimateToReek.new.cli_arguments)
+
+exit application.execute
--- a/spec/reek/report/code_climate_report_spec.rb
+++ b/spec/reek/report/code_climate_report_spec.rb
@@ -18,8 +18,8 @@ RSpec.describe Reek::Report::CodeClimateReport do
   context 'with empty source' do
     let(:source) { '' }
 
-    it 'prints empty json' do
-      expect { instance.show }.to output(/^\[\]$/).to_stdout
+    it 'prints an empty string' do
+      expect { instance.show }.to output('').to_stdout
     end
   end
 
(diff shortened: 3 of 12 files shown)


bin/code_climate_reek
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env ruby
+#
+# Wrapper for the CodeClimate integration.
+
+require_relative '../lib/reek'
+require_relative '../lib/reek/cli/application'
+
+module CodeClimate2Reek
+  CONFIGURATION_FILE_PATH = '/config.json'
+  INCLUDE_PATHS_KEY       = 'include_paths'
```
tuexss: what's with the `S` in `PATHS`? plural? possessive?

troessner: Nope, it's plural, because we get an array of those.

andyw8:
I built the image and added a `.codeclimate.yml` file but when I run it I'm seeing the following failure:

```
% codeclimate analyze --dev
Starting analysis
Running reek: Done!
error: (CC::Analyzer::Engine::EngineFailure) engine reek failed with status 1 and stderr
Configuration error: Directory `UnusedPrivateMethod` does not exist
```

andyw8:
Btw, I released v0.2 of [codeclimate-engine-rb](https://github.com/andyw8/codeclimate-engine-rb) which adds a `render` method to handle writing out the null byte characters.

bin/code_climate_reek
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env ruby
+#
+# Wrapper for the CodeClimate integration.
```
mvz: Since this file exists, would it make sense to not load the codeclimate-specific code when just running `reek`? It means we can also drop the codeclimate-engine-rb dependency I think. Of course, we would then have to make sure it is loaded inside the Docker container some other way.

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+ENV app_dir /usr/src/app
+ENV user app
+
+RUN apk --update add git
+
+ADD . ${app_dir}
+
+WORKDIR ${app_dir}
+
+RUN bundle install --without local_development development
```
mvz: Would just running `gem install reek` work here as well?

troessner: Yep, but then development would be more tedious. If you'd like to work on the docker set up including our latest changes (which **is** the default, at least that would make sense to me) you'd have to switch this line back and forth.
Additionally this seems to be idiomatic, at least if I check out the other codeclimate related docker files.

bin/code_climate_reek
troessner: >> Since this file exists, would it make sense to not load the codeclimate-specific code when just running reek? 

But then we would have code inside `Reek` that is basically not executable on its own. So if we go down this road, we should pull all out, otherwise this would be completely confusing and quite error prone. And then we're back again at our original discussion: Integrate it properly or use a wrapper :)
I'm not saying we can't change this in the future, but for the purpose of getting our integration up and running I'd rather postpone this discussion for now.

mvz: > For the purpose of getting our integration up and running I'd rather postpone this discussion for now.

Very sensible. :+1:

troessner: >> Very sensible. :+1:

That is my second forename!

tuexss: you have a loooooooooot of second forenames :D

troessner: Pssssscchhhhhhhht! :laughing:

troessner:
Still waiting for @mrb feedback if we're ready to go, but I believe this one is ready for final review @chastell && @mvz !

troessner:
Oh, and one last thing: Seems like my .dockerignore doesnt really work (spec/ e.g. gets copied to the image), I'll need to look into that, but this shouldn't keep us from reviewing ;)

Dockerfile
```diff
@@ -0,0 +1,28 @@
+# CodeClimate specification: https://github.com/codeclimate/spec/blob/master/SPEC.md
+#
+# Build and run via:
+#   docker build -t codeclimate/codeclimate-reek . && docker run codeclimate/codeclimate-reek
+
+FROM codeclimate/alpine-ruby:b38
```
chastell: Hmm, b38 seems to be five months old, but still the newest version. Does Code Climate has a plan to maintain this image?

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ENV user app
+
+RUN apk --update add git
+
+ADD . ${app_dir}
+
+WORKDIR ${app_dir}
+
+RUN bundle install --without debugging development
+RUN adduser -u 9000 -D ${user}
```
chastell: Why `-u 9000`? :)

bin/code_climate_reek
```diff
@@ -0,0 +1,57 @@
...
+
+require_relative '../lib/reek'
+require_relative '../lib/reek/cli/application'
+
+# Map input coming from CodeClimate to Reek.
+class CodeClimate2Reek
```
chastell: Is there any chance of spelling out the `2` as `To` or am I just old, not hip enough and should lrn2toler8 the barbarity? ;)

lib/reek/report/code_climate_formatter.rb
```diff
@@ -11,12 +11,12 @@ def initialize(warning)
         @warning = warning
       end
 
-      def to_hash
+      def render
```
chastell: :sparkling_heart:

Dockerfile
troessner: >> Hmm, b38 seems to be five months old, but still the newest version. 

But that's the base image - I wouldn't expect that to change often.

Dockerfile
troessner: It's in [the spec](https://github.com/codeclimate/spec/blob/master/SPEC.md#packaging). No idea why exactly :)

bin/code_climate_reek
troessner: I did change that now just for you ;)

spec/reek/report/code_climate_report_spec.rb
```diff
@@ -27,42 +27,19 @@
...
-            "location": {
-              "path": "string",
-              "lines": {
-                "begin": 1,
-                "end": 1
-              }
-            }
-          }
-        ]
+      expected = <<-EOS.delete("\n")
```
mvz: I can hardly wait to use `<<~` here :smile:.

Dockerfile
chastell: Ah, sorry, my thinking was clouded by my clogged sinuses – if `alpine:edge` (which `codeclimate/alpine-ruby:b38` uses) gets updated it’s enough to rebuild all of the descendant images, no need to bump their versions, right? If so then this is ok (although I wonder whether it shouldn’t be `codeclimate/alpine-ruby:latest` anyway).

Dockerfile
chastell: Sorry to hear that, but ok then. :laughing:

bin/code_climate_reek
chastell: Splendid. The world is better now, if only a b̶i̶t̶ byte. (I’ll show myself out.)

chastell:
I think I have no other comments. Let me know if you want to fix the `.dockerignore` or merge this and then fix separately. :)

Dockerfile
troessner: >> although I wonder whether it shouldn’t be codeclimate/alpine-ruby:latest anyway)

Actually it should just be 

>> codeclimate/alpine-ruby

since docker will then automatically look for the tag "latest". 
Unfortunately this doesn't exist which feels a little weird indeed.
@mrb an insights from you regarding this?

troessner:
>>  Let me know if you want to fix the .dockerignore or merge this and then fix separately. :)

I'd love to have @mrb 's final "Go" before we can merge this one.
Quite frankly, as much as I'd love to hear why the hell my dockerignore file is ignored, I **really** don't think it makes a difference here. `Reek` is already pretty tiny ;)
Maybe I'll look into it later.

mrb:
@troessner Hey! Sorry I've been traveling - I'll be able to check this out on Tues 19 Jan

mvz:
> dockerignore file is ignored

That's a pun waiting to happen.

troessner:
Haha!

mrb:
:+1: sent an email with some more details!

troessner:
@andyw8 I updated to codeclimate-engine-rb 0.3.0

Let me summarize what is left to do: 

1) Put content explaining in depth what an issue is about in a "content" hash in the JSON output. This gives users the ability to learn more about the issue that Reek is finding. If there aren't long form descriptions yet for anything we could probably work with a placeholder, maybe a link back to the Reek docs or something. 

2) Insert remediation_points into the output hash. This is what allows the Code Climate UI to "grade" files, which create the "grades" for repos, which people find very useful. It allows us to more granularly tell people in Pull Requests that they have made issues "better" or "worse," for example.

Resources:

* [Issues spec](https://github.com/codeclimate/spec/blob/master/SPEC.md#issues)
* [content](https://github.com/codeclimate/spec/blob/master/SPEC.md#contents)
* [remediation points](https://github.com/codeclimate/spec/blob/master/SPEC.md#remediation-points)

troessner:
I just implemented:

```
1) Put content explaining in depth what an issue is about in a "content" hash in the JSON output. This gives users the ability to learn more about the issue that Reek is finding. If there aren't long form descriptions yet for anything we could probably work with a placeholder, maybe a link back to the Reek docs or something.

2) Insert remediation_points into the output hash. This is what allows the Code Climate UI to "grade" files, which create the "grades" for repos, which people find very useful. It allows us to more granularly tell people in Pull Requests that they have made issues "better" or "worse," for example.
```

So as far as I can see, our (== Reek team) job here is done implementation-wise!

What's left to do:

1.) Merging this one

@mvz || @chastell please give this PR a final review and then let's merge it asap so we can close this matter. 
I know that [this here](https://github.com/troessner/reek/pull/822/files#diff-172c836b02c62ad22a91671c12a57db2R1) says "Dummy content". 
But we should merge it nevertheless - first of all nobody is using the integration at the moment and second will this allow @mrb to issue a pull request against master with the real content and remediation points

2.) @mrb  Fill in our [codeclimate configuration](https://github.com/troessner/reek/pull/822/files#diff-172c836b02c62ad22a91671c12a57db2R1) with real content and remediation points

@mrb after the merge of this pull request the ball is entirely in your park ;)

mrb:
@troessner Awesome!

chastell:
LGTM, merging!

Issue #788, opened 2015-12-17
Allow configuring Rake task with a FileList


avdgaag:
The Rake task currently allows you to specify a single glob pattern:

```ruby
Reek::Rake::Task.new do |task|
  task.source_files = 'lib/**/*rb'
end
```

Internally, this is fed to Rake’s `FileList` and converted to an array of actual filenames to pass to the `reek` executable.

What I would like is to define _multiple_ glob patterns, and exclude some specific files. This basically can’t be done with a glob pattern, but would be very easy if I could pass in my own `FileList`. This would be nice:

```ruby
Reek::Rake::Task.new do |task|
  task.source_files = FileList['lib/**/*rb'].exclude('lib/templates/**/*.rb')
end
```

Note that this is different from excluding files in the Reek configuration: that will silence errors, but will not stop Reek from unsuccessfully trying to parse .rb files that are not valid Ruby code (such as .rb files that contain ERB instructions, as with Rails generator templates).

My suggestion would be to allow setting both file lists and strings in the Rake task. Use the single string glob pattern if you want to, but you _can_ supply your own, hand-crafted filelist if you want to.

troessner:
Sounds like a good idea to me!
Care for a pull request?
I'd help you if you're up for it ;)

mvz:
> Note that this is different from excluding files in the Reek configuration: that will silence errors, but will not stop Reek from unsuccessfully trying to parse .rb files that are not valid Ruby code.

We should fix that as well probably.

avdgaag:
Sorry, forgot about this over the holidays. I’ll make a pull request for this particular change, stay tuned.

> We should fix that as well probably

That, however, is out of my league :smile:

troessner:
>> Note that this is different from excluding files in the Reek configuration: that will silence errors, but will not stop Reek from unsuccessfully trying to parse .rb files that are not valid Ruby code (such as .rb files that contain ERB instructions, as with Rails generator templates).

This is not correct, please see #814 for the corresponding discussion.

troessner:
#823 is merged, closing this one.

PR #823, opened 2016-01-12
Accept FileList as Rake task source files

avdgaag:
The Rake task currently insists on taking a single glob pattern, while Rake's FileList is perfectly capable of taking multiple patterns. By relaxing the type check and taking advantage of the fact that passing a `FileList` to `FileList.[]` returns the same `FileList`, we can allow developers to set a fully customized `FileList` without changing the original API.

The right place to test this is unclear to me. There are no unit tests for the Rake task, and only some integration tests. Any suggestions?

See #788.

### DIFF
--- a/docs/Rake-Task.md
+++ b/docs/Rake-Task.md
@@ -41,6 +41,14 @@ Reek::Rake::Task.new do |t|
 end
 ```
 
+Alternatively, you can create your own [Rake::FileList](http://rake.rubyforge.org/classes/Rake/FileList.html) and use that for `source_files`:
+
+```Ruby
+Reek::Rake::Task.new do |t|
+  t.source_files = FileList['lib/**/*.rb'].exclude('lib/templates/**/*.rb')
+end
+```
+
 ## Configuration via environment variables
 
 You can overwrite the following attributes by environment variables:
--- a/lib/reek/rake/task.rb
+++ b/lib/reek/rake/task.rb
@@ -85,17 +85,16 @@ module Reek
 
       # @public
       def source_files=(files)
-        raise ArgumentError, no_string_given_for_file_list_warning unless files.is_a?(String)
+        unless files.is_a?(String) || files.is_a?(FileList)
+          raise ArgumentError, 'File list should be a FileList or a String that can contain'\
+            " a glob pattern, e.g. '{app,lib,spec}/**/*.rb'"
+        end
         @source_files = FileList[files]
       end
 
       private
 
       private_attr_reader :fail_on_error, :name, :verbose
-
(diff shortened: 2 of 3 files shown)


lib/reek/rake/task.rb
```diff
@@ -85,7 +85,9 @@ def initialize(name = :reek)
 
       # @public
       def source_files=(files)
-        raise ArgumentError, no_string_given_for_file_list_warning unless files.is_a?(String)
+        unless files.is_a?(String) || files.is_a?(FileList)
+          raise ArgumentError, no_string_or_filelist_given_for_file_list_warning
```
troessner: Let's rather drop this ugly method (I know it was there before, but still) and just use:

```Ruby
raise ArgumentError, "File list should be a FileList or a String that can contain"\
  "a glob pattern, e.g. '{app,lib,spec}/**/*.rb'"
```

troessner:
>> There are no unit tests for the Rake task, and only some integration tests. 

An integration test would be good enough here.
Can you also please update our [Rake task docs](https://github.com/troessner/reek/blob/master/docs/Rake-Task.md)?

avdgaag:
@troessner I added a test and some docs. What do you think?

docs/Rake-Task.md
```diff
@@ -56,3 +64,5 @@ REEK_CFG="config/custom.reek" REEK_OPTS="-s" rake reek
 ```
 
 See also: [Reek-Driven-Development](Reek-Driven-Development.md)
+
+[Rake::FileList]: http://rake.rubyforge.org/classes/Rake/FileList.html
```
troessner: Why not use GFM and then do [link text](link)?

docs/Rake-Task.md
```diff
@@ -41,6 +41,14 @@ Reek::Rake::Task.new do |t|
 end
 ```
 
+Alternatively, you can create your own [Rake::FileList][] and use that for `source_files`:
```
troessner: Lets rather use GFM for links, see below.

troessner:
Nice job!
Please update with the 2 tiny changes mentioned above and squash to one commit, then we can merge this one!

avdgaag:
I’m note sure what you mean by a “GFM link” and a Markdown link — as far as I know, they’re the same. But I’ve changed my link to use the inline-style, guessing that is what you are referring to. I’ve also squashed the commits together.

troessner:
Yes, that's what i meant! ;)
Merging, thanks! 
Release upcoming in 20 minutes.

avdgaag:
Great, thanks!

troessner:
Aaaand 3.8.3 is released including this change ;)

PR #825, opened 2016-01-14
Improve our bundler and gemspec handling.

### LINKED ISSUE #821: Move all development dependencies in our gemspec to the Gemfile
Let's move all development dependencies in our gemspec to the Gemfile.
The separation right now doesn't make any sense and is academic, but not pragmatic.
And we use bundler for everything anyway.

WDYT?

troessner:
Fixes #821 with some small adjustments:

* I decided to put "development_dependencies" under the group "test" because we use them for running tests local and on travis
* I renamed "local_development" to "development"
* I removed the "yard" gem and the "redcarpet" gem. Why exactly did we have them in the first place? I can't really remember :)

Additionally I bumped `unparser` from 0.2.2 to 0.2.4 since bundler was acting up on ruby 2.1 and 2.0 for whatever reasons. I didn't investigate since this upgrade fixed it and seemed non-critical since we plan to remove `unparser` as soon as possible anyway.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,16 +1,26 @@
 source 'https://rubygems.org'
 
-# The gem's dependencies are specified in the gemspec
 gemspec
 
-group :local_development do
-  gem 'pry'
-  gem 'yard', '~> 0.8.7'
+group :development do
+  gem 'aruba',         '~> 0.10.0'
+  gem 'ataru',         '~> 0.2.0'
+  gem 'cucumber',      '~> 2.0'
+  gem 'factory_girl',  '~> 4.0'
+  gem 'rake',          '~> 10.0'
+  gem 'rspec',         '~> 3.0'
+  gem 'rubocop',       '~> 0.34.0'
+  gem 'yard',          '~> 0.8.7'
 
+  platforms :mri do
+    gem 'redcarpet', '~> 3.3.1'
+  end
+end
+
+group :debugging do
+  gem 'pry'
   platforms :mri do
     gem 'pry-byebug'
     gem 'pry-stack_explorer'
-
-    gem 'redcarpet', '~> 3.3.1'
   end
 end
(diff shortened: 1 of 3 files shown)


troessner:
Update: And I removed "activesupport" since we were not using this anymore ;)

mvz:
> I removed the "yard" gem and the "redcarpet" gem. Why exactly did we have them in the first place? 

To create the documentation of course :-).

troessner:
But none of us 3 is using this locally, are we? I *never* do that. I am all in for code documentation yarddoc style but i still think we can throw out the gem, cant we?

mvz:
I sometimes run it locally to check the result of documentation changes.

mvz:
That said, we can probably leave the gem out of the bundle.

Gemfile
```diff
@@ -3,14 +3,20 @@ source 'https://rubygems.org'
...
+  gem 'aruba',         '~> 0.10.0'
+  gem 'ataru',         '~> 0.2.0'
+  gem 'cucumber',      '~> 2.0'
+  gem 'factory_girl',  '~> 4.0'
+  gem 'rake',          '~> 10.0'
+  gem 'rspec',         '~> 3.0'
+  gem 'rubocop',       '~> 0.34.0'
+end
 
+group :development do
```
mvz: I think I preferred you original development/debugging proposal. With test/development, combined with how the groups are named in Rails, I get confused why rake would be in test.

Oh, maybe the gems in test don't actually need to be in a group?

troessner: I changed it back to my initial suggestion.

troessner:
>> That said, we can probably leave the gem out of the bundle.

But you're a team member and should feel comfortable with our gem set up -> I put it back in (and redcarpet ofc).

chastell:
> I removed "activesupport" since we were not using this anymore

:tada:

> Merge?

I’m happy to do that. :)

mvz:
:+1: thanks, @troessner

PR #827, opened 2016-01-15
Use stored source instead of Unparser to format ASTs

troessner:


mvz:
It seems the thing to do is to implement a simple formatter in each of our SexpExtensions, for the relevant node types. I wouldn't use `#to_s` though, since we need that to inspect the AST structure sometimes.

troessner:
>> It seems the thing to do is to implement a simple formatter in each of our SexpExtensions, for the relevant node types.

Yes, exactly. We'd need to drop SexpFormatter completely, first of all because we'd remove `unparser` from it, and second of all, the logic for "unparsing" is highly specific. We could either blow up SexpFormatter out of proportion .....or put the logic where we used to call SexpFormatter.

mvz:
@troessner, @chastell this is ready for review. I went a completely different route, using the actual source of the expression as provided by parser.

features/configuration_loading.feature
```diff
@@ -18,7 +18,7 @@ Feature: Offer different ways how to load configuration
       """
       smelly.rb -- 3 warnings:
         [4, 5]:DuplicateMethodCall: Smelly#m calls @foo.bar 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
-        [4, 5]:DuplicateMethodCall: Smelly#m calls puts(@foo.bar) 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
+        [4, 5]:DuplicateMethodCall: Smelly#m calls puts @foo.bar 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
```
troessner: Why not leave the brackets? I find this better readable.

features/samples.feature
```diff
@@ -77,8 +77,8 @@ Feature: Basic smell detection
...
       DuplicateMethodCall: OptionParser#make_switch calls default_style.guess(arg = a) 4 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
       DuplicateMethodCall: OptionParser#make_switch calls long << (o = q.downcase) 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
-      DuplicateMethodCall: OptionParser#make_switch calls notwice(NilClass, klass, "type") 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
-      DuplicateMethodCall: OptionParser#make_switch calls notwice(if a ... end, klass, "type") 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
+      DuplicateMethodCall: OptionParser#make_switch calls notwice(NilClass, klass, 'type') 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
+      DuplicateMethodCall: OptionParser#make_switch calls notwice(a ? Object : TrueClass, klass, 'type') 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
```
troessner: Hmmm, so we're replacing 

>> if a ... end

with

>> a ? Object : TrueClass

? Can you explain why we should do this?

lib/reek/ast/sexp_extensions/send.rb
```diff
@@ -52,6 +52,10 @@ def attribute_writer?
         def attr_with_writable_flag?
           method_name == :attr && args.any? && args.last.type == :true
         end
+
+        def name
+          method_name
```
troessner: I'd prefer `alias_method` to make our intention clearer.

spec/reek/ast/node_spec.rb
```diff
@@ -2,9 +2,57 @@
...
+    end
+
+    it 'gives te right result for send with a receiver' do
+      ast = Reek::Source::SourceCode.from('baz.foo(bar)').syntax_tree
+      expect(ast.format_to_ruby).to eq 'baz.foo(bar)'
+    end
+
+    it 'gives the right result for if' do
+      source = <<-SRC
+      if foo
```
troessner: Lets indent that 2 more ;)

spec/reek/smells/duplicate_method_call_spec.rb
```diff
@@ -47,7 +47,7 @@ def double_thing(other)
 
     it 'reports call parameters' do
       src = 'def double_thing() @other.thing(2,3) + @other.thing(2,3) end'
-      expect(src).to reek_of(:DuplicateMethodCall, name: '@other.thing(2, 3)')
+      expect(src).to reek_of(:DuplicateMethodCall, name: '@other.thing(2,3)')
```
troessner: Why this change?

troessner:
First review -> done. 
Couple of questions and comments but nothing bigger, looks great overall!

features/configuration_loading.feature
mvz: I'm now using the actual source text so it depends on how it's actually written there.

features/samples.feature
mvz: Same reason: It's the original source text.

lib/reek/ast/sexp_extensions/send.rb
mvz: Hm .. how about renaming `method_name` to `name` and making things consistent?

features/configuration_loading.feature
troessner: Makes sense ;)

features/samples.feature
troessner: Wow, that must be an ....interesting piece of code ;)

spec/reek/smells/duplicate_method_call_spec.rb
troessner: Ok, i can guess the answer :)

troessner:
Great job, merging!

PR #830, opened 2016-01-20
Introduce SimpleCov for specs

mvz:
This can help us find dead or untested code.

### DIFF
--- a/lib/reek/cli/reek_command.rb
+++ b/lib/reek/cli/reek_command.rb
@@ -12,11 +12,15 @@ module Reek
         options.sources.each do |source|
           reporter.add_examiner Examiner.new(source, smell_names, configuration: app.configuration)
         end
-        reporter.smells? ? app.report_smells : app.report_success
         reporter.show
+        result_code
       end
 
       private
+
+      def result_code
+        reporter.smells? ? options.failure_exit_code : options.success_exit_code
+      end
 
       def reporter
         @reporter ||= options.reporter
--- a/spec/reek/cli/application_spec.rb
+++ b/spec/reek/cli/application_spec.rb
@@ -1,0 +1,31 @@
+require_relative '../../spec_helper'
+require_lib 'reek/cli/application'
+
+RSpec.describe Reek::CLI::Application do
+  describe '#initialize' do
+    it 'exits with default error code on invalid options' do
+      call = lambda do
+        Reek::CLI::Silencer.silently do
+          Reek::CLI::Application.new ['--foo']
+        end
+      end
+      expect(call).to raise_error(SystemExit) do |error|
+        expect(error.status).to eq Reek::CLI::Options::DEFAULT_ERROR_EXIT_CODE
+      end
+    end
+  end
+
+  describe '#execute' do
+    let(:command) { double 'reek_command' }
+    let(:app) { Reek::CLI::Application.new [] }
+
+    before do
+      allow(Reek::CLI::ReekCommand).to receive(:new).and_return command
+    end
+
+    it "returns the command's result code" do
+      allow(command).to receive(:execute).and_return 'foo'
+      expect(app.execute).to eq 'foo'
+    end
+  end
+end
--- a/spec/reek/cli/reek_command_spec.rb
+++ b/spec/reek/cli/reek_command_spec.rb
@@ -1,0 +1,45 @@
+require_relative '../../spec_helper'
+require_lib 'reek/cli/reek_command'
+require_lib 'reek/cli/options'
+require_lib 'reek/cli/option_interpreter'
+
+RSpec.describe Reek::CLI::ReekCommand do
+  describe '#execute' do
+    let(:options) { Reek::CLI::Options.new [] }
+    let(:option_interpreter) { Reek::CLI::OptionInterpreter.new(options) }
+
+    let(:reporter) { double 'reporter' }
+    let(:app) { double 'app' }
+
+    let(:command) { described_class.new option_interpreter }
+
+    before do
+      allow(option_interpreter).to receive(:reporter).and_return reporter
+      allow(reporter).to receive(:show)
+    end
+
+    context 'when no smells are found' do
+      before do
+        allow(option_interpreter).to receive(:sources).and_return []
+        allow(reporter).to receive(:smells?).and_return false
+      end
+
+      it 'returns a success code' do
+        result = command.execute app
+        expect(result).to eq Reek::CLI::Options::DEFAULT_SUCCESS_EXIT_CODE
+      end
+    end
+
+    context 'when smells are found' do
+      before do
+        allow(option_interpreter).to receive(:sources).and_return []
+        allow(reporter).to receive(:smells?).and_return true
+      end
+
+      it 'returns a failure code' do
+        result = command.execute app
+        expect(result).to eq Reek::CLI::Options::DEFAULT_FAILURE_EXIT_CODE
+      end
+    end
+  end
+end
--- a/spec/reek/rake/task_spec.rb
+++ b/spec/reek/rake/task_spec.rb
@@ -1,0 +1,33 @@
+require_relative '../../spec_helper'
+require_lib 'reek/rake/task'
+
+RSpec.describe Reek::Rake::Task do
+  describe '#source_files' do
+    it 'is set to "lib/**/*.rb" by default' do
+      task = Reek::Rake::Task.new
+      expect(task.source_files).to eq FileList['lib/**/*.rb']
+    end
+  end
+
+  describe '#source_files=' do
+    it 'sets source_files to a FileList when passed a string' do
+      task = Reek::Rake::Task.new
+      task.source_files = '*.rb'
+      expect(task.source_files).to eq FileList['*.rb']
+    end
+  end
+
+  # SMELL: Testing a private method
+  describe '#command' do
+    let(:task) { Reek::Rake::Task.new }
+
+    it 'does not include a config file by default' do
+      expect(task.send(:command)).not_to include '-c'
+    end
+
+    it 'includes a config file when set' do
+      task.config_file = 'foo.reek'
+      expect(task.send(:command)[1..2]).to eq ['-c', 'foo.reek']
+    end
+  end
+end
--- a/tasks/test.rake
+++ b/tasks/test.rake
@@ -4,7 +4,7 @@ require 'rspec/core/rake_task'
 namespace 'test' do
   RSpec::Core::RakeTask.new('spec') do |t|
     t.pattern = 'spec/reek/**/*_spec.rb'
-    t.ruby_opts = ['-Ilib -w']
+    t.ruby_opts = ['-rsimplecov -Ilib -w']
   end
 
   desc 'Tests code quality'
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ pkg
 spec/output
 tmp
 doc
+coverage
 .yardoc
 .bundle
 .rspec
--- a/.simplecov
+++ b/.simplecov
@@ -1,0 +1,10 @@
+SimpleCov.start do
+  track_files 'lib/**/*.rb'
+  add_filter 'lib/reek/version.rb' # version.rb is loaded too early to test
+end
+
+SimpleCov.at_exit do
+  SimpleCov.result.format!
+  SimpleCov.minimum_coverage 98.9
+  SimpleCov.minimum_coverage_by_file 81.4
+end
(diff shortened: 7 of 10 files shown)


troessner:
How do you intend to use it? As casual rake task?

mvz:
SimpleCov always runs unless told not to, so after each `rake spec` there's up-to-date coverage info in `coverage/`.

troessner:
But then how about making this part of our build? Otherwise this will be lost, at least to me...

mvz:
It is already part of the build. Just run `bundle exec rake spec` as usual and the coverage report will appear in `coverage/`.

Or do you mean adding a check to require a certain coverage level?

troessner:
If it cant make the build fail, its not part of our build. So yes, i think we should have a check for it, otherwise it is meaningless - at least to me.

mvz:
I'll see what I can do.

mvz:
@troessner updated so the build fails when the coverage drops below current levels.

spec/reek/cli/application_spec.rb
```diff
@@ -0,0 +1,10 @@
+require_relative '../../spec_helper'
+require_lib 'reek/cli/application'
+
+RSpec.describe Reek::CLI::Application do
+  describe '#initialize' do
+    it 'works' do
+      Reek::CLI::Application.new []
+    end
```
troessner: I dont get it. What's the point of this test? Satisfying simplecov? Shouldnt we rather have proper specs?

mvz: I had a reasoning (i.e., at least get it loaded so its coverage isn't zero), but I realize it's no better than having zero coverage.

I'll see if I can add some proper specs here and in the Task spec as well.

spec/spec_helper.rb
```diff
@@ -1,4 +1,20 @@
...
+
+SimpleCov.start do
+  track_files 'lib/**/*.rb'
+  add_filter 'lib/reek/version.rb' # version.rb is loaded too early to test
+end
+
+if ENV['FAIL_ON_LOW_COVERAGE'] == '1'
+  SimpleCov.at_exit do
+    SimpleCov.result.format!
+    SimpleCov.minimum_coverage 98.71
```
andyw8: Seems having these values will lead to a lot of churn of this file.

One alternative approach I like is to specify 100% coverage and use `# :nocov:` markers for the lines to be ignored.

troessner: I'd rather not pollute our source code with "nocov" everywhere.
Especially since I am highly sceptical of simplecov anyways (i'd prefer setting up mutant properly, but thats a different story)

mvz: > Seems having these values will lead to a lot of churn of this file.

Possibly, but then again I'm not going to be watching this number very closely and update it every time our coverage increases by 0.01% :laughing:.

My main goal with this is to be able to quickly spot parts of the code that are not used anymore. If an external runner were available (e.g., `simplecov rspec spec`), I'd use that instead.

troessner:
@mvz you're still on this regarding proper specs above, right? 

Just asking, not that you're waiting for a merge and I'm waiting for an update ;)

spec/spec_helper.rb
```diff
@@ -1,4 +1,20 @@
 require 'pathname'
+
+require 'simplecov'
+
+SimpleCov.start do
```
bf4: Is there an advantage to putting the config and start here, rather than in '.simplecov.'? It gets loades when simplecov and is required, so no churn in spec helpers

I've also found it best to leave off when not running whole suite as it can slow down the prog exit

mvz:
@troessner Yes, although I got sidetracked on the separate runner idea :-).

spec/spec_helper.rb
mvz: Good points @bf4, I'll look into them.

mvz:
Tests added. I noticd a bug in Task that I'll create a separate ticket for.

lib/reek/cli/reek_command.rb
```diff
@@ -12,8 +12,8 @@ def execute(app)
         options.sources.each do |source|
           reporter.add_examiner Examiner.new(source, smell_names, configuration: app.configuration)
         end
-        reporter.smells? ? app.report_smells : app.report_success
         reporter.show
+        reporter.smells? ? options.failure_exit_code : options.success_exit_code
```
troessner: While we are at it, let's rename "options" here to "options_interpreter"!
It's super confusing like it is right now and trips me up every time :-(

lib/reek/rake/task.rb
```diff
@@ -92,6 +92,12 @@ def source_files=(files)
         @source_files = FileList[files]
       end
 
+      def command
```
troessner: Why this change?

spec/quality/reek_source_spec.rb
```diff
@@ -1,4 +1,5 @@
-require_relative '../spec_helper'
```
troessner: Isn't requiring the spec_helper kind of best practice even if it's technically not necessary?

I mean, we require files in /lib explicitly for the same reason even though quite a few times it wouldnt be necessary...

spec/reek/cli/application_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  describe '#execute' do
+    let(:command) { double 'reek_command' }
+    let(:app) { Reek::CLI::Application.new [] }
+
+    before do
+      allow(Reek::CLI::ReekCommand).to receive(:new).and_return command
+    end
+
+    it 'executes its command passing itself as a parameter' do
```
troessner: I'd consider this an implementation detail and rather not have this spec (since refactoring already could break tbis test, which it shouldnt).

spec/reek/cli/reek_command_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+        allow(option_interpreter).to receive(:sources).and_return []
+        allow(reporter).to receive(:smells?).and_return true
+      end
+
+      it 'returns a failure code' do
+        result = command.execute app
+        expect(result).to eq Reek::CLI::Options::DEFAULT_FAILURE_EXIT_CODE
+      end
+    end
+  end
```
troessner: Nice specs!

spec/reek/rake/task_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      task.source_files = '*.rb'
+      expect(task.source_files).to eq FileList['*.rb']
+    end
+  end
+
+  describe '#command' do
+    let(:task) { Reek::Rake::Task.new }
+
+    it 'does not include a config file by default' do
+      expect(task.command).not_to include '-c'
```
troessner: Hmm, so you made this public just for the test, now i see.
I'd then rather use `send` ....

tasks/test.rake
```diff
@@ -17,6 +17,12 @@ namespace 'test' do
     t.cucumber_opts = 'features --format progress --color'
   end
 
+  task 'enable_simplecov' do
+    ENV['SIMPLECOV'] = '1'
+  end
+
+  task 'spec' => 'test:enable_simplecov'
```
troessner: If you enable it on every spec run, why not drop that ENV charade completely and.....just require it unconditionally on spec run?

mvz: We don't want to enable it when running one spec file.

mvz: Another option is probably to add `-rsimplecov` inside the spec task, so `rspec something_spec.rb` won't activate it.

spec/reek/rake/task_spec.rb
mvz: Yes, that's probably the lesser evil :smile:.

spec/reek/cli/application_spec.rb
mvz: I agree.

spec/quality/reek_source_spec.rb
mvz: This spec is a special case and we don't want to enable simplecov for it.

troessner: Ok, makes sense. But then I'd prefer to have at least a comment on top of this file screaming in my face that this is special and why...

tasks/test.rake
troessner: >> We don't want to enable it when running one spec file.

Obvious reason which I completely missed :laughing: 

>> Another option is probably to add -rsimplecov inside the spec task, so rspec something_spec.rb won't activate it.

Ah, I think that's a better solution, let's go for that!

mvz:
Comments processed. @troessner please take another look. @chastell any comments on the RuboCop settings?

troessner:
Looks good to me now and can be merged. @chastell ?

spec/reek/cli/application_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      begin
+        Reek::CLI::Silencer.silently do
+          Reek::CLI::Application.new ['--foo']
+        end
+      rescue SystemExit => e
+        expect(e.status).to eq Reek::CLI::Options::DEFAULT_ERROR_EXIT_CODE
+        error_raised = true
+      end
+      expect(error_raised).to be_truthy
+    end
```
chastell: Hm, could this be written with the exception matcher? Something to the tune of (untested):

```ruby
call = -> { Reek::CLI::Silencer.silently { Reek::CLI::Application.new ['--foo'] } }
expect(call).to raise_error do |error|
 expect(error).to be_a SystemExit
 expect(error.status).to eq Reek::CLI::Options::DEFAULT_ERROR_EXIT_CODE
end
```

spec/reek/rake/task_spec.rb
```diff
@@ -0,0 +1,39 @@
+require_relative '../../spec_helper'
+require_lib 'reek/rake/task'
+
+RSpec.describe Reek::Rake::Task do
+  describe '#initialize' do
+    it 'works' do
+      Reek::Rake::Task.new
+    end
```
chastell: I think this should either be `it 'can be called without arguments'` (or whatever this test is testing) or dropped altogether. :)

chastell:
I have two minor nitpicks about tests, but LGTM otherwise. Good work! (Apologies for lagging recently, when I’m not sick I’m catching up with the exam session at the uni…)

spec/reek/cli/application_spec.rb
troessner: I think the rspec matcher only catches StandardError and below which wouldn't work for SystemExit.
Hence the way it is done above ;)

chastell: Hm, I just did a quick test and the below does work:

```ruby
call = -> { exit 10 }
expect(call).to raise_error(SystemExit)
```

troessner: Ah, interesting. It was a conjecture from my side, i thought thats why @mvz had written it that way :laughing:

chastell:
> @chastell any comments on the RuboCop settings?

Um, I must be missing something. What RuboCop settings?

spec/reek/cli/application_spec.rb
mvz: I was unaware of the extended matcher with blocks. I'll see if I can make that work.

spec/reek/rake/task_spec.rb
mvz: Oh, I thought I removed this one. It should go.

mvz:
> What RuboCop settings?

The ones that are part of a completely different pull request of course :laughing:.

spec/reek/cli/application_spec.rb
chastell: (In case you’re tempted – or, y’know, TOLD BY RUBOCOP – to go with `expect(call).to raise_error do |error|` just remember that `do` binds weaker than `{` and the block needs to bind to `raise_error`.) ;)

tasks/test.rake
```diff
@@ -4,7 +4,7 @@ require 'rspec/core/rake_task'
 namespace 'test' do
   RSpec::Core::RakeTask.new('spec') do |t|
     t.pattern = 'spec/reek/**/*_spec.rb'
-    t.ruby_opts = ['-Ilib -w']
+    t.ruby_opts = ['-rsimplecov -Ilib -w']
```
bf4: :100: : to only requiring (aka starting) simplecov when running the rake spec task and it will always be loaded before any app code!

mvz:
I resolved @chastell's nits. LGTM?

troessner:
I was hoping that my #852 would get merged first so that you had to fix the merge conflicts and not me :laughing: 

Merged!

mvz:
:laughing:

PR #834, opened 2016-01-22
Fetch a module's name from the source

### LINKED ISSUE #832: undefined method `full_name' for s(:lvar, :m):#<Class:0x00>
Code to reproduce

```ruby
m = Module.new

module m::N
end
```

Commit which I think introduced the regression: 8bf26f9dea4ed16e3282bbf9d4ed1dc98af21133

mvz:
Fixes #832.

### DIFF
--- a/spec/reek/ast/sexp_extensions_spec.rb
+++ b/spec/reek/ast/sexp_extensions_spec.rb
@@ -319,8 +319,7 @@ end
 RSpec.describe Reek::AST::SexpExtensions::ModuleNode do
   context 'with a simple name' do
     subject do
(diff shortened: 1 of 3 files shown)


spec/reek/ast/sexp_extensions_spec.rb
```diff
@@ -342,7 +341,7 @@
 
   context 'with a scoped name' do
     subject do
-      sexp(:module, sexp(:const, sexp(:const, nil, :Foo), :Bar), nil)
+      Reek::Source::SourceCode.from('module Foo::Bar; end').syntax_tree
```
troessner: We should have a helper for that, but we can add this another time.

PR #837, opened 2016-01-24
Use parser 2.3 in `SourceCode`.

troessner:
Kind of funny that none of us noticed ;)
Nothing broke because Parser::Ruby22 parses e.g.

>> user.&profile

just fine and the exact same way Parser::Ruby23 does.


mvz:
Well spotted!

PR #839, opened 2016-01-25
Descriptions and remediation points

mrb:
This adds remediation points and content to the Code Climate reporter, and adds Ruby files in the ratings paths section of the `.codeclimate.yml` file so this repo generates grades on Code Climate.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,3 +2,6 @@
 engines:
   reek:
     enabled: true
+ratings:
+  paths:
+  - "**.rb"
--- a/lib/reek/report/code_climate/code_climate_configuration.yml
+++ b/lib/reek/report/code_climate/code_climate_configuration.yml
@@ -1,76 +1,664 @@
 ---
 Attribute:
-  remediation_points: 500
-  content: Dummy content
+  remediation_points: 250_000
+  content: |
+    A class that publishes a setter for an instance variable invites client classes to become too intimate with its inner workings, and in particular with its representation of state.
+
+    The same holds to a lesser extent for getters, but Reek doesn't flag those.
+
+    ## Example
+
+    Given:
+
+    ```Ruby
+    class Klass
+      attr_accessor :dummy
+    end
+    ```
+
+    Reek would emit the following warning:
+
+    ```
+    reek test.rb
+
+    test.rb -- 1 warning:
+      [2]:Klass declares the writable attribute dummy (Attribute)
+    ```
 BooleanParameter:
-  remediation_points: 500
-  content: Dummy content
+  remediation_points: 500_000
+  content: |
+    `Boolean Parameter` is a special case of `Control Couple`, where a method parameter is defaulted to true or false. A _Boolean Parameter_ effectively permits a method's caller to decide which execution path to take. This is a case of bad cohesion. You're creating a dependency between methods that is not really necessary, thus increasing coupling.
+
+    ## Example
+
+    Given
+
+    ```Ruby
+    class Dummy
+      def hit_the_switch(switch = true)
+        if switch
+          puts 'Hitting the switch'
+          # do other things...
+        else
+          puts 'Not hitting the switch'
+          # do other things...
+        end
+      end
+    end
+    ```
+
+    Reek would emit the following warning:
+
+    ```
+    test.rb -- 3 warnings:
+      [1]:Dummy#hit_the_switch has boolean parameter 'switch' (BooleanParameter)
+      [2]:Dummy#hit_the_switch is controlled by argument switch (ControlParameter)
+    ```
+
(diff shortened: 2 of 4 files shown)


troessner:
Since you're more or less just copying our files in docs how about not having this in here at all but rather adjusting the `content` [method in the codeclimate formatter](https://github.com/troessner/reek/blob/master/lib/reek/report/code_climate/code_climate_formatter.rb#L51) like this:

```Ruby
def content
  @content_for_smell_types ||= begin
    Smells::SmellRepository.smell_types.each_with_object({}) do |fully_qualified_smell_type, accumulator|
      simple_smell_name = fully_qualified_smell_type.smell_type
      path = File.expand_path("../../../../../docs/#{simple_smell_name}.md", __FILE__)
      accumulator[simple_smell_name] = File.read path
    end
  end
  @content_for_smell_types[warning.smell_type]
end
```

I tested this in the console and it works fine, the only thing missing is that you make

```Ruby
File.expand_path("../../../../../docs/#{simple_smell_name}.md", __FILE__)
```

deal with simple_smell_name's like "BooleanParameter", which should be transformed to Boolean-Parameter.
And then update the tests that are right now red in travis ;)

mrb:
@troessner Yep that sounds good, I'll take care of it.

mrb:
@troessner One caveat is that there are some things that will need changing in the original docs, e.g. relative links from one smell doc to another. I can make this change and then go through and make PRs to fix those links, for example, if that works.

troessner:
>> @troessner One caveat is that there are some things that will need changing in the original docs, e.g. relative links from one smell doc to another. I can make this change and then go through and make PRs to fix those links, for example, if that works.

Very good point, I hadn't thought about that.
Ok, then maybe it really makes sense to just copy and adapt it for now. Our docs have been pretty stable because they are quite comprehensive already.
Can you then please fix the failing tests?
I'd then merge this one ;)

troessner:
Hold on a sec.

OR we make all links in our docs absolute.

@chastell @mvz wdyt?

chastell:
> OR we make all links in our docs absolute.

Relative links feel more elegant, but if absolute links will make things easier here (and they probably will) then let’s go with them; if they bother anyone in the future we can address that then.

mvz:
> OR we make all links in our docs absolute.

Well, I kind of like being able to pick up the whole set and move them elsewhere and things just working.

Where would you want to link? Full URL on rubydoc.info?

How hard is filtering out links from the markdown?

@mrb are there any other differences between the code climate descriptions and our docs? I noticed at least the Configuration sections are missing.

TL;DR: Perhaps we can go with the current solution first and then see if we can improve things by directly re-using the existing docs?

mrb:
@troessner @mvz 

```
Perhaps we can go with the current solution first and then see if we can improve things by directly re-using the existing docs?
```

That works for me. I think leaving the configuration stuff out of the Read up is probably a good idea for now.

troessner:
Good point @mvz 
Lets go with the current set up and maybe / maybe not improve it in the future.
@mrb lets fix the failing specs then I'll happily merge this one.

mrb:
@troessner Some style nits remaining in the test run - thoughts on those?

troessner:
>>@troessner Some style nits remaining in the test run - thoughts on those?

AFAIS its just:

>> Metrics/LineLength: Line is too long

You can just cut the string we're checking for in the specs so we are below the maximum. We dont have to be that pedantic in this case ;)

mrb:
@troessner Okey dokey!

troessner:
Looks great now! One last though: Please squash all your commits into the first, then we can merge this ;)

mrb:
@troessner Done!

troessner:
Merged! Awesome - now what's next? I guess the rest is up to you now or is there anything else we need to do to start the integration?

troessner:
Actually merging CC support warrants a minor release. I'll do that later this evening (on the road right now).

mvz:
@troessner is there any new functionality in Reek itself?

mrb:
@troessner I'm testing locally now and will have a beta version on .com for you to try soon!

troessner:
>> @troessner is there any new functionality in Reek itself?

I believe so! First of all now everybody can easily deduct his own docker image from the existing one and use it for his own CI set up in whatever ways he sees fit and secondly now everybody can use this locally via codeclimate cli which is something i am doing for instance ;)
This warrant a new minor version for me - what do you think?

chastell:
…also `bundler outdated` in third-party projects will no longer grumble about that `reek` gem depending on an outdated `codeclimate-engine-rb`. :kissing_heart:

mvz:
> This warrants a new minor version for me - what do you think?

:+1:

PR #841, opened 2016-01-27
Add Ruby 2.3 to Travis

chastell:


troessner:
And travis fails. Oh, the irony :laughing:

chastell:
Yeah, it’s `2.0.0` (reasonable, that was still the old versioning schema), `2.1`, `2.2` (both reasonable, this is the new versioning schema), but `2.3.0` (because `<%= reasons %>`). Green now though.

PR #847, opened 2016-02-05
Disable UnusedPrivateMethod detector by default.

### LINKED ISSUE #844: Disable the UnusedPrivateMethod smell by default
Let disable the UnusedPrivateMethod smell by default. There are just too many case where this detector will give a false positive, e.g. #843 and #842  and there is no way we can fix it unless we rewrite the entire backbone of `Reek` to use a global class / method / whatever registry that would allow us to detect those cases.

WDYT @chastell @mvz ?

troessner:
Fixes #844 

Regarding the "re-enable a by default deactivated smell detector for tests" I took some inspiration from the corresponding Attribute smell detector PR [here](https://github.com/troessner/reek/pull/613/files)

### DIFF
--- a/spec/reek/smells/unused_private_method_spec.rb
+++ b/spec/reek/smells/unused_private_method_spec.rb
@@ -4,6 +4,12 @@ require_lib 'reek/examiner'
 require_relative 'smell_detector_shared'
 
 RSpec.describe Reek::Smells::UnusedPrivateMethod do
+  let(:configuration) do
+    test_configuration_for(
+      described_class =>
+        { Reek::Smells::SmellConfiguration::ENABLED_KEY => true }
+    )
+  end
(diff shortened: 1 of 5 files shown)


spec/reek/smells/unused_private_method_spec.rb
```diff
@@ -4,6 +4,12 @@
 require_relative 'smell_detector_shared'
 
 RSpec.describe Reek::Smells::UnusedPrivateMethod do
+  let(:configuration) do
```
troessner: Sadly enough, due to the way we load SmellDetectors and Examiners something like:

```Ruby
  before do
    allow(described_class).
      to receive(:default_config).
      and_return(described_class.superclass.default_config)
  end
````

or

```Ruby
  before do
    described_class.instance_eval { def default_config; super; end }
  end
```

but given that we don't do this every other week I'd consider this good enough given the task at hand.

Issue #583, opened 2015-07-09
Handle statements outside method definitions


mvz:
While looking at how to handle code like the following, I noticed that there is no code context where this code shows up.
```ruby
CompilationError = Class.new(RuntimeError)
```

It seems we skip all statements outside of method definitions.

mvz:
This no longer occurs but it seemed sensible to add some specs for it.

troessner:
Closing this one since #848 is merged.

PR #848, opened 2016-02-06
Top-level dynamically created classes are detected

mvz:
Demonstrates and ensures that #583 is no longer a problem.


troessner:
Nice as well, merged!

PR #849, opened 2016-02-06
Handle any receiver in SendNode#module_creation_receiver?

### LINKED ISSUE #786: undefined method `simple_name' for (send (const nil :Class) :new)
Similar to #719 and #715 

The following produces an error when reek attempts to parse it:

```ruby
Foo = Class.new.new
```

I checked this bug against the latest revision of master (28e8b2f).

mvz:
Fixes #786.


troessner:
Nice, merged!

Issue #804, opened 2015-12-29
Drop private_attr?


chastell:
Now that Ruby 2.3 no longer warns on private accessors I’d love to drop the private_attr gem.

Unfortunately, the `attr_*` calls do not return the symbols for the created accessors, so we can’t go with `private attr_reader :foo, :bar`. Even more unfortunately we seem to have a policy of keeping private accessors at the top of classes (rather than with the rest of private methods). ;)

I see three options:

1. Drop private_attr and move `attr_*` calls below the `private` keyword; I assume this is a no-go, as it goes against the policy of keeping accessors at the top of classes.
1. Start the classes with `private` followed by `attr_*` calls followed by `public`. This feels fugly, but would allow us to both keep the accessors at the top of the classes and drop private_attr.
1. Keep private_attr for the sole purpose of having non-public accessors at the top of the classes (rather than not emitting warnings on private accessors).

Thoughts?

mvz:
> Drop `private_attr` and move `attr_*` calls below the private keyword;

This has my preference. It keeps the 'public' API of the class in one place, the private in another. Option 2 is very ugly I think; option 3 is silly and would make me want to change the visibility and attribute detection code :frowning:.

troessner:
>> Drop private_attr and move attr_* calls below the private keyword; I assume this is a no-go, as it goes against the policy of keeping accessors at the top of classes.

There is no policy for that. It just sucks. Because then you **use** attributes at the top of the class but you **declare** them at the bottom. That is against your reading flow. That's why it sucks. It's counter-intuitive. And doesn't seem fitting for a language where readability >>>> everything.

>> Drop private_attr and move attr_* calls below the private keyword;

Well, thankfully we're a team of three (== odd number) which gives us the opportunity for a clear decision.
If you two feel that's better then let's go for it. I'll still think that it sucks (see above) but I'll abide by that decision :)

mvz:
> That is against your reading flow.

That is true, but then it would make sense to move **all** private methods to the top.

I think there are two ways of looking at this: Either you start out defining all the little details (supporting private methods, private attributes, etc.), and then define the 'big picture' methods (public methods), *or* you first defined the big picture methods, and then define the details ('If you're really interested, you can read the fine print here'). The first variety reads more like a mathematical paper, the second more like a report with an executive summary. The first public, then private convention tends to sort of favour the second way of laying out code.

> opportunity for a clear decision.

[...]

PR #850, opened 2016-02-06
Update dependencies

mvz:
- Update to RubuCop 0.37.0
- Fix new offenses
- Sort RuboCop configuration
- Update Aruba dependency

Just good housekeeping :smile:.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,6 +4,10 @@ AllCops:
     - 'tmp/**/*'
     - 'vendor/**/*'
 
+# FIXME: Re-enable once private_attr_reader is no longer used
+Lint/UselessAccessModifier:
+  Enabled: false
+
 # FIXME: Make the class shorter
 Metrics/ClassLength:
   Exclude:
@@ -22,21 +26,19 @@ Metrics/LineLength:
 Metrics/ParameterLists:
   CountKeywordArgs: false
 
-# Indent one level for follow-up lines
-Style/MultilineOperationIndentation:
-  EnforcedStyle: indented
+Style/AccessorMethodName:
+  Exclude:
+    - 'lib/reek/context/visibility_tracker.rb'
 
-# Allow small arrays of words with quotes
-Style/WordArray:
-  MinSize: 3
-
-# Allow single-line method definitions
-Style/SingleLineMethods:
+# TODO: Discuss
+Style/Alias:
   Enabled: false
 
-# Always use raise to signal exceptions
-Style/SignalException:
-  EnforcedStyle: only_raise
+Style/Documentation:
+  Exclude:
+    - 'lib/reek/ast/sexp_extensions/send.rb'
+    - 'lib/reek/ast/sexp_extensions/super.rb'
+    - 'lib/reek/ast/sexp_extensions/variables.rb'
 
 # Place . on the previous line
 Style/DotPosition:
@@ -50,19 +52,30 @@ Style/EmptyLineBetweenDefs:
 Style/MultilineBlockChain:
   Enabled: false
 
+# Indent one level for follow-up lines
+Style/MultilineMethodCallIndentation:
+  EnforcedStyle: indented
+
+# Indent one level for follow-up lines
+Style/MultilineOperationIndentation:
+  EnforcedStyle: indented
+
+# TODO: Discuss
+Style/MutableConstant:
+  Enabled: false
+
+# There's nothing wrong with parallel assignment
+Style/ParallelAssignment:
+  Enabled: false
+
 # Allow Perl-style references to regex matches
 Style/PerlBackrefs:
   Enabled: false
 
-Style/Documentation:
-  Exclude:
-    - 'lib/reek/ast/sexp_extensions/send.rb'
-    - 'lib/reek/ast/sexp_extensions/super.rb'
-    - 'lib/reek/ast/sexp_extensions/variables.rb'
-
-Style/AccessorMethodName:
-  Exclude:
-    - 'lib/reek/context/visibility_tracker.rb'
-
-Style/ParallelAssignment:
+# Allow single-line method definitions
+Style/SingleLineMethods:
   Enabled: false
+
+# Allow small arrays of words with quotes
+Style/WordArray:
+  MinSize: 4
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,10 +26,6 @@ Style/AccessorMethodName:
   Exclude:
     - 'lib/reek/context/visibility_tracker.rb'
 
-# TODO: Discuss
-Style/Alias:
-  Enabled: false
-
 Style/Documentation:
   Exclude:
     - 'lib/reek/ast/sexp_extensions/send.rb'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -56,10 +56,6 @@ Style/MultilineMethodCallIndentation:
 Style/MultilineOperationIndentation:
   EnforcedStyle: indented
 
-# TODO: Discuss
-Style/MutableConstant:
-  Enabled: false
-
 # There's nothing wrong with parallel assignment
 Style/ParallelAssignment:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,10 +4,6 @@ AllCops:
     - 'tmp/**/*'
     - 'vendor/**/*'
 
-# FIXME: Re-enable once private_attr_reader is no longer used
-Lint/UselessAccessModifier:
-  Enabled: false
-
 # FIXME: Make the class shorter
 Metrics/ClassLength:
   Exclude:
--- a/lib/reek/ast/node.rb
+++ b/lib/reek/ast/node.rb
@@ -109,14 +109,14 @@ module Reek
         each_sexp do |elem|
           elem.look_for_type(target_type, ignoring, &blk) unless ignoring.include?(elem.type)
         end
-        blk.call(self) if type == target_type
+        yield self if type == target_type
       end
 
       # See ".find_nodes" for documentation.
       def look_for_types(target_types, ignoring = [], &blk)
         return if ignoring.include?(type)
         if target_types.include? type
-          blk.call(self)
+          yield self
         else
           each_sexp do |elem|
             elem.look_for_types(target_types, ignoring, &blk)
--- a/Gemfile
+++ b/Gemfile
@@ -3,13 +3,13 @@ source 'https://rubygems.org'
 gemspec
 
 group :development do
-  gem 'aruba',         '~> 0.10.0'
+  gem 'aruba',         '~> 0.12.0'
   gem 'ataru',         '~> 0.2.0'
   gem 'cucumber',      '~> 2.0'
   gem 'factory_girl',  '~> 4.0'
   gem 'rake',          '~> 10.0'
   gem 'rspec',         '~> 3.0'
-  gem 'rubocop',       '~> 0.34.0'
+  gem 'rubocop',       '~> 0.37.0'
   gem 'yard',          '~> 0.8.7'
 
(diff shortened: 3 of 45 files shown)


mvz:
As part of reviewing this, please consider the items marked for discussion in the rubocop configuration.

.rubocop.yml
```diff
@@ -4,6 +4,10 @@ AllCops:
     - 'tmp/**/*'
     - 'vendor/**/*'
 
+# FIXME: Re-enable once private_attr_reader is no longer used
+Lint/UselessAccessModifier:
```
troessner: So in the latest version

```Ruby
private_attr_reader :foo
````

is marked with that?

File: .rubocop.yml [hunk 2]
```diff
@@ -22,21 +26,19 @@ Metrics/LineLength:
...
-# Allow small arrays of words with quotes
-Style/WordArray:
-  MinSize: 3
+Style/AccessorMethodName:
+  Exclude:
+    - 'lib/reek/context/visibility_tracker.rb'
 
-# Allow single-line method definitions
-Style/SingleLineMethods:
+# TODO: Discuss
```
troessner: :+1:

.rubocop.yml
```diff
@@ -46,23 +48,42 @@ Style/DotPosition:
 Style/EmptyLineBetweenDefs:
   AllowAdjacentOneLineDefs: true
 
+# Maintain compatibility with older Rubies
```
troessner: But the freeze pragma will just be ignored by older rubies, won't it?
I thought about adding the pragma everywhere as well ...

.rubocop.yml
```diff
@@ -46,23 +48,42 @@ Style/DotPosition:
...
+# Indent one level for follow-up lines
+Style/MultilineMethodCallIndentation:
+  EnforcedStyle: indented
+
+# Indent one level for follow-up lines
+Style/MultilineOperationIndentation:
+  EnforcedStyle: indented
+
+# TODO: Discuss
+Style/MutableConstant:
```
troessner: :+1:

.rubocop.yml
```diff
@@ -46,23 +48,42 @@ Style/DotPosition:
...
+  Enabled: false
+
+# TODO: Discuss
+Style/ParallelAssignment:
+  Enabled: false
```
troessner: :+1: as well. I always hated this rule.

.rubocop.yml
```diff
@@ -46,23 +48,42 @@ Style/DotPosition:
...
-    - 'lib/reek/ast/sexp_extensions/send.rb'
-    - 'lib/reek/ast/sexp_extensions/super.rb'
-    - 'lib/reek/ast/sexp_extensions/variables.rb'
-
-Style/AccessorMethodName:
-  Exclude:
-    - 'lib/reek/context/visibility_tracker.rb'
+# Always use raise to signal exceptions
+Style/SignalException:
+  EnforcedStyle: only_raise
```
troessner: :-1: I like the differentiation between `fail` and `raise`

.rubocop.yml
```diff
@@ -46,23 +48,42 @@ Style/DotPosition:
...
 
-Style/ParallelAssignment:
+# Allow single-line method definitions
+Style/SingleLineMethods:
   Enabled: false
```
troessner: :+1:

lib/reek/ast/node.rb
```diff
@@ -109,14 +109,14 @@ def look_for_type(target_type, ignoring = [], &blk)
         each_sexp do |elem|
           elem.look_for_type(target_type, ignoring, &blk) unless ignoring.include?(elem.type)
         end
-        blk.call(self) if type == target_type
+        yield self if type == target_type
```
troessner: Yep, that's nicer!

troessner:
Reviewed, looks good, couple of discussion items around rubocop and then lgtm

.rubocop.yml
mvz: No, we have `private` before each of them and that triggers this cop.

.rubocop.yml
mvz: I tried this on another project in the mean time, and I agree. I'll change the enforced style to `always`.

.rubocop.yml
mvz: Another option is to settgis to RuboCop's old default.

.rubocop.yml
mvz: This setting was only moved to sort alphabetically.

.rubocop.yml
mvz: Ah, this also was just moved, but I added the discuss bit to hopefully get some explanatory comment for it. I suppose I can just add `# @troesnerr always hated this rule` :laughing:.

troessner: :laughing:

mvz:
Updated adding string freeze comments

Rakefile
```diff
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
```
troessner: Can we pls have a newline and here and for all other files below? I really think this would increase readability.
Also, can we make this a separate commit?

mvz: > Can we pls have a newline and here and for all other files below? I really think this would increase readability.

These changes are all autogenerated by RuboCop, so I'd rather not :smile:.

> Also, can we make this a separate commit?

Sure.

.rubocop.yml
mvz: We currently only actually use `raise` in Reek. Is there any place you would like to use `fail`. As a side note, RuboCop actually changed the default to `only_raise` recently.

troessner: >> Is there any place you would like to use fail

We haven't written any "fails" / "raises" ever since we introduced rubocop :laughing: 
But yes, grepping for "raise" shows me a couple of places where I would do that. 
On the other hand, that's the same discussion we recently had in [RubyCritic](https://github.com/whitesmith/rubycritic/pull/119) so let me just quote myself (ah, i love quoting myself!!!):

>> Not the biggest fan of this since I really like the semantic difference between "fail" and "raise". 
However, sticking a reasonable style guide project wide is more important than my personal taste here.

So let's just stick with rubocop's defaults ;)

.rubocop.yml
troessner: But then we could easily fix this by moving those declarations up like [here](https://github.com/troessner/reek/blob/master/lib/reek/source/source_locator.rb#L32) and then remove this configuration setting, can't we?
If yes, I'd prefer to do this in this  PR since it's also only a couple of occurences.

mvz: Yes, it is an easy fix. My reasoning was that we will be touching the same locations again in #804, adding back the `private` keyword that would be removed now.

mvz: On third thought, I can let RuboCop fix this automatically for now.

.rubocop.yml
mvz: To expand on this a bit further, RuboCop started out enforcing `alias_method` everywhere, and this is how things stand in Reek at the moment. With the new version it changed to requiring `alias_method` inside methods, and `alias` outside. I have a slight preference for the new style, but I think we pick a style and have RuboCop enforce it either way.

.rubocop.yml
mvz: So, to be clear, you are against RuboCop enforcing use of `freeze` on literals assigned to constants?

.rubocop.yml
mvz: Ok, I removed the setting so we use RuboCop's new default (which is `only_raise` so there are no code changes).

.rubocop.yml
mvz: > I can let RuboCop fix this automatically for now.

Oddly, this smell has no autocorrect :disappointed:.

mvz:
@chastell this is the PR where I'd like your input on the RuboCop settings.

.rubocop.yml
troessner: No, I'm for it.;)

It's super-hard for me to answer those comments because I just get an email with a link from github. Which doesn't work, cause it refers to an outdated diff. So I have to click on ALL outdated diffs to find your latest comment. @github in the name of all gods available, please fix that.
Let's continue discussing in top level comments.

mvz:
> Let's continue discussing in top level comments.

:+1:

So, to sum up, the following cop settings are/were under discussion:

* Style/MutableConstant: @troessner is for it, and so am I. I wlll enable it.
* Style/Alias: RuboCop's default changed, do we follow it?
* Lint/UselessAccessModifier: Enabled now, code fixed.
* Style/ParallelAssignment: @troessner doesn't like this rule, I'd like to take a look how relevant this is.

chastell:
(Sorry, still firefighting elsewhere.)

My general ideas (I’ve been dealing with this across all of my projects, updating RuboCop in each…):

* my general rule of thumb is to err on the side of RuboCop defaults
  * which is frustrating when they change
  * so it’s tempting to just configure it to pre-change defaults
  * but in the long run, if the change is not something I deeply oppose, I do `rubocop --auto-correct` to the new defaults eventually
* I have a few pet peeves with RuboCop defaults that I consider unacceptable and disable those
  * `Style/AndOr` – `and` + `or` have a different precedence in Ruby than `&&` + `||` _on purpose_ and there’s nothing wrong with using the right ones in the right places
  * `Style/Not` – I find self-standing `not` so much more readable than `!` glued to a predicate
  * `Style/SingleLineBlockParams` – this one is my biggest issue and should be srsly burned with fire, readable block params are so much better than `{ |a, e| … }`
  * `Style/TrailingCommaIn{Arguments,Literal}` – I prefer to enforce the comma in multiline for cleaner diffs
* that said, I’m open for discussion on each and every one (although I do believe there are better uses of our Reek devel time) ;)

As for particulars here (I just realised some of them are not changes but what we have now just moved around…):

* `Style/Alias` – I’d just let RuboCop auto-correct to the current default and get used to it
* `Style/Multiline*CallIndentation` – I assume `EnforcedStyle: indented` makes our code shorter (horizontally), so I’m ok with it :)
* `Style/ParallelAssignment` – I hit a few places in my projects where parallel assignment was the most readable thing to do, so I’m ok with disabling this, but I’m also ok with keeping it on
* `Style/SingleLineMethods` – I don’t like single-line methods, but I think we use them quite often, so I understand we need to disable this :)
* `Style/WordArray` – I love `%w(small word arrays)`, but I understand it takes a bit of eye training to switch to them, so I’m ok with keeping this

chastell:
My brain also says there should be an empty line after

```ruby
# frozen_string_literal: true
```

– but if it’s just me then I’ll cope. ;)

mvz:
> I’m open for discussion on each and every one (although I do believe there are better uses of our Reek devel time) ;)

Agreed, so I'm not going to adjust any of the settings that we already have, and we can defer discussion to when we have had too many beers.

As to the two remaining settings from my list:

* Style/Alias: I agree with @chastell we should just get used to the new default and let RC update our code.
* Style/ParallelAssignment: @troessner doesn't like this rule, @chastell is neutral; I took a look and we 'violate' this cop on one occasion, where I think use of parallel assignment is perfectly acceptable.

@troessner, final thoughts on Style/Alias?

troessner:
>> @troessner, final thoughts on Style/Alias?

Yep, let's go with rubocop defaults ;)

mvz:
> but if it’s just me then I’ll cope. ;)

No, it's not just you. I moved that part to a separate PR: #856.

chastell:
LGTM! @troessner?

(Let’s get this merged so I can drop private_attr without making others rebase their PRs.) :)

troessner:
Merging this bad boy!

troessner:
>> (Let’s get this merged so I can drop private_attr without making others rebase their PRs.) :)

So looking forward to this! One dependency less!
I think in the long run we should only have `parser` as dependency.

Issue #807, opened 2015-12-30
Adding Exceptions for Memoized Instance Variables in TooManyInstanceVariables


jfelchner:
I think the purpose of the `TooManyInstanceVariables` smell is great, but there's the use case for instance variables where they're simply used to hold state for a memoized method and are typically not allowed to be set.  I think that this use case of instance variables is actually an indicator of good code.  The developer has decided that the method's results don't rely on changing state and understands the mutability of the class/method.

I'd like to add an exception to `TooManyInstanceVariables` to account for this and exclude instance variables if they are seen _only_ during a memoization (eg `@foo ||= 'bar'`).  If in other parts of the code it is read from, or if it's assigned elsewhere (eg `@foo = 'baz'`) then it would count even if it is memoized elsewhere in the code.

First, is this acceptable and second where would I go to see an example of a place to start the PR off from?  I don't have experience dealing with Ruby parsers.

troessner:
Ad 1) great shoot!
Ad 2) https://github.com/troessner/reek/blob/master/docs/How-reek-works-internally.md

troessner:
And of course https://github.com/troessner/reek#developing-reek--contributing ;)

chastell:
> I'd like to add an exception to `TooManyInstanceVariables` to account for this and exclude instance variables if they are seen _only_ during a memoization (eg `@foo ||= 'bar'`).

Oooh, this is a very good idea! :+1:

mvz:
Awesome. :+1:

troessner:
@jfelchner I assume you won't pick this up any time soon, right?

jfelchner:
@troessner correct.  I've had too much on my plate lately.  Thanks for checking in!

mvz:
I still think this is a good idea.

waldyr:
I can pick this up. But probably we should focus on the spec refactor. What do you think?

troessner:
Yes, spec refactoring first please :)

waldyr:
@jfelchner we just released version 4.4.0 which contains the requested. 

Reek should not account memoized instance variables in the TooManyInstanceVariables smell anymore.

Could you try it out and check if everything is working as you requested?

troessner:
Closing this with #1032 merged.

PR #852, opened 2016-02-07
Implement Reek todo flag.

### LINKED ISSUE #723: reek --todo
Quoting from a discussion with @chastell so that we don't forget about this:

1. (Important) Introduce a `reek --todo` (or compatible) flag that
will output a todo.reek file which silences all current smells. This
would allow people to have a policy of ‘flag all new smells, let’s
go through the legacy ones one-by-one’. RuboCop’s --auto-gen-config
worked really well for me in a few projects.

2. (Not very) Verify that whitelisted smells are still there and
yell if they’re not. RuboCop does this and the benefit is that
(a) once you fix a smell you’re prompted to drop the whitelisting
and won’t reintroduce it unknowingly and (b) if you move / rename
a method you won’t have cruft entries forever in your reek config.

troessner:
First draft to fix #723  - just a skeleton at the moment.

I'd like to discuss this skeleton to make sure that we are all on the same page here.

@chastell @mvz what do you think?


lib/reek/cli/options.rb
```diff
@@ -104,6 +96,24 @@ def set_configuration_options
...
       end
 
+      def set_generate_todo_list_options
+        parser.separator '\nGenerate a todo list:'
+        parser.on('-t', '--todo', 'Generate a todo list') do
+          self.generate_todo_list = true
+        end
+      end
+
+      def set_alternative_formatter_options
```
troessner: I just moved this method down since it was too far up before.

mvz:
Looks good. How about calling the generated file `todo.reek` or something?

troessner:
>> Looks good. How about calling the generated file todo.reek or something?

Excellent suggestion, updated it to "todo.reek".

features/generate_todo_list.feature
```diff
@@ -0,0 +1,48 @@
...
+      IrresponsibleModule:
+        exclude:
+        - Dirty
+      UncommunicativeMethodName:
+        exclude:
+        - Dirty#a
+        - Dirty#b
+      """
+
+  Scenario: Respects an configuration file
```
tuexss: `an` -> `a`

features/step_definitions/reek_todo/generate_todo_list_steps.rb
```diff
@@ -0,0 +1,37 @@
+Given(/^a directory 'lib' with one clean file 'clean\.rb' and one dirty file 'dirty\.rb'$/) do
+  write_file('lib/clean.rb', <<-EOS.strip_heredoc)
```
tuexss: still no squiggly heredoc?

lib/reek/cli/options.rb
```diff
@@ -9,7 +9,7 @@ module CLI
     #
     # See {file:docs/Command-Line-Options.md} for details.
     #
-    # :reek:TooManyInstanceVariables: { max_instance_variables: 9 }
+    # :reek:TooManyInstanceVariables: { max_instance_variables: 10 }
```
tuexss: how could this be solved? replacing the bunch of instance variables with a struct? class variables?

troessner: >> replacing the bunch of instance variables with a struct?

This wouldn't fix the smell, just defer it to somewhere else. 

>>  class variables?

That would actually make it worse since class variables are evil. Not to mention that we actually have a [smell detector](https://github.com/troessner/reek/blob/master/docs/Class-Variable.md) for it :laughing: 

One solution I had in the back of my mind was to use composition, which in this case means to have several different objects encapsulating groups of cli arguments. We could e.g. group / abstract the exit codes in one object, and the format arguments in another. 
Since I find this "put all arguments into one instance annoying and slightly confusing as well I'll probably look into fixing this sooner or later (but in another PR then).

features/step_definitions/reek_todo/generate_todo_list_steps.rb
troessner: No, because that would limit us to 2.3. And right now we're supporting all rubies [up from 2.0](https://github.com/troessner/reek/blob/master/.travis.yml#L8)

lib/reek/cli/options.rb
mvz: There's an open ticket for ignoring instance variables used for memoization: #807.

troessner: But none of the instance variables below is used for memoization @mvz :)
In any case, we should group those options. And one of these days, I'm gonna look into that.

features/generate_todo_list.feature
```diff
@@ -0,0 +1,48 @@
...
+    And the todo file looks like this:
+      """
+      ---
+      IrresponsibleModule:
+        exclude:
+        - Dirty
+      UncommunicativeMethodName:
+        exclude:
+        - Dirty#a
+        - Dirty#b
```
chastell: I love the specificity of the generated todo; very easy to tackle the issues one-by-one. :)

chastell:
I think this is exactly what we need, at least for starters (but I can’t think of anything missing).

troessner:
@mvz && @chastell ready for review!

lib/reek/cli/command.rb
```diff
@@ -1,16 +0,0 @@
-module Reek
-  module CLI
-    #
-    # Base class for all commands
-    #
-    class Command
```
troessner: I removed the inheritance and replaced it with a mixin below since a) the inheritance lead to spaghetti code pretty fast (e.g. jumping back and forth between parent and kid to follow the flow) and b) was wrong in the first place since it violated the Liskov substitution principle big time.

lib/reek/cli/command/commandable.rb
```diff
@@ -0,0 +1,35 @@
+module Reek
+  module CLI
+    module Command
+      #
+      # Base class for all commands
+      #
+      # :reek:ModuleInitialize
```
troessner: Justified here in my mind.

lib/reek/cli/command/report_command.rb
```diff
@@ -0,0 +1,28 @@
...
+      #
+      # A command to collect smells from a set of sources and write them out in
+      # text report format.
+      #
+      class ReportCommand
+        include Commandable
+
+        def execute(app)
+          scan_for_smells app
+          reporter.smells? ? app.report_smells : app.report_success
```
troessner: I'm not really happy with the [object oriented clusterfuck pattern](https://github.com/troessner/reek/issues/775) used here but this would be something for another PR since this is the way it is right now.

lib/reek/cli/command/todo_list_command.rb
```diff
@@ -0,0 +1,35 @@
...
+          scan_for_smells app
+          File.write '.todo.reek', groups.to_yaml
+          puts "\n'.todo.reek' generated! You can now use "\
+                  'this as a starting point for your configuration.'
+        end
+
+        private
+
+        def groups
+          @groups ||= reporter.smells.group_by(&:smell_type).map do |type, smells|
```
troessner: Credits to @backus for this elegant solution!

lib/reek/cli/options.rb
```diff
@@ -55,10 +57,11 @@ def color_support?
         $stdout.tty?
       end
 
-      # :reek:TooManyStatements: { max_statements: 6 }
+      # :reek:TooManyStatements: { max_statements: 7 }
```
troessner: Justified as well. I believe we should overhaul our whole options handling but this is again something for a different PR.

troessner:
Grummlllll. 
Fails under 2.0 with:

>> /Users/timo/dev/reek/lib/reek/cli/command/todo_list_command.rb:26:in `groups': undefined method `to_h' for []:Array (NoMethodError)

Seems like we have to use a not so elegant solution :(
I'm looking into it.

troessner:
And green again for 2.0 ;)

lib/reek/cli/command.rb
mvz: I'm not sure I follow how using a mixin resolves a), since you're jumping between module and class?

Does b) mean the Command pattern violates Liskov?

lib/reek/cli/command/commandable.rb
mvz: Hm, i'm not so sure. Perhaps initialize belongs in a Command superclass, while the scanner code could be in a module, or perhaps even better in a class that the commands delegate to.

lib/reek/cli/command/commandable.rb
```diff
@@ -0,0 +1,35 @@
...
+
+        attr_reader :options
+
+        def initialize(options)
+          @options = options
+        end
+
+        def scan_for_smells(app)
+          sources.each do |source|
+            reporter.add_examiner Examiner.new(source,
```
mvz: `reporter` is very confusing here since it's not defined inside this module.

lib/reek/cli/command/report_command.rb
mvz: I resolved that in #830, so maybe merge that first :smile:?

lib/reek/cli/command/todo_list_command.rb
```diff
@@ -0,0 +1,35 @@
...
+            [type, { 'exclude' => smells.map(&:context) }]
+          end
+        end
+
+        def reporter
+          @reporter ||= Report.report_class(:yaml).new
```
mvz: I suppose the `:yaml` here is not really important? Any report class will do?

lib/reek/cli/options.rb
mvz: Ah yes, I see. This needs some thinking on how this kind of problem should be handled in general.

lib/reek/cli/options.rb
mvz: Agreed.

lib/reek/cli/command.rb
troessner: >>  a), since you're jumping between module and class?

Yes. True. I still find it better readable as it is right now, but that might just be me.

>> Does b) mean the Command pattern violates Liskov?
 
Let me contrast the command pattern:

>> In object-oriented programming, the command pattern is a behavioral design pattern in which an object is used to encapsulate all information needed to perform an action or trigger an event at a later time.

with the substitution principle:

>> Substitutability is a principle in object-oriented programming. It states that, in a computer program, if S is a subtype of T, then objects of type T may be replaced with objects of type S (i.e., objects of type S may substitute objects of type T) without altering any of the desirable properties of that program (correctness, task performed, etc.). 

I think it depends on your point of view. From a "command pattern" point of view where I just execute "something" it doesn't. My angle was coming from object orientated thinking and the fact that those 2 commands are completely different. In fact, they couldn't be more different. 

But while writing this down I realize that i'm probably yak-shaving right now. This discussion would make more sense if we had dozens of different commands. 
I'll rewire the commands as separate classes.

lib/reek/cli/command/commandable.rb
troessner: Agreed, see above, I'll change it.

lib/reek/cli/command/commandable.rb
troessner: I'll change this anyway after our discussion, but out of interest, why do you find this confusing?
This is something you will encounter with the majority of all mixins, won't you?

lib/reek/cli/command/todo_list_command.rb
troessner: It's indeed not important. Actually now that I think of it this is highly confusing. 
Let me refactor this.

troessner:
Revising and revamping this PR:

- Handling "no smells found" properly including specs and features
- Getting back to command classes
- Rebasing against the latest changes

I'll ping you @mvz when this is done - pretty sure tomorrow though ;)

troessner:
@mvz ready for final review!

I addressed all comments and also increased test coverage, both in specs and features.

Lgtm as soon as possible!

README.md
```diff
@@ -337,6 +337,42 @@ end
 
 This is an incredible powerful feature and further explained under [Smell Suppresion](docs/Smell-Suppression.md).
 
+### Generating a 'todo' list
+
+Integrating tools like `Reek` into an existing larger codebase can be daunting when you have to fix
```
mvz: Just a minor nitpick: Didn't we decide to not use backticks when mentioning Reek's name, and only use them for the command?

mvz:
One final nit and it's good to go!

README.md
troessner: Changed that. We then also need a follow-up PR:

>> grep -r '`Reek`' README.md docs/ | wc -l
      12

troessner: Actually I replaced all occurences of `reek` and `Reek` with Reek in our README and our docs. Boyscout rule ;)

troessner:
Fixed the `reek` spelling. Merge? :)

mvz:
Nice cleanup of the README!

PR #856, opened 2016-02-12
Always require frozen string literal comments

mvz:
Follow-up for #850.

I decided to separate this out because:
* Frozen string literals is really a thing that only becomes relevant with Ruby 3.0
* There is no actual code change yet enforced by this cop
* We may accidentally return frozen strings from our API that may trip up downstream users
* RuboCop's autocorrect results are ugly at the moment

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -40,6 +40,13 @@ Style/DotPosition:
 Style/EmptyLineBetweenDefs:
   AllowAdjacentOneLineDefs: true
 
+# Require comment for files in lib and bin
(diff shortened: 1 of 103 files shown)


troessner:
I'm fine with merging this bad boy for /lib.
What do you think about excluding everything else? I don't see much value in doing in our spec and feature files so I'd rather not add that there.

WDYT?

mvz:
Good point, I'll update accordingly (and fix whatever is causing the build to fail).

mvz:
@troessner updated

PR #857, opened 2016-02-12
Drop private_attr

### LINKED ISSUE #804: Drop private_attr?

chastell:
Fixes #804.

@troessner: this is the method placement you (grudgingly) agreed to, but feel free to veto and propose a better solution. :)

### DIFF
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -9,10 +9,4 @@ We use instance vars only:
 
 For everything else we use proper getters / setters.
 
-If possible those should be private. We use the [private_attr](https://github.com/jswanner/private_attr) for this, e.g.:
-
-```Ruby
-private_attr_reader :configuration
-private_attr_writer :failure_message
-private_attr_accessor :examiner
-```
+If possible those should be private.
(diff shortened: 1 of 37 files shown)


troessner:
Great, merged! On the road right now, can you please release a new version?

PR #862, opened 2016-02-18
Bump version to 3.11.

troessner:


troessner:
and released it is!

mvz:
I think it's better to always include the micro version (i.e., 3.11.0 in this case).

mvz:
Other than that: \o/

troessner:
I dont find that helpful, but i'll include it next time we have a *.*.0 version :)

mvz:
Hm .. maybe I should investigate how bundler and rubygems deal with 3.11 vs 3.11.0. That way I'll have more than a hunch to back up my assertions :smile:

troessner:
I always thought the .0 was implicit for bundler / rubygems. Hopefully  i am right :D

mvz:
> Hopefully i am right :D

Turns out that yes, you are indeed right :laughing:.

troessner:
Huzzaaaaaahhhh!

Issue #863, opened 2016-02-23
Select detectors by category on the command line


mvz:
We have nice smell categories and types, but only allow users to select detectors to run by type. We should expose category as well.

mvz:
See #864 for why this is a breaking change.

mvz:
If we choose option 2 in #866 then this PR becomes superfluous.

PR #864, opened 2016-02-23
Allow smell detectors to be specified by category

mvz:
#863

### DIFF
--- a/features/command_line_interface/options.feature
+++ b/features/command_line_interface/options.feature
@@ -41,7 +41,8 @@ Feature: Reek can be controlled using command-line options
 
       Configuration:
           -c, --config FILE                Read configuration options from FILE
-              --smell SMELL                Detect smell SMELL (default: all enabled smells)
+              --smell SMELL                Detect smell or smell category SMELL
+                                             (default: all enabled smells)
 
       Generate a todo list:
           -t, --todo                       Generate a todo list
--- a/lib/reek/cli/options.rb
+++ b/lib/reek/cli/options.rb
@@ -92,7 +92,9 @@ module Reek
           raise ArgumentError, "Config file #{file} doesn't exist" unless File.exist?(file)
           self.config_file = Pathname.new(file)
         end
-        parser.on('--smell SMELL', 'Detect smell SMELL (default: all enabled smells)') do |smell|
+        parser.on('--smell SMELL',
+                  'Detect smell or smell category SMELL',
+                  '  (default: all enabled smells)') do |smell|
           smells_to_detect << smell
         end
       end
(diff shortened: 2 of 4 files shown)


features/command_line_interface/options.feature
```diff
@@ -41,7 +41,8 @@ Feature: Reek can be controlled using command-line options
 
       Configuration:
           -c, --config FILE                Read configuration options from FILE
-              --smell SMELL                Detect smell SMELL (default: all enabled smells)
+              --smell SMELL                Detect smell or smell category SMELL
```
troessner: >> Detect smell or smell category SMELL

Can we remove the SMELL here? Seems unnecessary and reads pretty bad IMHO.

lib/reek/cli/options.rb
```diff
@@ -92,7 +92,9 @@ def set_configuration_options
           raise ArgumentError, "Config file #{file} doesn't exist" unless File.exist?(file)
           self.config_file = Pathname.new(file)
         end
-        parser.on('--smell SMELL', 'Detect smell SMELL (default: all enabled smells)') do |smell|
+        parser.on('--smell SMELL',
+                  'Detect smell or smell category SMELL',
```
troessner: Same here.....

spec/reek/smells/smell_repository_spec.rb
```diff
@@ -19,4 +19,21 @@
...
+    it 'returns smell detectors specified by type' do
+      result = described_class.eligible_smell_types(['IrresponsibleModule', 'TooManyStatements'])
+      expect(result).to eq [Reek::Smells::IrresponsibleModule,
+                            Reek::Smells::TooManyStatements]
+    end
+
+    it 'returns smell detectors specified by category' do
+      result = described_class.eligible_smell_types(['Duplication'])
+      expect(result).to eq [Reek::Smells::DuplicateMethodCall]
+    end
```
troessner: Nice specs, clean and concise.

features/command_line_interface/options.feature
```diff
@@ -41,7 +41,8 @@ Feature: Reek can be controlled using command-line options
...
+                                             (default: all enabled smells)
```
troessner: Also, can we add some examples for both (smell types and smell categories) to our README including helpful links to our docs so the user can explore that on its own?
If we just put it out like this I believe the mental barrier for the user is so high that nobody except for maybe us will use it...

troessner:
2 small comment and then lgtm!

features/command_line_interface/options.feature
mvz: Well, we need to put something to indicate that a parameter is required. Suggestions?

troessner: Hmm, how about adding <>, so <SMELL> to make it more obvious?
Right now it reads like a typo.

mvz:
I was updating the documentation and noticed that LongParameterList is both a smell type and a smell category that also includes LongYieldList. We should fix that before exposing categories like this (since there would be no way to only detect the *smell* LongParameterList).

troessner:
This one is now obsolete with #867 isn't it?

mvz:
Yes, this one is obsolete.

PR #870, opened 2016-02-27
Remove raise hack for default args

troessner:
Precondition for this one is #869

PR #872, opened 2016-02-27
Clean up Examiner constructor.

### LINKED ISSUE #650: Breaking change for reek 4: Clean up Examiner constructor

troessner:
Fixes #650, has #869 as precondition (specs and features pass, but rubocop complains)

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -32,7 +32,7 @@ source = <<-EOS
 EOS
 
 reporter = Reek::Report::TextReport.new
-examiner = Reek::Examiner.new(source)
+examiner = Reek::Examiner.new source
 reporter.add_examiner examiner
 reporter.show
 ```
@@ -175,7 +175,7 @@ source = <<-END
   end
 END
 
-examiner = Reek::Examiner.new(source)
+examiner = Reek::Examiner.new source
 examiner.smells.each do |smell|
   puts smell.message
 end
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -27,7 +27,7 @@ module Reek
     #
     # @public
     def initialize(source,
-                   filter_by_smells = [],
+                   filter_by_smells: [],
                    configuration: Configuration::AppConfiguration.default)
       @source           = Source::SourceCode.from(source)
       @collector        = CLI::WarningCollector.new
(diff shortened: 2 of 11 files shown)


lib/reek/examiner.rb
```diff
@@ -25,8 +25,8 @@ class Examiner
     #   The configuration for this Examiner.
     #
     # @public
-    def initialize(source,
-                   filter_by_smells = [],
+    def initialize(source:,
+                   filter_by_smells: [],
```
mvz: I'm not sure how I feel about making the main argument a keyword argument as well. It seems a little superfluous. Any reasons making all arguments keyword arguments would be better?

troessner: You're right. On second thought I realize that this would make our API a little more verbose and for no apparent reason. 
I'll revert this change.

troessner:
Addressed @mvz comment regarding the first argument. PR will be green after we merged #870 ;)

docs/API.md
```diff
@@ -32,7 +32,7 @@ source = <<-EOS
 EOS
 
 reporter = Reek::Report::TextReport.new
-examiner = Reek::Examiner.new(source)
+examiner = Reek::Examiner.new source
```
chastell: This is a super minor nitpick, but while I used to preach the Seattle-style ‘all parentheses must go!’ approach in the past I now got used to – and like _very much_ – the idea of keeping the parentheses when returning a value. I also think that keeping the parentheses here would make this a tiny bit clearer/unambiguous, which would be most welcome in API docs. :)

chastell:
One (easily veto-able) nitpick from me and this is good to go!

docs/API.md
troessner: But then we'd should update the rest of our api docs as well. Personally this feels a little like yak-shaving to me i have to say :laughing: 
But if you and @mvz feel that this communicates our intention clearer I'll happily change. 
But right now, I'd rather veto this change ;)

mvz: I prefer leaving parentheses out when possible in method calls.

troessner: Ok, veto complete :laughing: 

Now, of to serious internet bussiness: Can i get a merge pls?

PR #874, opened 2016-02-27
Introduce `private_constant`.

troessner:
I started out with the obvious cases which are inner classes. (And yes, Ruby has them, and no, i don't care that they mean something different in other languages ;) )
Note that there are quite a few other classes where we should use `private_constant` as well like DataClump. But there the soon-to-be private classes are at package level, not scoped to the actual class that uses them. 
So I'd like to refactor those classes (especially DataClump!) before introducing `private_constant` there.

### DIFF
--- a/lib/reek/smells/control_parameter.rb
+++ b/lib/reek/smells/control_parameter.rb
@@ -90,6 +90,8 @@ module Reek
         attr_reader :occurences, :param
       end
 
+      private_constant :FoundControlParameter
+
       # Finds cases of ControlParameter in a particular node for a particular parameter
       class ControlParameterFinder
         CONDITIONAL_NODE_TYPES = [:if, :case, :and, :or].freeze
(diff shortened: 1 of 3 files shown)


lib/reek/smells/control_parameter.rb
```diff
@@ -192,6 +196,8 @@ def find_matches(param)
           ControlParameterFinder.new(context.exp, param).find_matches
         end
       end
+
+      private_constant :ControlParameterCollector
```
mvz: Any thoughts on clustering these all at the end vs. putting the right after the class?

mvz:
I have no experience with `private_constant`, but the code looks good. Anything we should be careful about? @chastell, any thoughts?

chastell:
I started using it in my projects recently and in general I’m very :+1: on it. :)

I’d rather we keep it right after the class it relates to (like in this PR), as it’s clearer to me that they’re related.

(I’d rather have them before classes (ideally as `private_constant class Foo`), but – unlike `def` for methods – the `class` keyword does not return the class’s symbol.)

PR #875, opened 2016-02-28
Integrate mutant in our CI build.

troessner:
WIP PR - i wanted to discuss this first before going further.

* This runs mutant against everything we marked as public api (except for version.rb) comparing it to the master branch (--since master)
* Speed: In case there are no changes of our subjects in regards to the master branch the whole mutant run takes less than 3 seconds on my machine

Missing:

* Proper rake task integration, right now it's just a wall of hard to read text in our travis configuration

cc @mbj - what do you think of this set up?

P.S.: The travis failure come from ruby 2.0 which will be gone [pretty soon anyway](https://github.com/troessner/reek/pull/869)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ bundler_args: --without debugging
 script:
 - bundle exec rake
 - bundle exec ataru check
+- bundle exec mutant --include lib --require reek --use rspec --since master --jobs 4 "Reek*"
 rvm:
   - 2.1
   - 2.2
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ group :development do
   gem 'rubocop',       '~> 0.37.0'
   gem 'yard',          '~> 0.8.7'
   gem 'simplecov',     '~> 0.11.1'
+  gem 'mutant-rspec',  '~> 0.8.8'
 
   platforms :mri do
     gem 'redcarpet', '~> 3.3.1'


mbj:
My thoughts:

### Scope

I do think limiting mutant to the classes that implement public API is not giving you the mileage you expect. If you have private classes and a mutation cannot be killed on these via the specs for the public API - you definitively wanna know

### Whitelisting instead of blacklisting subjects

You currently list the public API match expressions to limit the scope as discussed above, while I do not agree on limiting the scope, the implementation of the limiting of the scope can be conceptually improved via "blacklisting". The idea is that you specify a namespace matcher `Reek*` and many exclusions via `--ignore-subject Reek::CurrentlyNotInScope*`. This way all additions to the `Reek` namespace automatically get included into the mutant run unless otherwise stated otherwise via an `--ignore-subject` flag.

The idea is that if you intend to "draw a line in the sand and mutation cover everything from now on" you do not have to maintain the whitelist, and can work on reducing the blacklist.

Where chances are high the need addition to the whitelist might be forgotten, and the amount of mutation covered subjects never grows.

troessner:
Thanks for the detailed explanation, Markus!

Makes sense to me - I'll update this PR to switch from a whitelisting approach to a blacklisting one.

Gemfile
```diff
@@ -12,6 +12,8 @@ group :development do
   gem 'rubocop',       '~> 0.37.0'
   gem 'yard',          '~> 0.8.7'
   gem 'simplecov',     '~> 0.11.1'
+  gem 'mutant',        '~> 0.8.10'
```
mbj: This should not be needed. It'll get pulled to the latest compatible version by `mutant-rspec` anyway.

troessner: Good point. I wanted to be super-explicit here but on second thought I realize this is questionable, I'll remove this line.

troessner:
Ready for review!

A couple of notes:

- I took @mbj advice to go with whitelisting instead of blacklisting
- Right now we only call mutant on travis so it's not part of our local build. 
- I suggest to give mutant a shot with the way it is integrated right now and see how this goes and if this works for us in general. If it does, I'd come up with a wrapper rake task that would integrate it in our local build properly
- time: At the moment the mutant run takes 5 seconds on my pretty mediocre two core macbook. I think that's ok even for our local build later

mvz:
Sounds good to me.

mvz:
I'll let @chastell have a say before merging.

.travis.yml
```diff
@@ -5,6 +5,7 @@ bundler_args: --without debugging
 script:
 - bundle exec rake
 - bundle exec ataru check
+- bundle exec mutant --include lib --require reek --use rspec --since master "Reek*"
```
mbj: You probably want to limit concurrency on CI. Mutants default is to spawn as many jobs as processors seen in `/proc/cpuinfo`, on CI environments like Travis this does not reflect the containers resource allocation.

I suggest adding `--jobs 4` for this project, 4 jobs had been proven a reasonable value in environments like Travis.

troessner: Thanks for the tip, updated!

chastell:
Apologies for lagging – I’m all for it, thanks for merging!

PR #878, opened 2016-03-02
Issue with unused param detection.

franciscoj:
I've found an edge case for the UnusedParam smell. In a case like this:

```ruby
def with_retries(tries = 3)
  yield
rescue
  retry if tries =- 1 > 0
  raise
end
```

It detects `tries` as an unused param while it is used in the rescue. Is this the expected behavior? Or is it a false positive?

Kudos for such an amazing and useful gem.


sauloperez:
This looks as a false positive to me as well

troessner:
Yep, that's a bug in Reek :)

Thanks for reporting it, pretty sure we'll fix this soon.

troessner:
Closing this one since the fix is up in #881 - thanks for the failing spec and the report!

franciscoj:
@troessner thanks to you for the fix!

PR #879, opened 2016-03-02
codebeat badge

korzonek:
Is it fine to add codebeat badge to README? 

codebeat is automated code review tool for Swift, Ruby & Go that helps get instant feedback on code quality. 

"Quick wins" suggested by codebeat could be a nice candidate for a pull request and help other developers become contributors.

FYI. To be fully open and honest. I'm co-founder of codebeat

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@
 [![Gem Version](https://badge.fury.io/rb/reek.svg)](https://badge.fury.io/rb/reek)
 [![Dependency Status](https://gemnasium.com/troessner/reek.png)](https://gemnasium.com/troessner/reek)
 [![Inline docs](https://inch-ci.org/github/troessner/reek.png)](https://inch-ci.org/github/troessner/reek)
+[![codebeat](https://codebeat.co/badges/42fed4ff-3e55-4aed-8ecc-409b4aa539b3)](https://codebeat.co/projects/github-com-troessner-reek)
 
 ## Quickstart
 


troessner:
Not the biggest fan of dozens of badges but on the other hand it's interesting to see how other's are handling code quality. Talking about [this](https://codebeat.co/projects/github-com-troessner-reek)

So from my POV that's ok to merge. @mvz && @chastell wdyt?

mvz:
I'm a bit reluctant since if we start with codebeat, we might as well add codeclimate and pullreview badges.

It might make more sense to add this and other online tools to the *Brothers and Sisters* section in the README.

korzonek:
It would be super!

chastell:
I’m not a fan of having a plethora of badges in other projects, _but_ in Reek’s case badges _from code quality sites_ are very much inlined with what Reek is all about (and where it is used…), so I’m all for adding this one.

troessner:
>> I'm a bit reluctant since if we start with codebeat, we might as well add codeclimate and pullreview badges.

Thinking about it, why not?
I believe those integrations are actually something where Reek really shines.

>> It might make more sense to add this and other online tools to the Brothers and Sisters section in the README.

But it doesnt fit in there if you look at the section right now. It would rather fit in a new section called "nieces and nephews" :laughing: 

>> I’m not a fan of having a plethora of badges in other projects, but in Reek’s case badges from code quality sites are very much inlined with what Reek is all about (and where it is used…), so I’m all for adding this one.

Agreed. @mvz veto or ok for you?

mvz:
Sure, let's go for it and make a mental note to add badges for code climate etc. as well.

PR #880, opened 2016-03-04
Fix UnusedParameter detector bug for rescue.

troessner:
Fixes #878

PR #881, opened 2016-03-05
Fix UnusedParameter detector for lvasgn.

troessner:
Same as #880 but now with more appropriate branch name and commit message.
Fixes #878

### DIFF
--- a/lib/reek/context/method_context.rb
+++ b/lib/reek/context/method_context.rb
@@ -21,7 +21,21 @@ module Reek
       end
 
       def uses_param?(param)
-        local_nodes(:lvar).find { |node| node.var_name == param.to_sym }
+        # local_nodes(:lvasgn) catches:
+        #   def foo(bar); bar += 1; end
+        # In this example there is no `lvar` node present.
+        #
+        # local_nodes(:lvar) catches:
+        #   def foo(bar); other(bar); end
+        #   def foo(bar); tmp = other(bar); tmp[0]; end
+        #
+        # Note that in the last example the `lvar` node for `bar` is part of an `lvasgn` node for `tmp`.
+        # This means that if we would just search for [:lvar, :lvasgn]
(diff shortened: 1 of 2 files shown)


lib/reek/context/method_context.rb
```diff
@@ -21,7 +21,21 @@ def references_self?
...
+        # local_nodes(:lvar) catches:
+        #   def foo(bar); other(bar); end
+        #   def foo(bar); tmp = other(bar); tmp[0]; end
+        #
+        # Note that in the last example the `lvar` node for `bar` is part of an `lvasgn` node for `tmp`.
+        # This means that if we would just search for [:lvar, :lvasgn]
+        # (e.g. via Reek::AST::Node#find_nodes) this would fail for this example since we would
+        # stop at the `lvasgn` and not detect the contained `lvar`.
+        # Hence we first get all `lvar` nodes followed by all `lvasgn` nodes.
+        #
```
mvz: This seems like an awfully large amount of commentary for a one-liner. Perphaps you can fix `local_nodes` to do the right thing, or turn the examples into specs?

troessner: >> This seems like an awfully large amount of commentary for a one-liner.

I disagree. In my mind a method should document and explain everything that is not self-explanatory. And the specs then make sure, that it works as expected. That is kind of a documentation, but not the same thing.

>> Perphaps you can fix local_nodes

We could make local_nodes take an array of node types, but I believe that's not entirely trivial since it would require to fix some of the "plumbing" for `local_nodes` as well and that would reach down deep into the core of Reek. Not opposed to that, but this should certainly be a separate endeavour. 

Long story short: I believe it should be like it is, I think the long documentation here is legit.
@chastell what's your take on this?

mvz: OK, I've been planning to clean up our node finders for a long time, I'll revisit this then.

Issue #871, opened 2016-02-27
Let's get Reek 4 going!


troessner:
We need to merge review and merge everything which is labeled as [breaking change](https://github.com/troessner/reek/pulls?q=is%3Apr+is%3Aopen+label%3Areek-4-breaking-change)

mvz:
Yes! There's an old reek-4 branch that we should look at as well. It can probably be rebased!

troessner:
>> Yes! There's an old reek-4 branch that we should look at as well. It can probably be rebased!

Done! There was just one commit from @chastell - I cherrypicked this one into a new branch and created #873 out of it.

troessner:
Summary of the status quo:

#864 - obsolete, isn't it?
#869 - needs review
#870 - needs review. Fails only because of #869 (specs and features pass, just rubocop complains because of 2.0)
#872 - needs review. Fails only because of #869 (specs and features pass, just rubocop complains because of 2.0)
#867 - reviewed. Just some smaller changes and ready to merge.
#873 - was reviewed a long time ago already, i mereley cherrypicked it from the reek-4 (which didn't contain anything else) into a fresh one with a proper name

@mvz && @chastell waiting for your reviews ;)

mvz:
Merged #869 and #873.

mvz:
How about releasing at least one `.pre` version to test this thing in the wild? I'd like to iron out anything bad before we're stuck with it for a whole major release.

troessner:
>> How about releasing at least one .pre version to test this thing in the wild? I'd like to iron out anything bad before we're stuck with it for a whole major release.

Great idea! Let's do that after we merged all breaking changes.

chastell:
Food for thought: do we want to keep Ruby 2.1 compatibility, or just go with Ruby 2.2+ for Reek 4?

Pros for keeping Ruby 2.1:
* it’s currently still maintained, which is a simple answer to ‘why we drop 2.0 and keep 2.1’,
* it’s better to support as many Ruby versions as possible/convenient,
* technically there might be projects that use it.

Pros for going Ruby 2.2+ only:
* `Binding#local_variables`, `Binding#receiver`, `Method#super_method` might be useful to us,
* `Kernel#itself`, `Method#curry` and `String#unicode_normalize` might be nice-to-haves,
* (`Enumerable#slice_after` and `Enumerable#slice_when` probably less so,)
* it’s mature enough to expect people to upgrade to it,
[...]

PR #882, opened 2016-03-05
Add Reek 4 breaking changes to README.

troessner:
Part of #871 and should be merged after we officially released Reek 4.
You know, make it easy for the user ;)


mvz:
Shouldn't that be part of the change log? We can improve the presentation there, of course, and point people there from the README.

troessner:
Yeah, on second thought I realize you're right. What we have in our changelog already pretty much says it all. Closed.

PR #884, opened 2016-03-05
Add rb-readline as development dependency.

troessner:
Solves [pry history not working correctly under macos](https://github.com/guard/guard/wiki/Add-Readline-support-to-Ruby-on-Mac-OS-X#option-4-using-a-pure-ruby-readline-implementation)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,8 @@ group :development do
 end
 
 group :debugging do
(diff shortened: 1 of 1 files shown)


mvz:
Sure, but let's put it in the same group as pry and maybe give it a comment or commit message that explains why it's there.

troessner:
>> Sure, but let's put it in the same group as pry and maybe give it a comment or commit message that explains why it's there.

Good point, updated!

Issue #816, opened 2016-01-05
Using `exclude` in the Reek config does not work for the UnusedPrivateMethod detector


troessner:
Seems like we shipped a bug in the UnusedPrivateMethod detector that went unnoticed - apparently using `exclude` in the config doesn't work:

```
  ( $ ) cat unused_private_method.reek 
---
UnusedPrivateMethod:
  exclude:
    - "Foo#bar"

  ( $ ) cat foo.rb 
class Foo
  private
  def bar; end
end

  ( $ ) reek foo.rb 
  [3]:UnusedPrivateMethod: Foo has the unused private instance method `bar` [https://github.com/troessner/reek/blob/master/docs/Unused-Private-Method.md] 
```

Disabling via source code comment works though.
I'll look into this.

chastell:
> Disabling via source code comment works though.

I was _just_ about to open an issue – no, [it doesn’t](https://github.com/chastell/lovely_rufus/commit/23ea8b6bd4d4a4a62a6473322a5d826c26225b18)…

troessner:
It does, my friend, it does!
But on a class level, not a method level.
Please see the top of ContextBuilder for an example. We probably should make this clearer in the docs.;)

troessner:
https://github.com/troessner/reek/blob/master/lib/reek/context_builder.rb#L20

chastell:
Oooh. Is there a reason why we don’t support this on a per-method basis? (See #817 for a use-case.)

troessner:
>> Oooh. Is there a reason why we don’t support this on a per-method basis? (See #817 for a use-case.)

You could make the argument that a "UnusedPrivateMethod" already screams "class scope" and not "method scope" ;)

Apart from that disabling by regex would be useless which is an incredible powerful feature. Just imagine how shitty [this](https://github.com/troessner/reek/blob/master/lib/reek/context_builder.rb#L20) would look like on a method base. You'd have to write 25 identical source code comments. Or disable it completely. Not cool, don't you agree? ;)

troessner:
BTW our official docs mention the "disable via comment on a class level" [explicitly](https://github.com/troessner/reek/blob/master/docs/Unused-Private-Method.md#configuration) so it's not like it's a secret :laughing:

chastell:
[...]

PR #885, opened 2016-03-06
Fix `disable UnusedPrivateMethod via `exclude` in configuration.

troessner:
Getting started with fixing #816

### DIFF
--- a/lib/reek/smells/unused_private_method.rb
+++ b/lib/reek/smells/unused_private_method.rb
@@ -78,9 +78,14 @@ module Reek
       # @param ctx [Context::ClassContext]
       # @return [Boolean]
       #
+      # :reek:FeatureEnvy
       def ignore_method?(ctx, method)
-        ignore_methods = value(EXCLUDE_KEY, ctx, DEFAULT_EXCLUDE_SET)
-        ignore_methods.any? { |ignore_method| method.name[ignore_method] }
+        # ignore_contexts will be e.g. ["Foo::Smelly#my_method", "..."]
+        ignore_contexts = value(EXCLUDE_KEY, ctx, DEFAULT_EXCLUDE_SET)
+        ignore_contexts.any? do |ignore_context|
+          full_name = "#{method.parent.full_name}##{method.name}"
+          full_name[ignore_context]
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


troessner:
Ready for review!

mvz:
Damn, I remember looking at this on my phone, thinking I need to check this since I think we're missing something. Then, I forgot all about it. I'll have a look now.

lib/reek/smells/unused_private_method.rb
```diff
@@ -79,8 +79,12 @@ def unused_private_methods(ctx)
       # @return [Boolean]
       #
       def ignore_method?(ctx, method)
-        ignore_methods = value(EXCLUDE_KEY, ctx, DEFAULT_EXCLUDE_SET)
-        ignore_methods.any? { |ignore_method| method.name[ignore_method] }
+        # ignore_contexts will be e.g. ["Foo::Smelly#my_method", "..."]
+        ignore_contexts = value(EXCLUDE_KEY, ctx, DEFAULT_EXCLUDE_SET)
+        ignore_contexts.any? do |ignore_context|
+          method_from_context = ignore_context.split('#').last
```
mvz: Aren't you ignoring the whole class name part? AFAICS, this will make Reek ignore `Smelly#foobar` even if you put `Awesome#foobar` here.

lib/reek/smells/unused_private_method.rb
```diff
@@ -79,8 +79,12 @@ def unused_private_methods(ctx)
...
+          method.name[method_from_context]
```
mvz: This will also match on a substring. Is this what we want?

lib/reek/smells/unused_private_method.rb
troessner: Ah, that might be :laughing: 
I'll add a "scenario" in the feature above that adds another class with a method present in smelly.rb as well and then we'll see if the warning for this is also swallowed. Then let me think about a better solution ;)

lib/reek/smells/unused_private_method.rb
troessner: Please see my updated version of this.

troessner:
Ready for final review!
`mutant` is red again which is ironic given that I increased spec and feature coverage significantly. 
So let's ignore that.

LGTM?

mvz:
\o/

PR #887, opened 2016-03-14
[WIP] Fix `accept` configuration for single values.

### LINKED ISSUE #886: UncommunicativeParameterName claims to allow `accept: /<regex>/`, doesn't
according to https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Parameter-Name.md#configuration, the UncommunicativeParameterName smell should allow a specification like:

```yaml
UncommunicativeParameterName:
  accept: /^[a|b]$/
```

unfortunately, when i do that (OSX, ruby 2.0.0p481, reek-3.11), i get an error:
```
reek -c .reek ./**/*.rb
/Users/conor/.rvm/gems/ruby-2.0.0-p481/gems/reek-3.11/lib/reek/smells/uncommunicative_parameter_name.rb:64:in 
`acceptable_name?': undefined method `any?' for "/^[a|b]$/":String (NoMethodError)
```

since i just added some UncommunicativeVariableName expectations in the same way, i was a bit confused. 

i assumed that these two smells would be implemented similarly, but they are not. 

there are 2 problems as i see it:
  * the return coming out of `accept_names(context)` is not iterable, so `any?` cannot be used
  * https://github.com/troessner/reek/blob/master/lib/reek/smells/uncommunicative_parameter_name.rb#L61 explicitly calls out `accept_names` and `reject_patterns`, and then uses `==` and `match` for comparison

until i send a PR/someone else fixes this, i'm working around this problem by specifying my UncommunicativeParameterName configuration as:
```yaml
UncommunicativeParameterName:
  accept:
    - a
    - b
```

troessner:
Preliminary PR, fixes #886

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ tags
 .tags*
 .DS_Store
 .idea/
+.rbenv-gemsets
--- a/docs/Uncommunicative-Method-Name.md
+++ b/docs/Uncommunicative-Method-Name.md
@@ -20,5 +20,36 @@ Reek's Uncommunicative Method Name detector supports the [Basic Smell Options](B
 
 | Option         | Value       | Effect  |
 | ---------------|-------------|---------|
-| `reject` | array of regular expressions | The set of regular expressions that Reek uses to check for bad names. Defaults to `[/^[a-z]$/, /[0-9]$/, /[A-Z]/]`. |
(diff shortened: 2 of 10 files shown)


troessner:
Branch name doesn't fit anymore, closing and re-opening it in #890

Issue #889, opened 2016-03-14
Validate our configuration across Reek


troessner:
Coming from #886 we should validate our configuration, this means yaml config, hash via api call and via source comment.

We have 3 options:

1) Ideal world scenario: Validate all the things
- Validate our yaml config with yaml schema (so we can even point to specific lines in the config file, something that is impossible later after it got transformed into a hash)
- Validate AppConfiguration separately since that might be passed via API users
- Remove all hacks in subsequent SmellDetectors. I didn't check but I assume we have a couple of occurences

2) Validate only hashes:
- Validate only the hash we get after loading the yaml, not the yaml itself
- The same validation could then be used for the hash we get from api users
- We might be able to use something like Classy Hash or similar

3) Move all validation and transformation rules into the AppConfiguration

Configuration via source code comments is probably something we need to do separately.

troessner:
Connected to #1062

troessner:
Continuing in #1317 - closing this one

PR #890, opened 2016-03-15
Fix `accept` and `reject` configuration handling.

### LINKED ISSUE #886: UncommunicativeParameterName claims to allow `accept: /<regex>/`, doesn't

troessner:
Preliminary PR, fixes #886, successor of #887 (which had a bad branch name in the end)
What does this bad boy do?
Glad you asked!

* consistent handling of `reject` and `accept`
* makes both settings handle strings and regexes properly
* makes both settings handle list and single values
* tests!
* Updates and improves our documentation

Tomorrow I'll make some final tweaks and then this is ready for review!

### DIFF
--- a/lib/reek/smells/uncommunicative_method_name.rb
+++ b/lib/reek/smells/uncommunicative_method_name.rb
@@ -16,18 +16,19 @@ module Reek
     # Currently +UncommunicativeMethodName+ checks for
     # * 1-character names
     # * names ending with a number
+    # * names containing a capital letter (assuming camelCase)
     #
     # See {file:docs/Uncommunicative-Method-Name.md} for details.
     class UncommunicativeMethodName < SmellDetector
       REJECT_KEY = 'reject'.freeze
       ACCEPT_KEY = 'accept'.freeze
       DEFAULT_REJECT_PATTERNS = [/^[a-z]$/, /[0-9]$/, /[A-Z]/].freeze
-      DEFAULT_ACCEPT_NAMES = [].freeze
+      DEFAULT_ACCEPT_PATTERNS = [].freeze
 
       def self.default_config
         super.merge(
           REJECT_KEY => DEFAULT_REJECT_PATTERNS,
-          ACCEPT_KEY => DEFAULT_ACCEPT_NAMES
+          ACCEPT_KEY => DEFAULT_ACCEPT_PATTERNS
         )
       end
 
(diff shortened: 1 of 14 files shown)


troessner:
Ladies and gentlemen, this little rascal is ready for review!
+353 −35 
I know it's scary but most of the additions are missing tests ;)

troessner:
Well, well, well. Mutant fails :laughing: 
I'll add specs but this thing is still reviewable.

lib/reek/smells/uncommunicative_method_name.rb
```diff
@@ -49,17 +50,17 @@ def inspect(context)
 
       private
 
-      def acceptable_name?(name:, context:)
-        accept_names(context).any? { |accept_name| name == accept_name } ||
-          reject_patterns(context).none? { |pattern| name.match pattern }
+      def acceptable_pattern?(name:, context:)
```
mvz: I would keep the name `acceptable_name?`, since that's what you're checking.

`accept_patterns` etc. are all good name-wise.

mvz:
Looks good, just one small nit.

Do you want to move these methods into a module, or do you want to move straight to implementing #889?

I was going to comment on only having cucumber scenarios for this change, but mutant did it for me :smile:.

lib/reek/smells/uncommunicative_method_name.rb
troessner: Good point, fixed in all 3 affected smell detectors.

mvz:
Regarding the mutant output, I'm wondering about this one:

```
-  super.merge(REJECT_KEY => DEFAULT_REJECT_PATTERNS, ACCEPT_KEY => DEFAULT_ACCEPT_PATTERNS)
+  super().merge(REJECT_KEY => DEFAULT_REJECT_PATTERNS, ACCEPT_KEY => DEFAULT_ACCEPT_PATTERNS)
```

For a method without any parameters, this is not really a meaningful change. I suppose we'll see if it remains once the other items have been taken care of.

troessner:
Well, well, well. To quote the current youth: FML. That looks like I need to add a gazillion specs to make this pass not to mention that I need to refactor the specs for all 3 detectors because at the moment the look like crap.

-> wip, pls stand by.

troessner:
Ready for final review!

mutant is still red, but never mind. I improved cucumber and spec coverage for the `accept` and `reject` feature by roughly 5000%. If you check out the PR you'll see that it adds hundreds of lines of test coverage to something that was literally untested before.

I feel that this is more than enough.

Let's get this merged asap.

Issue #891, opened 2016-03-16
Blocker for Reek 4: keyword arguments for your spec matchers


troessner:
We almost forgot about this: https://github.com/troessner/reek/blob/master/lib/reek/spec/should_reek_of.rb#L14
We should fix it before releasing Reek 4.

troessner:
and let's not forget about: https://github.com/troessner/reek/blob/master/lib/reek/configuration/app_configuration.rb#L52
But that one should be trivial.

mvz:
@troessner I think this can be closed?

troessner:
Indeed ;)

PR #892, opened 2016-03-16
Remove deprecated method `from_map` from AppConfiguration.

troessner:
Fixes the second part of #891 
Speedy review would be awesome since this blocks my rspec refactoring necessary for mutant necessary for #890


troessner:
I believe the changes are non-controversial and this is blocking my second part of #891 so I'm merging this myself.

Issue #606, opened 2015-07-19
Add .with_config option to RSpec matchers


mvz:

troessner:
@mvz is it realistic that we tackle this?
The last couple of months have shown that our current solution works pretty well.

I'd close this issue, if this is ok for you....

mvz:
Maybe just tag it `maybe-someday`?

troessner:
>> Maybe just tag it maybe-someday?

Good idea!

troessner:
Let's face it: We'll probably never address this. Let's rather have a clean issue overview. Closing this one.

mvz:
Hm, how am I going to keep track of ideas then?

troessner:
Seems like we had a misunderstanding, re-opening, see my last comment on #633

mvz:
This has been implemented :tada:.

PR #893, opened 2016-03-16
[WIP] Improve argument handling for our spec matchers.

troessner:
-> WIP
Fixes the first part of #891 
Speedy review would be awesome since this blocks my rspec refactoring necessary for mutant necessary for #890

### DIFF
--- a/lib/reek/spec.rb
+++ b/lib/reek/spec.rb
@@ -54,7 +54,7 @@ module Reek
     # It is recommended to pass this as a symbol like :UtilityFunction. However we don't
     # enforce this.
     #
-    # Additionally you can be more specific and pass in "smell_details" you
+    # Additionally you can be more specific and pass in "details" you
     # want to check for as well e.g. "name" or "count" (see the examples below).
     # The parameters you can check for are depending on the smell you are checking for.
     # For instance "count" doesn't make sense everywhere whereas "name" does in most cases.
@@ -63,14 +63,15 @@ module Reek
     # much sense.
     #
     # @param smell_type [Symbol, String, Class] The "smell type" to check for.
-    # @param smells_details [Hash] A hash containing "smell warning" parameters
+    # @param details [Hash] A hash containing "smell warning" parameters
+    # @param configuration [Reek::Configuration::AppConfiguration] The configuration that should be used.
     #
-    # @example Without smell_details
+    # @example Without details
     #
     #   reek_of(:FeatureEnvy)
     #   reek_of(Reek::Smells::UtilityFunction)
     #
-    # @example With smell_details
+    # @example With details
     #
     #   reek_of(:UncommunicativeParameterName, name: 'x2')
     #   reek_of(:DataClump, count: 3)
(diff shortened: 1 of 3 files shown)


lib/reek/spec.rb
```diff
@@ -83,8 +83,8 @@ module Spec
     #
     # :reek:UtilityFunction
     def reek_of(smell_type,
-                smell_details = {},
-                configuration = Configuration::AppConfiguration.default)
+                smell_details: {},
```
mvz: Maybe just `details:`

spec/reek/smells/attribute_spec.rb
```diff
@@ -34,7 +34,7 @@ class Klass
           attr_writer :my_attr
         end
       EOS
-      expect(src).to reek_of(:Attribute, name: 'my_attr')
+      expect(src).to reek_of(:Attribute, smell_details: { name: 'my_attr' })
```
mvz: Hm, I'm not sure I like this extra indirection.

lib/reek/spec.rb
troessner: Good idea!

spec/reek/smells/attribute_spec.rb
troessner: Not even when it's just "details" ? ;)

troessner:
@mvz I removed all changes again and started. Please review and then let's discuss if this is the way to go. If we agree on this (or something else) I'll start to adjust the specs ;)

mvz:
I'll have a look this weekend.

mvz:
I think the interface for `reek_of` and `reek_only_of` is nice as it currently is, and making details a named parameter doesn't improve things for the user of those methods. It only becomes a little awkward if there is also a configuration, but that can be fixed with #606. I can take a stab at that this week I think.

troessner:
>>  It only becomes a little awkward if there is also a configuration, but that can be fixed with #606. I can take a stab at that this week I think.

But right now #606 is blocking multiple bugfixes (because we did a bad job in our branch handling ;) ) and the Reek 4 release.
How about leaving it like it is and fix it in Reek 5?

mvz:
Wait, which fix is blocked by this?

mvz:
.. And which fixes are blocked by #606?

troessner:
>> Wait, which fix is blocked by this?

Let me explain this properly ;)

What we should have done: 

* create a reek-4 release branch
* that gets rebased against master
* so that current bug fixes can still be merged and released

Since we just merged all the breaking changes into master this is not possible.
Right now we quite a few bug fixes in master that we can not release because of that:

* #881 
* #885 
* #890 

And now #893 blocks the release of Reek 4. Which in turn is blocked by #606 
#606 might take a long time or very hard / complicated. We don't know that at the moment. 
#893 on the other hand is conceptually simple once we agreed on an API.
I can finish this off today or tomorrow. And then we could release.

Which is exactly what I'm suggesting ;)

mvz:
Ah, ok, thanks, that clears things up. In that, case, I would say the interface for `reek_of` and friends is good as it is and we shouldn't change it. Which means we can basically go straight to 'Release Reek 4'.

troessner:
Release PR incoming ;)

troessner:
Killing this PR now that #606 is merged.

PR #897, opened 2016-03-21
Update format of warning Attribute.md

munyari:
The warning message now has a slightly different format from what was specified in doc


troessner:
Well spotted, thanks!

PR #899, opened 2016-03-25
Fix excluded paths for custom config formats

thepry:
In current master, configuration is never passed to source locator, so it uses default value:
```ruby
# /lib/reek/source/source_locator.rb
  def initialize(paths, configuration: Configuration::AppConfiguration.default)
```

`Configuration::AppConfiguration.default` finds `*.reek` files in you app directory and uses them. Which causes problems in the following case:
If your configuration file isn't located in project root, or uses different format, like `config.reek.yml`, `exclude_paths` option would not work. But if you have `*.reek` file in your app root, reek will use `exclude_paths` from there, not from specified configuration file.

In my fix configuration is passed from `Application` down to ` Source::SourceLocator`

### DIFF
--- a/lib/reek/cli/application.rb
+++ b/lib/reek/cli/application.rb
@@ -13,13 +13,14 @@ module Reek
     # command line.
     #
     class Application
+      include Input
       attr_reader :configuration
 
       def initialize(argv)
         @options = configure_options(argv)
         @status = options.success_exit_code
         @configuration = configure_app_configuration(options.config_file)
-        @command = command_class.new(OptionInterpreter.new(options))
+        @command = command_class.new(OptionInterpreter.new(options), sources: sources)
       end
 
       def execute
@@ -47,6 +48,10 @@ module Reek
 
       def command_class
         options.generate_todo_list ? Command::TodoListCommand : Command::ReportCommand
+      end
+
+      def argv
+        options.argv
       end
     end
   end
--- a/lib/reek/cli/input.rb
+++ b/lib/reek/cli/input.rb
@@ -34,7 +34,7 @@ module Reek
 
       # :reek:UtilityFunction
       def working_directory_as_source
-        Source::SourceLocator.new(['.']).sources
+        Source::SourceLocator.new(['.'], configuration: configuration).sources
       end
 
       def sources_from_argv
--- a/lib/reek/cli/option_interpreter.rb
+++ b/lib/reek/cli/option_interpreter.rb
@@ -9,7 +9,6 @@ module Reek
     # Interprets the options set from the command line
     #
     class OptionInterpreter
-      include Input
       extend Forwardable
       def_delegators :options, :smells_to_detect, :success_exit_code, :failure_exit_code
 
--- a/spec/reek/cli/input_spec.rb
+++ b/spec/reek/cli/input_spec.rb
@@ -7,9 +7,21 @@ RSpec.describe Reek::CLI::Input do
     include Reek::CLI::Input
 
     def argv; end
+
+    def configuration; end
   end
 
+  let(:path_excluded_in_configuration) do
+    SAMPLES_PATH.join('source_with_exclude_paths/ignore_me/uncommunicative_method_name.rb')
+  end
+
+  let(:configuration) { test_configuration_for(SAMPLES_PATH.join('configuration/with_excluded_paths.reek')) }
+
   subject { DummyClass.new }
+
+  before do
+    allow(subject).to receive(:configuration).and_return(configuration)
+  end
 
   describe '#sources' do
     context 'when no source files given' do
@@ -37,6 +49,10 @@ RSpec.describe Reek::CLI::Input do
 
         it 'should use working directory as source' do
           expect(subject.sources).to_not be_empty
+        end
+
+        it 'should use configuration for excluded paths' do
+          expect(subject.sources).to_not include(path_excluded_in_configuration)
         end
       end
     end
--- a/lib/reek/cli/command/base_command.rb
+++ b/lib/reek/cli/command/base_command.rb
@@ -6,20 +6,17 @@ module Reek
       # Base class for all commands
       #
       class BaseCommand
-        def initialize(options)
+        def initialize(options, sources:)
           @options = options
+          @sources = sources
         end
 
         private
 
-        attr_reader :options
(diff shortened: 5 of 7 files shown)


troessner:
Hmmmm. But doesn't this problem **only** appear when you have and want to use 2 or more Reek configuration files?

mvz:
Perhaps it also happens when using Reek through its API?

mvz:
@thepry can you elaborate a little on how you are using reek when you encounter this problem?

spec/reek/cli/input_spec.rb
```diff
@@ -35,9 +39,20 @@ def argv; end
...
+        let(:path_excluded_in_configration) do
+          SAMPLES_PATH.join('source_with_exclude_paths/ignore_me/uncommunicative_method_name.rb')
+        end
+
         it 'should use working directory as source' do
           expect(subject.sources).to_not be_empty
         end
+
+        it 'should use configuration for excluded paths' do
+          expect(subject).to receive(:configuration).
```
mvz: Mocking the object under test is not recommended. I would move this test to OptionInterpreter, or use a different implementation of `#configuration` in the dummy class here, that allows to actually pass in the value in the initializer.

lib/reek/cli/option_interpreter.rb
```diff
@@ -13,9 +13,10 @@ class OptionInterpreter
       extend Forwardable
       def_delegators :options, :smells_to_detect, :success_exit_code, :failure_exit_code
 
-      def initialize(options)
+      def initialize(options, configuration: Configuration::AppConfiguration.default)
         @options = options
         @argv = options.argv
+        @configuration = configuration
```
mvz: There is no test for this change.

lib/reek/cli/input.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require_relative '../source/source_locator'
+require_relative '../configuration/app_configuration'
```
mvz: This line is not needed since you don't refer to the class defined there by name (it is always injected/defined by implementations of `#configuration`.

thepry:
@mvz i just typed in configration file name `reek.config` instead of `config.reek`. And when i called from a console `reek -c reek.config` i notices that `exclude_paths` option doesn't work.

But if i have `*.reek` file, reek wil use `exclude_paths` from there, not from config i specified with `- c` option

thepry:
>Hmmmm. But doesn't this problem only appear when you have and want to use 2 or more Reek configuration files?

No, it can happen if you have only 1 config file, but with not `.reek` format.

mvz:
I think this change makes sense and we should merge this. After that, we should refactor how Application interacts with the configuration and the option interpreter.

mvz:
On second thought, maybe another solution would make more sense: Move `#sources` out of `OptionInterpreter` altogether (since it doesn't actually look at any of the option arguments), and move it to `Application` instead, where `configuration` is already available.

thepry:
> Move #sources out of OptionInterpreter altogether (since it doesn't actually look at any of the option arguments), and move it to Application instead, where configuration is already available.

@mvz I like this one! I'll take a look in a few days, when i have time :-)

thepry:
@mvz what do you think about latest changes?

lib/reek/cli/application.rb
```diff
@@ -48,6 +49,10 @@ def configure_app_configuration(config_file)
       def command_class
         options.generate_todo_list ? Command::TodoListCommand : Command::ReportCommand
       end
+
+      def argv
```
troessner: You're not using this method anywhere as far as I can see. Let's remove it.

lib/reek/cli/input.rb
```diff
@@ -44,6 +44,14 @@ def sources_from_argv
       def source_from_pipe
         [$stdin]
       end
+
+      def argv
+        raise NotImplementedError
+      end
+
+      def configuration
+        raise NotImplementedError
```
troessner: Also not sure about those 2 methods here. I mean, it's a module already with instance methods, so it's clear that it is used as a mixin. Raising NotImplementedError here is something I find rather confusing than helpful. I'd rather remove those.

spec/reek/cli/input_spec.rb
```diff
@@ -7,10 +7,22 @@ class DummyClass
     include Reek::CLI::Input
 
     def argv; end
+
+    def configuration; end
+  end
+
+  let(:path_excluded_in_configration) do
```
troessner: typo: "configuration" ;)

troessner:
Don't get me wrong, I love it that you fixed this bug, I just believe that we can drop at least half of the changes to our files in /lib because they are unnecessary. 
If you address my comments and squash this bad boy I'm more than happy to merge this ;)

spec/reek/cli/input_spec.rb
```diff
@@ -7,10 +7,22 @@ class DummyClass
     include Reek::CLI::Input
 
     def argv; end
+
+    def configuration; end
```
troessner: Wait, thats also unnecessary, isn't it?

lib/reek/cli/application.rb
mvz: I think it's needed by Files, but I think that whole module may just have to be moved here instead.

spec/reek/cli/input_spec.rb
thepry: It is necessary, since we are stubbing this method for specs.

lib/reek/cli/application.rb
thepry: Yeah, `argv` is used in `/cli/input.rb` module.

thepry:
@troessner Hi. I've removed unnecessary code and squashed all changes into one commit. Plz take a look again.

Also, if `Reek::CLI::Input` module is used only in `Reek::CLI::Application` and relies on it's instance method and variables, i think we should just move all methods from `Input` to `Application`. But this refactoring is not related to bugfix PR.

troessner:
Looks good to me! Build failing unfortunately, but that should be fixed with #900 - waiting for the build now.
You'll then just need to rebase against master and then this one is ready to go!

troessner:
Merging this one since #900 is in master.
Great job, thanks!

troessner:
And 4.0.1 is released with the fix.

PR #900, opened 2016-04-10
Remove factory_girl warning.

troessner:
factory_girl is now at 4.7.0 which makes #899 fail


troessner:
Trivial fix, merging this one myself.

Issue #901, opened 2016-04-10
Kill cli/input.rb


troessner:
Follow up from #899 
CLI::Input is confusing and we should kill it.

mvz:
Done :rocket:

troessner:
Yes!

PR #908, opened 2016-04-13
Remove CLI::Input and other refactorings

mvz:
#901.


lib/reek/cli/application.rb
```diff
@@ -13,18 +13,31 @@ module CLI
...
+                                     sources: sources,
+                                     configuration: configuration)
       end
 
       def execute
-        command.execute self
+        command.execute
+      end
+
+      def sources
```
mvz: Maybe I should make this private too.

troessner: 👍

troessner:
Very nice, love it!
From a PR point I think this is ready to merge after squashing.
Seems like we have mutant problem though. If you check out e.g. the build for [2.3](https://travis-ci.org/troessner/reek/jobs/122722995) you'll see this:

>> (382/552)  69% - killtime: 149.78s runtime: 6901.11s overhead: 6751.33s
(382/552)  69% - killtime: 149.78s runtime: 6902.11s overhead: 6752.33s
Your test run exceeded 120 minutes. 
One possible solution is to split up your test run.

I'm fine with merging it as it is but we then should look into the mutant problem.

mvz:
> we then should look into the mutant problem.

Yes. We need to take another look at how mutant works and what it expects from our tests.

mvz:
@troessner squashed.

Issue #505, opened 2015-05-08
Enhance reek_of matcher to enable smell during testing


beanieboi:
extracted from #501
> I'm thinking we sould enhance the reek_of matcher to automatically enable the smell to be detected.

currently you need to enable the matcher by hand if you want to use the `reek_of` matcher. thats not a problem when the smell is enabled by default. if it's not it leads to unnecessary boiler code. see https://github.com/troessner/reek/blob/develop/spec/reek/smells/attribute_spec.rb#L17-L21

using `reek_of` should automatically enable the smell during the tests

mvz:
See #503. Sorry for not linking it up to #501.

mvz:
This has been implemented and released in Reek 4.0.2.

beanieboi:
awesome, thanks!

PR #909, opened 2016-04-16
Automatically enable smell in reek_of matcher

mvz:
#505.

### DIFF
--- a/spec/reek/smells/unused_private_method_spec.rb
+++ b/spec/reek/smells/unused_private_method_spec.rb
@@ -4,12 +4,6 @@ require_lib 'reek/examiner'
 require_relative 'smell_detector_shared'
 
 RSpec.describe Reek::Smells::UnusedPrivateMethod do
-  let(:configuration) do
-    test_configuration_for(
-      described_class =>
-        { Reek::Smells::SmellConfiguration::ENABLED_KEY => true }
-    )
-  end
   let(:detector) { build(:smell_detector, smell_type: :UnusedPrivateMethod) }
 
   it_should_behave_like 'SmellDetector'
@@ -24,8 +18,8 @@ RSpec.describe Reek::Smells::UnusedPrivateMethod do
         end
       EOF
 
-      expect(source).to reek_of(:UnusedPrivateMethod, { name: :start }, configuration)
-      expect(source).to reek_of(:UnusedPrivateMethod, { name: :drive }, configuration)
+      expect(source).to reek_of(:UnusedPrivateMethod, name: :start)
+      expect(source).to reek_of(:UnusedPrivateMethod, name: :drive)
     end
 
     it 'creates warnings correctly' do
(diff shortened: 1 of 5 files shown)


spec/reek/smells/unused_private_method_spec.rb
```diff
@@ -37,19 +31,8 @@ def drive; end
...
-      expect(first_warning.smell_type).to eq(Reek::Smells::UnusedPrivateMethod.smell_type)
-      expect(first_warning.parameters[:name]).to eq(:drive)
-      expect(first_warning.lines).to eq([4])
-
-      second_warning = examiner.smells.last
-      expect(second_warning.smell_type).to eq(Reek::Smells::UnusedPrivateMethod.smell_type)
-      expect(second_warning.parameters[:name]).to eq(:start)
-      expect(second_warning.lines).to eq([3])
+      expect(source).to reek_of(:UnusedPrivateMethod, { name: :drive, lines: [4] })
+      expect(source).to reek_of(:UnusedPrivateMethod, { name: :start, lines: [3] })
```
troessner: Reads a lot nicer, awesome

troessner:
Travis is red...code looks good though.

mvz:
Yeah, I accidentally broke everything :smile:. Should be fixed now.

mvz:
I :heart: RuboCop.

mvz:
We may need to learn a little more about mutant. /cc @mbj

mbj:
> We may need to learn a little more about mutant. /cc @mbj

Meet me in mutation testing Slack channel to discuss the questions?

troessner:
>> We may need to learn a little more about mutant.

Looking at the output:

>> Coverage:        93.15%
Expected:        100.00%
The command "bundle exec mutant --include lib --require reek --use rspec --since master --jobs 4 "Reek*"" exited with 1.

and especially at the mutants like this one:

>>
-  self[detector] = (self[detector] || {}).merge(config)
+  self[detector] = (self[detector] || self).merge(config)

it's obvious that our tests are not 100% bulletproof. Not sure if there is anything we can do about besides...writing more tests.
Problem is that we have been ignoring mutant so far because having a 100% spec coverage for mutation testing is just simply exhausting 😆 (and it also takes a lot of fun away for a non-commercial project)
But then it's kind of pointless. Maybe we should switch back to whitelisting only API relevant files?

mbj:
> Problem is that we have been ignoring mutant so far because having a 100% spec coverage for mutation testing is just simply exhausting

I'd argue it is not, since you likely fell into the most common mistake using it:

Mutation testing is NOT about adding tests, its about removing or structuring code better so the mutation cannot occur anymore. The idea is that an alive mutation only when nothing else works should be cured with adding (or changing) a test.

Mutant "always" (there are some well choosen exceptions) reduces semantics. If mutant finds a an alive reduction, for most cases you can simply use the reduction in your code and kill the mutant without adding a test.

If the reduction cannot be killed via code change (because the code would turn into invalid semantics): You miss an important test.

And adding that important test is to complex, the design is smelly and should be fixed.

Obviously I argue without detailed understanding of that specific case, but for educational experience (on both sides): What about arranging a small pairing session where we work out how the case you cited applies to the above?

troessner:
>> I'd argue it is not, since you likely fell into the most common mistake using it:
Mutation testing is NOT about adding tests, its about removing or structuring code better so the mutation cannot occur anymore. The idea is that an alive mutation only when nothing else works should be cured with adding (or changing) a test.

Hmmm. Good point. Seems like we misunderstood this indeed. At least I did 😆 

>> Obviously I argue without detailed understanding of that specific case, but for educational experience (on both sides): What about arranging a small pairing session where we work out how the case you cited applies to the above?

Awesome idea! I'll nudge you as soon as I'll tackle the next issue in Reek, thanks!

Issue #907, opened 2016-04-12
Suggestion: Detect subclassing of core classes


andyw8:
I came across some code today which reminded me of this post: http://words.steveklabnik.com/beware-subclassing-ruby-core-classes

I think it would be fairly easy to detect this in Reek.

Some considerations:

* What classes should be considered? String, Hash and Array come to mind but I'm sure there's plenty others.
* What should the smell be called?

troessner:
Good idea!

I think your selection is good to get started.

How about SubclassedFromCoreClass smell? 😆

andyw8:
Ok, I'll start looking into it.

troessner:
Nice!

mvz:
No need to close this, right?

andyw8:
Sure, I just thought since it's not waiting on any action from the Reek team it would help to keep the noise down. Could be a while until I get around to actually implementing it.

troessner:
#954 is merged and released, closing this one.

PR #917, opened 2016-04-29
WIP: Add SubclassedFromCoreClass smell detector

andyw8:
Addresses #907 

Opening for early feedback. There's one pending spec, I'm still getting the hang of Parser so any advice would be welcome.


troessner:
Looks very promising!

A couple of generic suggestions:

1.) Don't forget to require the detector in lib/reek/smells.rb
2.) An integration test would be great (that one would have also caught [1]

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -0,0 +1,32 @@
...
+      def self.contexts
+        [:class]
+      end
+
+      # Checks +ctx+ for if it is subclasssed from a core class
+      #
+      # @return [Array<SmellWarning>]
+      def inspect(ctx)
+        _class_node, ancestor_node, _body_node = ctx.exp.to_a
+        return [] if ancestor_node.nil?
```
troessner: How about moving this to a new predicate method `ancestor?` on `exp`?

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -0,0 +1,32 @@
...
+        ancestor_namespace, ancestor = ancestor_node.to_a
+        return [] unless CORE_CLASSES.include?(ancestor) && ancestor_namespace.nil?
```
troessner: Same here. I'd prefer it we had proper methods for things like `ancestor_namespace` on `exp` with a proper method documentation that shows to what kind of sexp this functionality applies....

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -0,0 +1,32 @@
...
+
+        [smell_warning(
+          context: ctx,
+          lines: [ctx.exp.line],
+          message: "inherits from a core class #{ancestor}",
+          parameters: { count: 1 })]
```
mvz: You could put `ancestor: ancestor` in the parameters.

spec/reek/smells/subclassed_from_core_class_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+          class Dummy < ::Array
+          end
+        end
+      EOS
+      syntax_tree = Reek::Source::SourceCode.from(src).syntax_tree
+      ctx = Reek::Context::ModuleContext.new(nil, syntax_tree)
+      smells = detector.inspect(ctx)
+      expect(smells.length).to eq(1)
+      expect(smells[0].smell_type).to eq(described_class.smell_type)
+      expect(smells[0].parameters[:count]).to eq(1)
```
mvz: You should probably changes some of these complex tests to use something like:

```
  expect(src).to reek_of(:SubclassedFromCoreClass, ancestor: 'Array')
```

lib/reek/smells/subclassed_from_core_class.rb
andyw8: Would that be `Reek::AST::Node`?

andyw8:
I'm thinking that `SubclassedFromCoreClass` may not be a good name, since there are cases where subclassing a core class is the correct thing to do, e.g. with `RuntimeError`.

andyw8:
@troessner re. an integration test, do you mean using Cucumber? Should I add some code containing this smell to the `inline.rb` sample file?

lib/reek/smells/subclassed_from_core_class.rb
mvz: Those extensions are defined in modules under `Reek::AST::SexpExtensions`.

lib/reek/smells/subclassed_from_core_class.rb
troessner: 👍

troessner:
>> @troessner re. an integration test, do you mean using Cucumber? Should I add some code containing this smell to the inline.rb sample file?

No, we shouldn't touch those since they exist in the wild so this would be very confusing.
You can e.g. check out [this feature](https://github.com/troessner/reek/blob/master/features/configuration_files/directory_specific_directives.feature) to check out the current idiomatic Reek way.

lib/reek/smells/subclassed_from_core_class.rb
andyw8: I'm having trouble understanding to structure this. Could you give me some guidance? (cc @troessner)

mvz: The modules under `SexpExtensions` are included in the classes representing the AST nodes. The modules that is included depends on the node type.

So nodes of type `:class` get `ClassNode` included, which is defined in `lib/reek/ast/sexp_extensions/module.rb`. In fact, you can see there that the relevant methods are defined in `ModuleNodeBase`, because many methods that are relevant for `:class` are also relevant for `:module` and `:casgn` nodes.

Long story short, you would probably want to define `ancestor` on `ClassNode` to start with, and then change the code above to:

```ruby
  exp = ctx.exp
  return [] unless exp.ancestor
```
```

waldyr:
Sorry for blending in but I'm interested in this feature how is that going?

troessner:
I think @andyw8 is the only one who can answer this question at the moment ;)

@waldyr  are you interested in how smell detectors are structured (i can explain that on request) or in this specific smell detector?

waldyr:
@troessner I'm interested in this specific smell detector.

andyw8:
Hi @waldyr, thanks for your interest. I have a lot on at the moment, but hope to get back to this next week. Do you have any thoughts on the implementation so far?

waldyr:
@andyw8 I think the implementation is okay. Keep up with this great work :)

troessner:
@andyw8 if you don't have the time, no problem, but let's not kill @waldyr impulse and motivation by inactivity. Let's deprecate this PR in favor of #945 - @waldyr care to over and push your PR forward?
I'll happily review and help you as soon as something is reviewable.

troessner:
Closing this one in favor of #954 - thanks for getting this started @andyw8 !

andyw8:
Apologies, the replies for this were filtered out of my inbox. Thanks for taking over @waldyr.

PR #918, opened 2016-05-09
Add emacs plugin page

hanmoi-choi:
Hi

I like your gem so much and am Emacs user.

There is no plugin for Reek so I have created a simple package for interfacing Reek inside emacs.

Cheers


troessner:
Merged, very nice!

PR #919, opened 2016-05-10
Add ToC to README.

troessner:
Our README is pretty big and kind of unwieldy for new users. 
This ToC hopefully makes it easier for people to browse our README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,37 @@
 
 # Code smell detector for Ruby
 
-## Overview
+**Table of Contents**
 
+- [Overview](#overview)
+- [Quickstart](#quickstart)
+- [Example](#example)
+- [Supported rubies](#supported-rubies)
+- [Fixing Smell Warnings](#fixing-smell-warnings)
+- [Sources](#sources)
+- [Code smells](#code-smells)
+- [Configuration](#configuration)
+  - [Command-line interface](#command-line-interface)
+  - [Configuration file](#configuration-file)
+    - [Configuration loading](#configuration-loading)
+    - [Configuration options](#configuration-options)
+  - [Source code comments](#source-code-comments)
+  - [Generating a 'todo' list](#generating-a-todo-list)
+- [Usage](#usage)
+- [Developing Reek / Contributing](#developing-reek--contributing)
+- [Output formats](#output-formats)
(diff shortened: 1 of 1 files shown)


mvz:
Good idea. However, the links don't seem to work when viewing on GitHub (can't try loyally RN). Also, we should probably move some of that stuff into other files (but that can be done as a next step).

troessner:
>> However, the links don't seem to work when viewing on GitHub (can't try loyally RN). 

The links work for me when viewing on github and locally as well, so I'm not sure what's happening here. Could you try again?

>> Also, we should probably move some of that stuff into other files (but that can be done as a next step).

Disagree. I really hate those README's that just contain a few lines and then tell you "oh, please check out the docs". I think this is a real conversion killer and one of __the__ main reasons why new users immediately jump ship.

I think our README is awesome as it is. It covers a lot of ground but not too much. It just needs a ToC ;)

mvz:
Maybe it just doesn't work on mobile. I can check on my laptop Saturday.

troessner:
Works on my phone as well 😆

mvz:
Works fine on desktop Firefox too. Both Chrome and Firefox on Android don't work, but that's less important imho.

troessner:
Using the default application for web browsing (which is a kind of chrome I believe) works on my Android phone (the application name is just "internet").

troessner:
Ahhhhh: http://webmasters.stackexchange.com/questions/47615/what-is-the-default-web-browser-on-android

Ok, so apparently it works in "Android browser" ;)

mvz:
Yep, that's the chrome I used.

troessner:
But its not chrome apparently. It is webkit based, but not chrome (i always assumed "internet" would be a special chrome version).

mvz:
Ah no our comments crossed. I have a newer Android whose default browser uses the Chrome logo, so I'm assuming it's actually Chrome.

Asking GitHub for the desktop site makes the links work, by the way.

mvz:
Good god that is some messed up HTML GitHub is producing: The a elements inside the headings use the `href` attribute instead of `name` to store the anchor. WTF?!?

troessner:
Yeah, a little weird :)

Issue #914, opened 2016-04-23
Incomplete todo file generation?


rogerkk:
I'm upgrading from reek 1.x, and a lot of stuff has changes since I last touched this gem, and I got a lot of new warnings.

I was glad to discover the new todo file generation, but it seems like the file generated is somewhat incomplete. I'm not sure if I've missed something here or if this is actually a bug, but the user experience here is a bit confusing so I'm hoping this issue is useful.

Here's what I'm seeing:
```
$ reek -t app/

'.todo.reek' generated! You can now use this as a starting point for your configuration.
$ wc -l .todo.reek
1094 .todo.reek
$ reek app/
...
47 total warnings
```

The warnings are of different types, like Attribute, TooManyInstanceVariables, TooManyStatements, IrresponsibleModule etc.

Is this a known issue? If so, please let me know what info would be most relevant to help you debug it.

troessner:
Can't really reproduce this:

Sample file:

```
  ( $ ) cat foo.rb
class Foo
  attr_accessor :bar

  def hola(arg = true)
    bla arg
    bla arg
  end
end
```

Running Reek:

```
( $ ) reek foo.rb
foo.rb -- 4 warnings:
  [2]:Attribute: Foo#bar is a writable attribute [https://github.com/troessner/reek/blob/master/docs/Attribute.md]
  [4]:BooleanParameter: Foo#hola has boolean parameter 'arg' [https://github.com/troessner/reek/blob/master/docs/Boolean-Parameter.md]
  [5, 6]:DuplicateMethodCall: Foo#hola calls bla arg 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [1]:IrresponsibleModule: Foo has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]
```

Generating the todo file:

```
  ( $ ) reek --todo foo.rb

'.todo.reek' generated! You can now use this as a starting point for your configuration.
```

And running Reek again:


```
  ( $ ) reek -c .todo.reek foo.rb
  ( $ )
```

Can you please give me a reproducible example similar to my example?

mvz:
@rogerkk is another `.reek` file present in the directory?

rogerkk:
> Can you please give me a reproducible example similar to my example?

I'm sorry. I've tried but I haven't been able to reproduce this with a constructed example.

> is another .reek file present in the directory?

[...]

PR #921, opened 2016-05-21
Improve todo generation

mvz:
This does several things:
* Clean up specs and cukes for todo file generation.
* Make generated context list unique. (This was a bug uncovered by the previous step.)
* Fix issue #914.


lib/reek/cli/command/todo_list_command.rb
```diff
@@ -36,13 +36,13 @@ def scan_for_smells
         end
 
         def groups_for(smells)
-          @groups ||= begin
-            Hash[
-              smells.group_by(&:smell_type).map do |smell_type, smells_for_type|
-                [smell_type, { 'exclude' => smells_for_type.map(&:context) }]
+          @groups ||=
+            begin
+              todos = smells.group_by(&:smell_class).map do |smell_class, smells_for_class|
```
troessner: I like the change to `smell_class`, don't get me wrong. But shouldn't we then kill `SmellDetector.smell_type` throughout Reek?

lib/reek/smells/smell_detector.rb
```diff
@@ -55,6 +55,12 @@ def exception?(context)
         context.matches?(value(EXCLUDE_KEY, context, DEFAULT_EXCLUDE_SET))
       end
 
+      def self.todo_configuration_for(smells)
+        default_exclusions = default_config['exclude']
+        exclusions = default_exclusions + smells.map(&:context)
+        { smell_type => { 'exclude' => exclusions.uniq } }
```
troessner: Nice and concise.
The only thing that remains is too fix the UsesClassMethod smell, but that is something bigger for another pull request ;)

spec/reek/cli/command/todo_list_command_spec.rb
```diff
@@ -37,6 +37,40 @@
         result = command.execute
         expect(result).to eq(Reek::CLI::Options::DEFAULT_SUCCESS_EXIT_CODE)
       end
+
+      it 'writes a todo file' do
+        command.execute
+        expected_yaml = { 'FeatureEnvy' => { 'exclude' => ['self'] } }.to_yaml
```
troessner: That's very confusing here and below. The `expected_yaml` is based on the default values of our factory_girl. But you have no reference here to our factory definitions so an innocent reader will be lost here why this is "expected".
It's also duplicated from our factories - if we change the value in the factory those specs will break which is detrimental to our maintenance. 
How about rather using factory_girls `attributes_for` -> https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#using-factories

spec/reek/cli/command/todo_list_command_spec.rb
```diff
@@ -37,6 +37,40 @@
...
+
+    context 'smells with duplicate context found' do
+      before do
+        smells = [FactoryGirl.build(:smell_warning), FactoryGirl.build(:smell_warning)]
+        allow(command).to receive(:scan_for_smells).and_return(smells)
+      end
+
+      it 'writes the context into the todo file once' do
+        command.execute
+        expected_yaml = { 'FeatureEnvy' => { 'exclude' => ['self'] } }.to_yaml
```
troessner: Same here.

lib/reek/cli/command/todo_list_command.rb
mvz: Do you mean `SmellDetector.smell_type` or `SmellWarning#smell_type`?

spec/reek/cli/command/todo_list_command_spec.rb
mvz: The hash doesn't have the same structure as SmellWarning's attributes, so I'll add an explicit value for the context to the smell instead.

troessner: Sure, I have nothing against using the hash - but the values in that hash shouldn't be literal but instead come from `attributes_for` ...

mvz:
Hm, looks like I can't innocently create a dummy subclass of SmellDetector for testing :boom:. Down the rabbit hole we go ... :rabbit:.

lib/reek/cli/command/todo_list_command.rb
troessner: Sorry for being ambiguous. And actually I'd rather retract my statement. Why not stick with `SmellWarning#smell_type` but instead using another method? If you'd go ahead with that change we'd have `SmellWarning#smell_type`  and `SmellWarning#smell_class` and both would return the same.

mvz: Ah, but they don't: Smell type is a string, smell class is a class.

troessner: Ok, the returned type is different but the semantics are very similar. Shouldnt we unify it?

spec/reek/cli/command/todo_list_command_spec.rb
mvz: There is no current object that has attributes that match that hash.

lib/reek/cli/command/todo_list_command.rb
mvz: I see two options:
* Use the class. Here we will run into trouble during the conversion to YAML. Providing a `#to_yaml` that returns the smell type string might work, but it's a can of worms I'm not that eager to open.
* Use the string. This is possible if we can fetch the corresponding detector class from a smell repository object. I would prefer this since it will allow decoupling the smell warning from the smell detector in the end. Also, it may allow us to use one smell repository shared by all examiners.

spec/reek/cli/command/todo_list_command_spec.rb
troessner: Ok, I feel that at the moment we are completely misunderstanding each other.
My problem is that this:

```Ruby
{ 'FeatureEnvy' => { 'exclude' => ['self'] } }
```

is completely obscure to every reader. There is no hint, reference or clue that 'self' is something we set in our factory, buried deep down in spec/factories/factories. Additionally, when I now edit the factory in spec/factories/factories this whole spec goes booom. Which it shouldn't.

Hence I suggested to use something like this:

```Ruby
# bundle exec rake console

require 'factory_girl'
require_relative 'spec/factories/factories'

FactoryGirl.attributes_for(:smell_warning)

# => {:context=>"self", :source=>"dummy_file", :lines=>[42], :message=>"smell warning message", :parameters=>{}}
```

You see what I mean?

lib/reek/cli/command/todo_list_command.rb
troessner: >>  Providing a #to_yaml that returns the smell type string might work, but it's a can of worms I'm not that eager to open.

I don't follow - like what?

>> Use the string. This is possible if we can fetch the corresponding detector class from a smell repository object. I would prefer this since it will allow decoupling the smell warning from the smell detector in the end. Also, it may allow us to use one smell repository shared by all examiners.

👍  - but I'd still like to hear more about the can of worms above ;)

spec/reek/cli/command/todo_list_command_spec.rb
mvz: Yes, I think I do see what you mean. I've solved the problem of the obscure 'self' by providing the context value in the spec itself (hence, we're commenting on an outdated diff here).

My point o.t.o.h. was that there is no `:foo` for which `attributes_for(:foo)` gives me the expected hash. Hence, the only way I can use `attributes_for` would be something like:

```
context = attributes_for(:smell_warning)['context']
expected_yaml = { 'FeatureEnvy' => { 'exclude' => [context] } }.to_yaml
```

But then I can just do:
```
expected_yaml = { 'FeatureEnvy' => { 'exclude' => [smell.context] } }.to_yaml
```

So `attributes_for` isn't really helpful in this particular case.

lib/reek/cli/command/todo_list_command.rb
mvz: > the can of worms above

We once tried to get conversion to YAML working for SmellWarning, using `#encode_with` for Psych. We removed it in the end: fbda199884b834e74e1899daf32d5ee0cfcb5788

> Use the string

I'd like to do this, but for a good implementation, I think it needs a bit of refactoring. So, how about deferring that to another PR?

troessner: >> I'd like to do this, but for a good implementation, I think it needs a bit of refactoring. So, how about deferring that to another PR?

Fine by me ;)

troessner:
Looks good to me now! Can you please squash your commits now? I have nothing against multiple commits in one PR as long as they are meaningful. Looking at the current commits for this PR, I think at least half of them can and should go ;)

mvz:
> I think at least half of them can and should go ;)

Definitely :-).

mvz:
I squashed the commits. They now correspond to each of the three points in the description.

troessner:
Merged this bad boy, awesome job!

troessner:
Release incoming...

troessner:
And 4.0.3 is released!

mvz:
Wow speedy ... :grinning:.

troessner:
Well, this is how I roll! ;)

PR #926, opened 2016-05-29
Report all nested iterators

### LINKED ISSUE #923: Report all cases of nested iterators in a given method
The NestedIterators specs specify this:
```
it 'reports nested iterators only once per method even if levels are different'
```

This is a bit odd. If there are two (unrelated) cases of NestedIterators in the same method, it makes sense to report both.

mvz:
Fixes #923. Nested iterator warnings with the same depth are combined, but all lines are reported.

### DIFF
--- a/lib/reek/smells/nested_iterators.rb
+++ b/lib/reek/smells/nested_iterators.rb
@@ -13,9 +13,8 @@ module Reek
     class NestedIterators < SmellDetector
       # Struct for conveniently associating iterators with their depth (that is, their nesting).
       Iterator = Struct.new :exp, :depth do
-        include Comparable
-        def <=>(other)
-          depth <=> other.depth
+        def line
+          exp.line
         end
       end
 
@@ -36,41 +35,55 @@ module Reek
         )
       end
 
-      #
-      # Attempts to find the deepest nested iterator and warns if it's depth
-      # is bigger than our allowed maximum.
+      # Generates a smell warning for each independent deepest nesting depth
+      # that is greater than our allowed maximum. This means if two iterators
+      # with the same depth were found, we combine them into one warning and
+      # merge the line information.
       #
       # @return [Array<SmellWarning>]
       #
-      # :reek:TooManyStatements: { max_statements: 6 }
       def inspect(ctx)
-        configure_ignore_iterators(ctx)
-        deepest_iterator = find_deepest_iterator ctx
-        return [] unless deepest_iterator
-        depth = deepest_iterator.depth
-        return [] unless depth > max_nesting(ctx)
+        configure_ignore_iterators ctx
+        violations = find_violations ctx
 
-        [smell_warning(
-          context: ctx,
-          lines: [deepest_iterator.exp.line],
-          message: "contains iterators nested #{depth} deep",
-          parameters: { name: ctx.full_name, count: depth })]
+        violations.group_by(&:depth).map do |depth, group|
+          lines = group.map(&:line)
+          smell_warning(
+            context: ctx,
+            lines: lines,
+            message: "contains iterators nested #{depth} deep",
+            parameters: { name: ctx.full_name, count: depth })
+        end
       end
 
       private
 
       attr_accessor :ignore_iterators
 
+      # Finds the set of independent most deeply nested iterators that are
+      # nested more deeply than allowed.
       #
-      # @return [Iterator|nil]
+      # Here, independent means that if iterator A is contained within iterator
+      # B, we only include A. But if iterators A and B are both contained in
+      # iterator C, but A is not contained in B, nor B in A, both A and B are
+      # included.
       #
-      def find_deepest_iterator(ctx)
+      # @return [Array<Iterator>]
+      #
+      def find_violations(ctx)
+        candidates = find_candidates ctx
+        max_allowed_nesting = max_nesting(ctx)
+        candidates.select { |it| it.depth > max_allowed_nesting }
+      end
+
+      # Finds the set of independent most deeply nested iterators regardless of
+      # nesting depth.
+      #
+      # @return [Array<Iterator>]
+      #
+      def find_candidates(ctx)
         exp = ctx.exp
-        return nil unless exp.find_nodes([:block])
-        scout(parent: exp, exp: exp, depth: 0).
-          flatten.
-          sort.
-          last
+        scout(exp: exp, depth: 0)
       end
 
       # A little digression into parser's sexp is necessary here:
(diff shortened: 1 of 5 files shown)


mvz:
Something is very wrong with the timeouts on Ruby 2.3. Also, why is there negative overhead for mutant? :confounded:

mvz:
@mbj Our mutant run doesn't seem to end on Travis with MRI 2.3.0 for this PR. Locally, it finishes due to the timeout I added. Any thoughts on what could be going on and/or how to debug this?

lib/reek/smells/nested_iterators.rb
```diff
@@ -37,40 +36,42 @@ def self.default_config
...
+            message: "contains iterators nested #{depth} deep",
+            parameters: { name: ctx.full_name, count: depth })
+        end
       end
 
       private
 
       attr_accessor :ignore_iterators
 
+      def find_violations(ctx)
```
troessner: Let's add method docs ;)

lib/reek/smells/nested_iterators.rb
```diff
@@ -79,31 +80,36 @@ def find_deepest_iterator(ctx)
...
+        exp.find_nodes([:block]).flat_map do |node|
+          examine_block_node(node, depth)
+        end
+      end
+
+      def examine_block_node(iterator, depth)
+        new_depth = increment_depth(iterator, depth)
+        # 1st case: we recurse down the given block of the iterator. In this case
+        # we need to check if we should increment the depth.
+        block_result = scout(exp: iterator.block, depth: new_depth)
```
troessner: Yep, and with that we are exactly back to the way this detector used to work before I refactored it: Away from proper recursion and back to "A calls B, B calls A, A calls B, B calls....".
That was one of the reasons why I refactored it. Not really a fan, I have to say.

troessner: On the other hand I don't want to block this improvement just because of that. But isn't there a way to not have circular calls and instead use recursion like before?

mvz: I see what you mean. It does complicate the flow. I'll see if I can restructure it while keeping the methods small enough.

mbj:
> @mbj Our mutant run doesn't seem to end on Travis with MRI 2.3.0 for this PR. Locally, it finishes due to the timeout I added. Any thoughts on what could be going on and/or how to debug this?

This can have many reasons:
1. Mutant auto detects the number of CPUs via `/proc/cpuinfo` this does NOT reflect the resources a particular build can use. So I typically have a Travis/Circle CI specific hard limit via `-j N` where N is choosen by experiment. For reek I expect `4` to work.
2. There are rare mutations that cause infinit loops. Try to isolate it locally and refactor away, if not possible add an timeout on tests that are supposed to cover the subject with the infinite loop mutation.

mvz:
@troessner I addressed all your comments.

lib/reek/smells/nested_iterators.rb
```diff
@@ -37,40 +36,45 @@ def self.default_config
...
       private
 
       attr_accessor :ignore_iterators
 
+      # Finds deepest iterators that are nested more deeply than allowed.
+      #
+      # @return [Array<Iterator>]
+      #
+      def find_violations(ctx)
+        deepest_iterators = find_deepest_iterators ctx
```
troessner: But now the naming is confusing (sorry, should have noticed that in my initial review). It's not the "deepest" iterators, it's just all the iterators, isn't it? You check for the depth of the iterators 2 lines below so I'd remove the "deepest" in the local variable name and in the method name.

spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -112,49 +113,48 @@ def super_call_with_block
...
         end
       EOS
-      warnings = detector.inspect(build(:method_context, source: source))
-      expect(warnings.size).to eq(1)
-      warning = warnings.first
-
-      expect(warning.parameters[:name]).to eq('bad')
-      expect(warning.lines).to eq([3])
+      expect(source).to reek_of(:NestedIterators, name: 'bad', lines: [2], count: 2)
+      expect(source).to reek_of(:NestedIterators, name: 'bad', lines: [3], count: 3)
```
troessner: Very nice spec improvements, way more readable.

lib/reek/smells/nested_iterators.rb
```diff
@@ -79,31 +83,33 @@ def find_deepest_iterator(ctx)
...
           # the depth stays the same.
-          scout(parent: iterator, exp: iterator.block, depth: increment_depth(iterator, depth)) +
-            scout(parent: iterator, exp: iterator.call, depth: depth)
+          nested_iterators = scout(exp: iterator.block, depth: new_depth) +
+            scout(exp: iterator.call, depth: depth)
+          if nested_iterators.empty?
+            Iterator.new(iterator, new_depth)
+          else
+            nested_iterators
+          end
```
troessner: Hmmm, just a small thing: Now it's not tail-recursive anymore. Given the non-performance critical domain of Reek this is not a big problem, but I wanted at least to point it out.

troessner:
Reviewed, looks good to me now! Just one comment to address and then we're good to go!

mbj:
@mvz was the mutant problem resolved?

lib/reek/smells/nested_iterators.rb
troessner: On second thought I realize it was never tail recursive anyway due to the "+" and the increment_depth call, but hey, it looked cooler, didn't it?!

mvz: It did look cooler, yes :smile:.

lib/reek/smells/nested_iterators.rb
mvz: Well, it's the minimal set of independent deepest iterators (which may not be deep enough to count as violations), but that's a bit long. Perhaps I can call them `candidates` instead?

mvz:
@mbj I debugged it last night: There was a mutation that caused a SystemStackLevel exception, and the stack trace was apparently too much for mutant to handle. I got mutant to correctly kill the mutation by setting `RUBY_THREAD_VM_STACK_SIZE=64000`.

I had to modify mutant's source to get it to print each mutation candidate, since the `--debug` option doesn't seem to do anything.

mvz:
There's an actual test failure after the rebase to master. Investigating ...

lib/reek/smells/nested_iterators.rb
troessner: >>  Perhaps I can call them candidates instead?

Yes, love it!

spec/spec_helper.rb
```diff
@@ -82,6 +82,11 @@ def sexp(type, *children)
   config.mock_with :rspec do |mocks|
     mocks.verify_partial_doubles = true
   end
+
+  # Avoid infinitely running tests
```
troessner: Can we add a comment that this is only relevant for `Mutant`?

mbj:
> I had to modify mutant's source to get it to print each mutation candidate, since the --debug option doesn't seem to do anything.

Ohh, sorry that code was removed long time ago but the option is still in `--help`. I've to fix that.

https://github.com/mbj/mutant/issues/601

mbj:
> @mbj I debugged it last night: There was a mutation that caused a SystemStackLevel exception, and the stack trace was apparently too much for mutant to handle. I got mutant to correctly kill the mutation by setting RUBY_THREAD_VM_STACK_SIZE=64000.

Could this be related to: https://github.com/mbj/mutant#the-crash--stuck-problem-mri

mvz:
> Could this be related to: https://github.com/mbj/mutant#the-crash--stuck-problem-mri

I don't think so, since it happens with MRI 2.3.0 and 2.3.1 and if I understand correctly the stuck VM problem should be fixed in 2.3?

mvz:
> This branch has conflicts that must be resolved

:crying_cat_face:

mvz:
And it's green!

troessner:
Merged, very nice! Let's release this bad boy right away!

mbj:
> I don't think so, since it happens with MRI 2.3.0 and 2.3.1 and if I understand correctly the stuck VM problem should be fixed in 2.3?

Its less likely to happen, but I saw instances of this. when the stack grows to big and threads / timeouts mix in a bad way it can "still" happen.

Do you have a stable reproduction?

mvz:
> Do you have a stable reproduction?

You can try the following:

```
git clone git@github.com:troessner/reek.git
cd reek
git checkout 3dd6cc9d
bundle
bundle exec mutant --include lib --require reek --use rspec --since d9297feb "Reek::Smells*"`
```

mbj:
@mvz I can confirm its a reproduction of the crash/stuck problem.

Debugging shows that the killfork hangs in a futex, while handling the stacktrace..

```
$ strace -p 22350
futex(0x7ff29c000020, FUTEX_WAIT_PRIVATE, 2, NULL
```

It does not occur on manually inserting the mutations, likely because the VM is in a different state on mutation testing because of the killfork.

This is IMO a ruby quality problem, not mutant. As I do not perform any "unusual" unix API calls (just, `fork`, `popen`, `write`, `close`) for the killfork. Nor does mutant use anything else but plain ruby.

A mitigation mutant could do is adding mutant side timeouts, something I always hated to do for various reasons. I'll think about it.

PR #927, opened 2016-05-29
Allow building branches only for the master branch

mvz:
By using this setting, we can turn building for pushes back on in Travis CI, so we can have an actual build history for master.


troessner:
Good idea ;)

PR #928, opened 2016-05-30
Remove unused fallback parameter from config handling

mvz:
The fallback value is already set in the (default) configuration, so there is no need to specify it again when fetching the value.

### DIFF
--- a/lib/reek/smells/data_clump.rb
+++ b/lib/reek/smells/data_clump.rb
@@ -53,8 +53,8 @@ module Reek
       #
       # :reek:FeatureEnvy
       def inspect(ctx)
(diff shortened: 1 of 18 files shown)


mvz:
Build failures are mutant-related, so @troessner this is ready for review.

mvz:
This change was discovered with mutant, by the way :+1:.

troessner:
>> This change was discovered with mutant, by the way

Because the corresponding mutants that removed the fallback parameter were not killed by specs even though it was spec'ed?

Nice improvement, merged!

mvz:
> Because the corresponding mutants that removed the fallback parameter were not killed by specs even though it was spec'ed?

Yes. There was only a spec for the behavior inside SmellRepository.

PR #931, opened 2016-05-31
Make build steps depend on each other.

### LINKED ISSUE #929: Fail CI tests at first opportunity
Travis will now actually run ataru and mutant even if the specs and/or cukes fail. This is silly and leads to problems such as current master failing the build. We should make these steps dependent.

troessner:
Fixes #929


mvz:
That's an interesting message from Git from inside Mutant there.

troessner:
Travis greeeeeeen! Merge? :)

Issue #836, opened 2016-01-22
Investigate Ruby 2.3 syntax consequences


chastell:
We should handle the new syntax introduced in Ruby 2.3.

- [ ] `&.` safe navigation
- [ ] `<<~` squiggly heredoc

chastell:
(In the future we might want to introduce a code smell that suggests turning on frozen string literals in per-file comments, but I’d make it a separate issue.)

seanhandley:
:+1:

I'm being held back from using the new 2.3 syntax features because my build process relies on being able to run rubycritic.

troessner:
"RubyCritic"? Wrong browser tab? ;)

Reek should be able to deal with 2.3 syntax already. This issue is more about if and how we should handle new node types like for the safe navigation operator. We just ignore it right now but it shouldnt raise any errors.

seanhandley:
@troessner I get an error when running rubycritic because it has a hard dependency on Reek. See the below stack trace:

```
  app/models/user.rb:4 :: parse error on value ":" (tCOLON)
  skipping app/models/user.rb
ERROR: parsing ruby file app/models/user.rb
Racc::ParseError: app/models/user.rb:4 :: parse error on value ":" (tCOLON) at:
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/2.3.0/racc/parser.rb:528:in `on_error'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/ruby_parser-3.7.3/lib/ruby_parser_extras.rb:1125:in `on_error'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/2.3.0/racc/parser.rb:259:in `_racc_do_parse_c'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/2.3.0/racc/parser.rb:259:in `do_parse'
  /Users/sean/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/ruby_parser-3.7.3/lib/ruby_parser_extras.rb:1037:in `block in process'
ERROR: parsing ruby file app/models/user.rb
```

My code runs fine with the lonely operator but it's reek's use of ruby-parser, (and in turn racc) that generates language parse errors.

troessner:
`Reek` uses `parser`, *not* `ruby_parser`.
Please show me the full stacktrace, the actual call and relevant version (Reek & Rubycritic)

seanhandley:
[...]

PR #932, opened 2016-06-01
Update parser.

troessner:


reek.gemspec
```diff
@@ -22,6 +22,6 @@ Gem::Specification.new do |s|
   s.summary = 'Code smell detector for Ruby'
 
   s.add_runtime_dependency 'codeclimate-engine-rb', '~> 0.3.1'
-  s.add_runtime_dependency 'parser',                '~> 2.3', '>= 2.3.0.6'
+  s.add_runtime_dependency 'parser',                '~> 2.3.1'
```
mvz: Let's ensure that at least `2.3.1.1` is used since it includes the fix for [this safe navigation issue](https://github.com/troessner/reek/issues/836#issuecomment-222181329).

troessner: Updated!

PR #940, opened 2016-06-06
Add `name` to `parameters` for `SmellWarnings`.

### LINKED ISSUE #939: json output seems to miss "name" property on readme-example
This is my output when I run reek with json output over your example from the readme:
```
[{"context":"Dirty#awful",
  "lines":[3],
  "message":"has boolean parameter 'log'",
  "smell_type":"BooleanParameter",
  "source":"STDIN",
  "name":"log",
  "wiki_link":"https://github.com/troessner/reek/blob/master/docs/Boolean-Parameter.md"},
 {"context":"Dirty",
  "lines":[1],
  "message":"has no descriptive comment",
  "smell_type":"IrresponsibleModule",
  "source":"STDIN",
  "name":"Dirty",
  "wiki_link":"https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md"},
 {"context":"Dirty#awful",
  "lines":[3],
  "message":"has 4 parameters",
  "smell_type":"LongParameterList",
  "source":"STDIN",
  "count":4,
  "wiki_link":"https://github.com/troessner/reek/blob/master/docs/Long-Parameter-List.md"},
 {"context":"Dirty#awful",
  "lines":[3],
  "message":"has the parameter name 'x'",
  "smell_type":"UncommunicativeParameterName",
  "source":"STDIN",
  "name":"x",
  "wiki_link":"https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Parameter-Name.md"},
 {"context":"Dirty#awful",
  "lines":[5],
  "message":"has the variable name 'w'",
  "smell_type":"UncommunicativeVariableName",
  "source":"STDIN",
  "name":"w",
  "wiki_link":"https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Variable-Name.md"},
 {"context":"Dirty#awful",
  "lines":[3],
  "message":"has unused parameter 'log'",
  "smell_type":"UnusedParameters",
  "source":"STDIN",
  "name":"log",
  "wiki_link":"https://github.com/troessner/reek/blob/master/docs/Unused-Parameters.md"},
 {"context":"Dirty#awful",
  "lines":[3],
  "message":"has unused parameter 'offset'",
  "smell_type":"UnusedParameters",
  "source":"STDIN",
  "name":"offset",
  "wiki_link":"https://github.com/troessner/reek/blob/master/docs/Unused-Parameters.md"},
 {"context":"Dirty#awful",
  "lines":[3],
  "message":"has unused parameter 'y'",
  "smell_type":"UnusedParameters",
  "source":"STDIN",
  "name":"y",
  "wiki_link":"https://github.com/troessner/reek/blob/master/docs/Unused-Parameters.md"}]
```
The third one got a "count" attribute instead of "name". Shouldn't be there a name actually, otherwise you don't know to which function is referred to?

troessner:
Fixes #939 

@mvz wdyt? I believe it does make indeed sense to add this everywhere.

I'll stop here to get your feedback first.

Missing so far:

* Everything below "TooManyInstanceVariables"
* DataClump (kind of missed that)

### DIFF
--- a/lib/reek/smells/long_parameter_list.rb
+++ b/lib/reek/smells/long_parameter_list.rb
@@ -43,7 +43,7 @@ module Reek
           context: ctx,
           lines: [exp.line],
           message: "has #{count} parameters",
-          parameters: { count: count })]
+          parameters: { count: count, name: ctx.name })]
       end
     end
   end
--- a/lib/reek/smells/utility_function.rb
+++ b/lib/reek/smells/utility_function.rb
@@ -68,7 +68,7 @@ module Reek
           context: ctx,
           lines: [ctx.exp.line],
           message: "doesn't depend on instance state (maybe move it to another class?)",
-          parameters: { name: ctx.full_name })]
+          parameters: { name: ctx.name.to_s })]
       end
 
       private
--- a/spec/reek/smells/prima_donna_method_spec.rb
+++ b/spec/reek/smells/prima_donna_method_spec.rb
@@ -3,28 +3,19 @@ require_lib 'reek/context/module_context'
 require_relative 'smell_detector_shared'
 
 RSpec.describe Reek::Smells::PrimaDonnaMethod do
+  it 'reports the right values' do
+    src = 'class C; def m!; end; end'
+    expect(src).to reek_of :PrimaDonnaMethod,
+                           lines: [1],
+                           name: 'C',
+                           message: 'has prima donna method `m!`'
+  end
+
   it 'should report nothing when method and bang counterpart exist' do
     expect('class C; def m; end; def m!; end; end').not_to reek_of(:PrimaDonnaMethod)
   end
 
   it 'should report PrimaDonnaMethod when only bang method exists' do
     expect('class C; def m!; end; end').to reek_of(:PrimaDonnaMethod)
-  end
-
-  describe 'the right smell' do
-    let(:detector) { build(:smell_detector, smell_type: :PrimaDonnaMethod) }
-    let(:src)      { 'class C; def m!; end; end' }
-    let(:ctx)      do
-      Reek::Context::ModuleContext.new(nil,
-                                       Reek::Source::SourceCode.from(src).syntax_tree)
-    end
-
-    it 'should be reported' do
-      smells = detector.inspect(ctx)
-      warning = smells[0]
-
-      expect(warning.smell_type).to eq('PrimaDonnaMethod')
-      expect(warning.lines).to eq([1])
-    end
   end
 end
--- a/lib/reek/smells/long_yield_list.rb
+++ b/lib/reek/smells/long_yield_list.rb
@@ -36,7 +36,7 @@ module Reek
             context: ctx,
             lines: [yield_node.line],
             message: "yields #{count} parameters",
-            parameters: { count: count })
+            parameters: { count: count, name: ctx.name })
         end
       end
     end
--- a/lib/reek/smells/module_initialize.rb
+++ b/lib/reek/smells/module_initialize.rb
@@ -28,7 +28,8 @@ module Reek
               smell_warning(
                 context: ctx,
                 lines:   [ctx.exp.line],
-                message: 'has initialize method')
+                message: 'has initialize method',
+                parameters: { name: ctx.name })
             ]
           end
         end
--- a/lib/reek/smells/nil_check.rb
+++ b/lib/reek/smells/nil_check.rb
@@ -18,7 +18,8 @@ module Reek
           smell_warning(
             context: ctx,
             lines: [node.line],
-            message: 'performs a nil-check')
+            message: 'performs a nil-check',
+            parameters: { name: ctx.name })
         end
       end
 
--- a/lib/reek/smells/prima_donna_method.rb
+++ b/lib/reek/smells/prima_donna_method.rb
@@ -46,11 +46,11 @@ module Reek
         end
 
         return if version_without_bang
-
         smell_warning(
           context: ctx,
           lines: [ctx.exp.line],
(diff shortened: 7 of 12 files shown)


troessner:
@mvz  wdyt?

mvz:
Oh, I meant to think more on this one and then forgot about it. Sorry.

I'm not so sure this makes sense. As I understand it, name is for an extra something that is relevant to the particular smell. So, for UnusedParameters it's the name of the parameter. As I understand it, your proposal is to put `ctx.name` wherever name is not currently defined. But the context name is already present in `context`.

troessner:
Well, the problem is right now that we are very inconsistent when it comes to the `name` attribute. Sometimes we set it, sometimes we don't, even if it makes sense. Dropping it completely would be possible but if we can make life easier for clients we should do it, dont you think?
I believe both `context` and `name` are valuable on its own. 
Question is how to make it consistent and i'd rather just output `name` for every smell type cause there' s no harm in it, is it?
Put differently, what are your concrete objections?

mvz:
I think name only makes sense if it is different from context.

troessner:
>> I think name only makes sense if it is different from context.

Ack - I'm gonna look into this.

troessner:
Ready for review! I fixed most problems mutant pointed out but it is still not 100% satisfied. Looking at the 2 remaining complaints I think it's ok to ignore them.

lib/reek/smells/long_parameter_list.rb
```diff
@@ -43,7 +43,7 @@ def inspect(ctx)
           context: ctx,
           lines: [exp.line],
           message: "has #{count} parameters",
-          parameters: { count: count })]
+          parameters: { count: count, name: ctx.name })]
```
mvz: Hm .. but now name is the same as context, isn't it?

lib/reek/smells/utility_function.rb
```diff
@@ -68,7 +68,7 @@ def inspect(ctx)
           context: ctx,
           lines: [ctx.exp.line],
           message: "doesn't depend on instance state (maybe move it to another class?)",
-          parameters: { name: ctx.full_name })]
+          parameters: { name: ctx.name.to_s })]
```
mvz: Ah, here name was already the same as context. Which makes it superfluous, I think.

spec/reek/smells/prima_donna_method_spec.rb
```diff
@@ -3,28 +3,19 @@
 require_relative 'smell_detector_shared'
 
 RSpec.describe Reek::Smells::PrimaDonnaMethod do
+  it 'reports the right values' do
+    src = 'class C; def m!; end; end'
+    expect(src).to reek_of :PrimaDonnaMethod,
+                           lines: [1],
+                           name: 'C',
```
mvz: Here, it would make more sense to me if name was `m`.

mvz:
I think we should think and discuss a bit about what information we want to include in the JSON and YAML exactly, and, most importantly, what each key means. Those keys grew organically during Reek's development and were only really used internally, so it makes sense that they didn't become consistent.

mvz:
Nice spec cleanups, by the way! :+1:

lib/reek/smells/long_parameter_list.rb
troessner: Nope, we're calling `full_name` in the `smell_warning` method, so that's different. 
I was confused by this as well, but we should fix this separately.

lib/reek/smells/utility_function.rb
troessner: I'd rather go the other way round: Before this change it was a bug.
Because before `name` would have been "Foo::Bar#my_method" for instance.
Now it's just "my_method" and the client can decide if he wants to use the `context` or the `name`.

spec/reek/smells/prima_donna_method_spec.rb
troessner: But this wouldn't make much sense AFAIS. [1] the message below even says "has prima donna method ...." and [2] the smell detector operates on class level, not on method level.
Sure we can fix [1] but [2] still suggests that the name should be 'C' here, not 'm'.

troessner:
Addressed all comments. Merge or do you want to discuss this further @mvz ?

mvz:
I'll take another look at this tomorrow. I'm still not sure what we want 'name' to mean.

lib/reek/smells/utility_function.rb
mvz: Yes, this makes sense.

mvz:
All right, I took a look at how `:name` and parameters in general are currently used, and most of the time, `:name` is the name of something that is not the context, **or** the smell really concerns the context's actual name (as in UncommunicativeModuleName).

Looking at it from that perspective, we should simply remove `:name` from UtilityFunction, NestedIterators and IrresponsibleModule.

Another way of looking at it, is that `:parameters` should contain all values that we interpolate into the smell warning message. No more, no less.

Regarding the problem in #939, yes, you need to look at the `:context` of LongParameterList to see which function the smell refers to, but that is *also* the case for UncommunicativeParameterName, the fourth item, which *has* the `:name` parameter. Adding `:name` everywhere and making it only sometimes be the thing that smells would only add to the confusion I think.

Perhaps `:context` simply isn't such a great name, and the real solution is to rename that?

troessner:
>> Another way of looking at it, is that :parameters should contain all values that we interpolate into the smell warning message. No more, no less.

Hmm, I like the idea!
Simple and consistent. Let's go for that.
Deal?

mvz:
> Deal?

Deal!

troessner:
Awesome, I'll pick this up again as soon as I can ;)

troessner:
Closing this one in favor of #993

PR #941, opened 2016-06-10
Add `ci` Rake task.

troessner:
Also add tasks for ataru and mutant.


troessner:
Whoopsie, i firgot to update the .travis.yml correspondingly. On the road now, will do that later.

PR #945, opened 2016-06-11
Subclassed from core class

waldyr:
This is a pull request into another pull request. Sorry for blending in but I'd like to help in the development of that smell detector. Even though it enhances the current smell detector it still have some issues to work on.

The previous detector did not account this kind of inheritance `A = Class.new(Array)`.


lib/reek/ast/sexp_extensions/constant_assignment.rb
```diff
@@ -0,0 +1,19 @@
...
+  module AST
+    module SexpExtensions
+      # Utility methods for :casgn nodes.
+      module CasgnNode
+        def assignment
+          children[2]
+        end
+
+        def class_creation?
+          assignment.respond_to?(:name) &&
```
mvz: Can we avoid this check somehow?

lib/reek/ast/sexp_extensions/constant_assignment.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+module Reek
+  module AST
+    module SexpExtensions
+      # Utility methods for :casgn nodes.
+      module CasgnNode
+        def assignment
```
mvz: I find this name confusing. Assignment to me means an expression of the form `a = b`. The thing returned from this method would be just `b`. Perhaps `value` would be a better name for this method.

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -6,26 +6,49 @@ module Reek
...
-        ancestor_namespace, ancestor = ancestor_node.to_a
-        return [] unless CORE_CLASSES.include?(ancestor) && ancestor_namespace.nil?
+        public_send(METHODS[ctx.type], ctx)
+      end
+
+      def inspect_class(ctx)
+        expression = ctx.exp
+        superclass = expression.superclass
+
+        return [] unless expression.superclass? && superclass.core_class?
```
mvz: You can just replace `expression.superclass?` with `superclass` here.

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -6,26 +6,49 @@ module Reek
...
+          lines: [ctx.exp.line],
+          message: "inherits from a core class #{superclass}",
+          parameters: { ancestor: superclass.name })]
+      end
+
+      def inspect_casgn(ctx)
+        expression = ctx.exp
+        assignment = expression.assignment
+
+        return [] unless expression.class_creation? && assignment.emitted.core_class?
```
mvz: Since `expression` has a method `class_creation?`, you can also add a method `superclass`, which will allow you to unify `inspect_casgn` and `inspect_class`.

lib/reek/ast/sexp_extensions/send.rb
```diff
@@ -10,6 +10,10 @@ def receiver
           children.first
         end
 
+        def emitted
```
mvz: I'm not sure what this name means.

mvz:
Thanks! This looks good in general, I just have some nitpicks. I think @andyw8 should be the one who decides how to proceed with this.

mvz:
Oh, it also looks like you broke some cucumber scenarios.

troessner:
>> Thanks! This looks good in general, I just have some nitpicks. I think @andyw8 should be the one who decides how to proceed with this.

Agreed. But we should act swift before this one goes stale as well.

waldyr:
With this PR I was trying to make subclassed-from-core-class move faster. Sorry for those problems, even though the code base is awesome I didn't quite understand or find the best way to do some things and I'll evidence them later meanwhile I fix the suggestions. I won't let this goes stale! 👍

lib/reek/ast/sexp_extensions/constant_assignment.rb
waldyr: This is a brute way to avoid exceptions while calling `class_creation?` on something like:
```
s(:casgn, nil, :Iterator,
  s(:block,
    s(:send,
      s(:const, nil, :Struct), :new,
      s(:sym, :exp),
      s(:sym, :depth)),
    s(:args),
    s(:begin,
      s(:send, nil, :include,
        s(:const, nil, :Comparable)),
      s(:def, :<=>,
        s(:args,
          s(:arg, :other)),
        s(:send,
          s(:send, nil, :depth), :<=>,
          s(:send,
            s(:lvar, :other), :depth))))))
```

This is an example of class_creation:
```
s(:casgn, nil, :Dummy,
  s(:send,
    s(:const, nil, :Class), :new,
    s(:const, nil, :Array)))
```

Can you help me with something more robust?

waldyr:
> Since expression has a method class_creation?, you can also add a method superclass, which will allow you to unify inspect_casgn and inspect_class.

This isn't fixed yet.

mvz:
> Sorry for the commented problems

No need to apologise, this is what reviews are for.

>  I won't let this goes stale!

Excellent :+1:

mvz:
I'll have a look at `class_creation?` later in the week.

troessner:
Let's get this bad boy going in favor of #917 !

Looking forward to this feature @waldyr !

Anything else you need help with besides what @mvz is looking into?

waldyr:
> Looking forward to this feature @waldyr !
Hahah. We all do!

> Anything else you need help with besides what @mvz is looking into?
Everything is fine. While @mvz come up with something better for `class_creation?` I'll probably tackle the broken build.

waldyr:
@troessner I couldn't make one cucumber scenario pass. I also didn't understand what is wrong with it. Explaining it would be helpful :)

mvz:
@waldyr since that step tests the exit code, it probably means reek encountered an error during that run. If you run just that scenario on your own machine, you should then be able to `cd` into `tmp/aruba` and run `bundle exec reek` there and look at the result.

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -4,28 +4,55 @@
...
 
       # Checks +ctx+ for if it is subclasssed from a core class
       #
       # @return [Array<SmellWarning>]
       def inspect(ctx)
-        _class_node, ancestor_node, _body_node = ctx.exp.to_a
-        return [] if ancestor_node.nil?
-        ancestor_namespace, ancestor = ancestor_node.to_a
-        return [] unless CORE_CLASSES.include?(ancestor) && ancestor_namespace.nil?
+        public_send(METHODS[ctx.type], ctx)
```
mvz: I would prefer just a simple `if` or `case` here, since there are only two possible values for `ctx.type`.

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -4,28 +4,55 @@
...
+      end
+
+      def inspect_class(ctx)
+        superclass = ctx.exp.ancestor
+
+        return [] unless superclass && superclass.core_class?
```
mvz: My gut feeling is that I would prefer knowledge of what core classes are to live in the smell detector.

lib/reek/ast/sexp_extensions/constant_assignment.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+module Reek
+  module AST
+    module SexpExtensions
+      # Utility methods for :casgn nodes.
+      module CasgnNode
```
mvz: This module is also defined in `lib/reek/ast/sexp_extensions/module.rb`. You may want to merge these definitions.

lib/reek/ast/sexp_extensions/constant_assignment.rb
```diff
@@ -0,0 +1,23 @@
...
+        def send_expression
```
mvz: This expression is defined as `#value` in `lib/reek/ast/sexp_extensions/module.rb`

lib/reek/ast/sexp_extensions/constant_assignment.rb
troessner: Then lets move that out of Module :)

lib/reek/ast/sexp_extensions/constant_assignment.rb
```diff
@@ -0,0 +1,23 @@
...
+          children[2]
+        end
+
+        def ancestor
+          send_expression.value_assigned
+        end
+
+        def class_creation?
```
mvz: You can use `defines_module?` from `lib/reek/ast/sexp_extensions/module.rb` to simplify this method.

lib/reek/smells/subclassed_from_core_class.rb
troessner: Yes  strong +1 from my side. The detector should hold this knowledge.

lib/reek/ast/sexp_extensions/send.rb
```diff
@@ -10,6 +10,10 @@ def receiver
           children.first
         end
 
+        def value_assigned
```
mvz: Since this method does not makes sense for all `:send` nodes, you should move this logic to its caller.

lib/reek/ast/sexp_extensions/constant_assignment.rb
```diff
@@ -0,0 +1,23 @@
...
+module Reek
+  module AST
+    module SexpExtensions
+      # Utility methods for :casgn nodes.
+      module CasgnNode
+        def send_expression
+          children[2]
+        end
+
+        def ancestor
```
mvz: This method does not make sense for all constant assignments, so I think you should move its logic to the smell detector.

mvz:
@waldyr see my inline comment about `class_creation?`. I also added some more comments about other bits.

troessner:
@waldyr can you please rebase against master? Your codebase seems to be not really up to date.

mvz:
@troessner rebasing against master may not make sense since this pull request targets the `subclassed-from-core-class` branch.

Hmm .. maybe we can rebase that one first, and then @waldyr can rebase on the rebased branch. :confounded:

mvz:
Perhaps we should first just review this as-is, and only rebase on master at the end?

troessner:
How about keeping it simple:

* we merge waldyr's branch with Andy's branch into a new branch that was created of a fresh master
* we close the other 2 open PRs

This would give us a fresh, clean start. Wdyt?

waldyr:
> Hmm .. maybe we can rebase that one first, and then @waldyr can rebase on the rebased branch.

This seems to be the correct way, but I'm fine with whatever you guys decide. 👍

troessner:
>> Hmm .. maybe we can rebase that one first, and then @waldyr can rebase on the rebased branch.

This would be great if we cared about the branch history. I believe we don't.
I'd rather get this restarted french. I also believe that my way is faster in this case :)

mvz:
> This would be great if we cared about the branch history. I believe we don't.

No, we don't. I was trying to figure out a way to keep all discussions together, i.e., avoid having to open a new PR.

troessner:
>> No, we don't.

You mean, "we DO"? This gets confusing 😆 
You guys were more involved in this PR so I'll just shut up, please pick whatever makes sense for you ;)

mvz:
:laughing: I meant I *do not* care about commits, but I *do* care about comments.

mvz:
Alright, we can discuss this endlessly, but I don't see any solution that will:
* Result in a pull request we can merge into master
* Preserve comments

So, let's go with @troessner's first suggestion: @waldyr can you rebase your branch on master? This will include @andyw8's commits but that's no problem. Then, please open a new PR. If we need access to the old review comments we'll just have to look at the old PRs.

Issue #947, opened 2016-06-15
Undefined method `name' for s(:lambda)


mvz:
From https://github.com/troessner/reek/issues/942#issuecomment-226103640:

> this is also happening to me... was using an old version, upgraded to 4.0.4 and it stopped working; I can't determine what is triggering this...

> ```
⚡ reek
$HOME/.rvm/gems/ruby-2.2.1/gems/reek-4.0.4/lib/reek/smells/nested_iterators.rb:139:in `block in ignored_iterator?': undefined method `name' for s(:lambda)
:#<Class:0x000000059b7670> (NoMethodError)
        from $HOME/.rvm/gems/ruby-2.2.1/gems/reek-4.0.4/lib/reek/smells/nested_iterators.rb:139:in `any?'
        from $HOME/.rvm/gems/ruby-2.2.1/gems/reek-4.0.4/lib/reek/smells/nested_iterators.rb:139:in `ignored_iterator?'
        from $HOME/.rvm/gems/ruby-2.2.1/gems/reek-4.0.4/lib/reek/smells/nested_iterators.rb:130:in `increment_depth'
        from $HOME/.rvm/gems/ruby-2.2.1/gems/reek-4.0.4/lib/reek/smells/nested_iterators.rb:110:in `block in scout'
        from $HOME/.rvm/gems/ruby-2.2.1/gems/reek-4.0.4/lib/reek/smells/nested_iterators.rb:109:in `each'
        from $HOME/.rvm/gems/ruby-2.2.1/gems/reek-4.0.4/lib/reek/smells/nested_iterators.rb:109:in `flat_map'
```

nachokb:
Oh I see, just created a PR.

PR #948, opened 2016-06-15
Fix crash when NestedIterators found a lambda #947

nachokb:
When NestedIterators found a (failing) lambda, it crashed. Convoluted example:

```ruby
def fred()
  mult = ->(one, two) {(1..one).map { |n| two}.sum}
  mult.call(4, 3)
end
```


troessner:
Closing this one since #949 is merged now. Thanks for looking into it though!

nachokb:
Thank you for Reek

PR #954, opened 2016-06-18
Subclassed from core class

waldyr:
This is a PR that addresses #907

This PR contains work from the following pull requests: #917 and #945.

This PR aims to detect smells from classes that inherits from core classes such as Array, String and etc.


waldyr:
Comments addressed. Green Build. Waiting for review and probably some commits' fixup.

@mvz @troessner Thanks for everything. No words can describe how awesome was to work on reek's codebase.

features/samples.feature
```diff
@@ -129,6 +129,8 @@ Feature: Basic smell detection
       RepeatedConditional: OptionParser tests default_pattern at least 7 times [https://github.com/troessner/reek/blob/master/docs/Repeated-Conditional.md]
       RepeatedConditional: OptionParser tests not_style at least 3 times [https://github.com/troessner/reek/blob/master/docs/Repeated-Conditional.md]
       RepeatedConditional: OptionParser tests s at least 7 times [https://github.com/troessner/reek/blob/master/docs/Repeated-Conditional.md]
+      SubclassedFromCoreClass: OptionParser::CompletingHash inherits from a core class Hash [https://github.com/troessner/reek/blob/master/docs/Subclassed-From-Core-Class.md]
+      SubclassedFromCoreClass: OptionParser::OptionMap inherits from a core class Hash [https://github.com/troessner/reek/blob/master/docs/Subclassed-From-Core-Class.md]
```
troessner: The message is a little odd, how about "from a core class (Hash)"?

features/smells/subclassed_from_core_class.feature
```diff
@@ -0,0 +1,20 @@
+Feature: Smell - SubclassedFromCoreClass
+  Scenario: Configure multiple directories
```
troessner: That scenario title is highly misleading. That our configuration works across multiple directory is a feature that is independent from SmellDetectors. How about just "Reports smell"?
You can then also simplify the paths below, "web_app/app/..." is not necessary.

lib/reek/ast/sexp_extensions/module.rb
```diff
@@ -45,28 +45,45 @@ module ModuleNode
       # Utility methods for :class nodes.
       module ClassNode
         include ModuleNodeBase
+
         def superclass() children[1] end
+        def ancestor() superclass end
```
troessner: Let's rather use `alias_method` to communicate our intention clearer.

lib/reek/ast/sexp_extensions/module.rb
```diff
@@ -45,28 +45,45 @@ module ModuleNode
...
 
+        def module_definition
+          return nil unless value
+
+          case value.type
+          when :block
+            value.call
+          when :send
+            value
+          end
```
troessner: Let's add an `else` that raises RuntimeError to make sure we don't silently swallow errors.

lib/reek/ast/sexp_extensions/module.rb
```diff
@@ -45,28 +45,45 @@ module ModuleNode
...
-                   value.call
-                 when :send
-                   value
-                 end
+          call = module_definition
           call && call.module_creation_call?
         end
 
+        def module_definition
+          return nil unless value
```
troessner: Can you please add a comment above when you expect `value`  to be nil?

lib/reek/ast/sexp_extensions/module.rb
```diff
@@ -45,28 +45,45 @@ module ModuleNode
...
-                 when :block
-                   value.call
-                 when :send
-                   value
-                 end
+          call = module_definition
           call && call.module_creation_call?
         end
 
+        def module_definition
```
troessner: Would be great if you could add a comment above what the expected AST could look like. Just a simple example.
I intend to do that throughout the code base in the long run so let's start right here ;)

lib/reek/ast/sexp_extensions/send.rb
```diff
@@ -10,6 +10,10 @@ def receiver
           children.first
         end
 
+        def value_assigned
```
troessner: Not a fan of this name I have to say. Shouldn't it be "assigned_value"? Also, we're talking about the "right hand assigment", aren't we? Then how about calling it "right_hand_assignment"?

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -0,0 +1,64 @@
...
+      #
+      # @return [Array<SmellWarning>]
+      def inspect(ctx)
+        public_send(METHODS[ctx.type], ctx) || []
+      end
+
+      def inspect_class(ctx)
+        superclass = ctx.exp.ancestor
+
+        inspect_superclass(ctx, superclass.name) if superclass
```
troessner: Given how Ruby's object system is set up, only BasicObject doesnt have a superclass.
Wdyt about checking rather for BasicObject instead to make this clearer?

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      def self.contexts
+        [:class, :casgn]
+      end
+
+      # Checks +ctx+ for if it is subclasssed from a core class
+      #
+      # @return [Array<SmellWarning>]
+      def inspect(ctx)
+        public_send(METHODS[ctx.type], ctx) || []
```
troessner: I believe that's a tad over optimized here. inspect_class and inspect_casgn are almost identical and contain some duplication.
How about dissolving both of them and handle everything in "inspect"? I dont think this will blow up "inspects" method size and i believe it will be clearer in the end.

spec/reek/smells/subclassed_from_core_class_spec.rb
```diff
@@ -0,0 +1,86 @@
+require_relative '../../spec_helper'
+require_lib 'reek/smells/subclassed_from_core_class'
+require_relative 'smell_detector_shared'
+
+RSpec.describe Reek::Smells::SubclassedFromCoreClass do
+  let(:detector) { described_class.new }
+
+  it_should_behave_like 'SmellDetector'
+
+  it 'should not report if the class has no ancestor' do
```
troessner: Thats a little misleading dince it does have a superclass / ancestor. How about "does not inherit from a core class"?

spec/reek/smells/subclassed_from_core_class_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    src = <<-EOS
+      module Namespace
+        class Dummy
+          Dummiest = Class.new(Array)
+
+          def initialize
+          end
+
+          def test
+          end
```
troessner: Nice specs overall, but whags the point of a class having methods or not? Cant we just ....not care?

lib/reek/smells/subclassed_from_core_class.rb
waldyr: If I understand correctly, I think this is not possible due the expressions that do not explicitly inherit from a class return `nil` instead of BasicObject.

troessner: Ah, yeah, you're right, nevermind!

lib/reek/ast/sexp_extensions/module.rb
waldyr: `alias_method` raises a Rubocop offense. It says to use `alias` instead.

lib/reek/smells/subclassed_from_core_class.rb
waldyr: Tried other options and always get different offenses from Reek.

lib/reek/ast/sexp_extensions/module.rb
waldyr: I'll have to study the codebase little more to say that. I just extracted this method from `defines_module?`.

lib/reek/ast/sexp_extensions/module.rb
waldyr: Good news: I'll add this `else`. Bad news: We already swallow errors 😄 . Build is going to break again.

troessner: I think a 

```Ruby
else
  raise RuntimeError, "Can't process type #{value.type} for #{value}"
end
```

would be already an improvement we should do right away :)

lib/reek/ast/sexp_extensions/module.rb
troessner: Yep, that's the problem with our codebase. Things like that are all over the place and you have no idea why 😆 
I normally then open pry and debug my way through which is time consuming and tedious.
Unfortunately AST and especially Ruby's AST is not self-explanatory or trivial.
My goal is to explain all those place in our code because this trips you up all the time.
So let's start here ;)

lib/reek/smells/subclassed_from_core_class.rb
troessner: Ok, then let's just leave it like it. It is pretty straightforward and readable after all.

troessner:
@waldyr almost there! Let's just add a couple of explanatory comments regarding the expected AST structure (see my 2 comments above) and then this bad boy is ready for merge!
Speaking of, might sense for you to squash your commits now already (required anyway before we can merge).

lib/reek/ast/sexp_extensions/module.rb
mvz: I thought the specs might help, but it just says 'with a single assignment' for a sexp where value is not there :crying_cat_face:.

lib/reek/ast/sexp_extensions/module.rb
mvz: I think the intent was for `call` to be blank if the type is neither `:send` nor `:block`, e.g., if the value is an integer literal.

lib/reek/smells/subclassed_from_core_class.rb
mvz: Ugh, no. This makes it look like something complicated is going on, where there really isn't. If a simple `if` or `case` makes Reek detect a smell here we should either suppress it or tweak Reek.

lib/reek/ast/sexp_extensions/send.rb
```diff
@@ -10,6 +10,10 @@ def receiver
           children.first
         end
 
+        def right_hand_value
```
mvz: This name makes no sense for most cases of `:send` nodes.

lib/reek/ast/sexp_extensions/module.rb
```diff
@@ -45,28 +45,46 @@ module ModuleNode
...
+
+          case value.type
+          when :block
+            value.call
+          when :send
+            value
+          end
+        end
+
+        def class_definition
```
mvz: Doesn't this return the *superclass* for the class definition?

lib/reek/ast/sexp_extensions/module.rb
```diff
@@ -45,28 +45,46 @@ module ModuleNode
...
+          return nil unless module_definition
+
+          module_definition.right_hand_value
```
mvz: Let's keeps things simple and use `module_definition.args.first`.

lib/reek/ast/sexp_extensions/module.rb
mvz: I *think* this method returns more than just any module creation calls. For example, it would return a value for the code

```
Foo = foo(42)
```

Can we think of a better name?

lib/reek/ast/sexp_extensions/module.rb
mvz: Ah, there's an explanation of the possible `:casgn` ASTs further down in the file.

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      def inspect_class(ctx)
+        superclass = ctx.exp.ancestor
```
mvz: Why not just use the existing `#superclass` method here and avoid the alias in ClassNode?

mvz:
Ah, I was wondering how:
* The build could be green (i.e., Mutant found nothing), and
* There being some methods that seem to be unused.

The reason is that our new setup doesn't seem to fail when Mutant fails :laughing:.

@waldyr can you take a look at the mutant output in the build? You should be able to simplify some of the code. Shout if you need help.

lib/reek/ast/sexp_extensions/send.rb
waldyr: With `args.first` this method isn't needed.

lib/reek/ast/sexp_extensions/module.rb
waldyr: What do you think about `constant_definition`?

waldyr:
@troessner @mvz It's been a long PR 😅 . But I won't give up until I deliver the feature with quality.

What do you guys think of this version?

PS: @mvz I understand the concept of Test Mutants, however I do not understand a thing in this output, sorry 😖

features/smells/subclassed_from_core_class.feature
```diff
@@ -0,0 +1,14 @@
...
+      # The MyHash class
+      class MyHash < Hash
+      end
+      """
+    When I run `reek my_hash.rb`
+    Then it reports:
+    """
+    my_hash.rb -- 1 warning:
+      [2]:SubclassedFromCoreClass: MyHash inherits from a core class (Hash) [https://github.com/troessner/reek/blob/master/docs/Subclassed-From-Core-Class.md]
+    """
```
troessner: Yes, feature looking good now ;)

lib/reek/ast/sexp_extensions/module.rb
troessner: Ok, understood, then let's just leave it.

lib/reek/report/code_climate/code_climate_configuration.yml
```diff
@@ -662,3 +662,4 @@ UtilityFunction:
   remediation_points: 250_000
   content: |
     A _Utility Function_ is any instance method that has no dependency on the state of the instance.
+# TODO: add entry for SubclassedFromCoreClass
```
troessner: Let's use the same remediation_points as for UtilityFunction above and add a short description. We can just c&p the description from the smell detector below, so : https://github.com/troessner/reek/pull/954/files#diff-8c155dc71177f640b21a8834d1aec4abR8

lib/reek/smells/subclassed_from_core_class.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+      # Checks +ctx+ for either expressions:
+      #
+      # Foo = Class.new(Bar)
+      #
+      # class Foo < Bar; end;
+      #
+      # @return [Array<SmellWarning>]
+      def inspect(ctx)
```
troessner: Looks good to me now! Easy, straightforward and readable!

spec/reek/smells/subclassed_from_core_class_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+  it 'should not report if inner class inherits from allowed classes' do
+    src = <<-EOS
+      module Namespace
+        class Dummy
+          Dummiest = Class.new(StandardError)
+        end
+      end
+    EOS
+    expect(src).to_not reek_of(:SubclassedFromCoreClass, ancestor: 'StandardError')
+  end
```
troessner: Excellent specs, kudos!

troessner:
Looks good now for me!
Anything from your side @mvz ?

@waldyr unless @mvz has further comments we now need to:

- add an entry for CodeClimate which is basically just a c&p (see my comment above)
- add a corresponding page under /docs. Should take more than 10 minutes, you can just take an existing one like the one for the [PrimaDonna smell](https://github.com/troessner/reek/blob/master/docs/Prima-Donna-Method.md), use the structure there and fill in the gaps
- squash this PR to one commit

Then I'll happily merge this one - awesome job!

mvz:
I'll have another look tomorrow.

mvz:
LGTM!

troessner:
Great, then lets get this finished @waldyr - please see my last comment above.

waldyr:
@troessner What's left?

troessner:
Ah, hadn't seen that you had already addressed this!

Highly awesome job, merged!

troessner:
And I just released Reek 4.1.0!

Issue #621, opened 2015-07-24
ExcessiveConstantUsage smell


troessner:
I think such a smell would be really helpful.

E.g. code like this:

```Ruby
class Omg
  FOO = ....
  BAR = ....
  WALDFEE = ....
end
```

would reek of this smell. The category for this smell would be "SRP violation" in my mind.

What are your thoughts on this?

mvz:
How would you resolve this smell if it occurs?

troessner:
>> How would you resolve this smell if it occurs?

For me ExcessiveConstantUsage is similar to (TooManyInstanceVariables)[https://github.com/troessner/reek/blob/master/docs/Too-Many-Instance-Variables.md]

chastell:
The devil, as usual, is in the detail. How many is too many? :smiling_imp:

troessner:
How about the same as for TooManyInstanceVariables, so 4?

chastell:
I’m not sure how well it’s going to play with your campaign against literals in code. ;)

troessner:
YMMD :)

troessner:
Cleaning up our issues: Linked this issue in #556 - closing this one.

mvz:
Reopening this since we now have a candidate implementation.

troessner:
Closing this one with #956 merged.

PR #956, opened 2016-06-23
Add Too Many Constants smell detector

waldyr:
#621 

Even thought this might still be in discussion I went ahead and implemented. It may help in the discussion. What do you guys think?

There's one spec that I've skipped because I didn't know what would be the correct context to inspect.

I used 5 as max quantity because I didn't want to fix [nested_iterators](https://github.com/troessner/reek/blob/master/lib/reek/smells/nested_iterators.rb) 😂 

Also I want to add Exception Handling into the discussion because It's pretty common to find code like this:
```
MyException = Class.new(StandardError)
```


spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+          class Dummy
+          end
+        end
+      EOS
+
+      expect(src).to reek_of(:TooManyConstants)
+    end
+  end
+
+  context 'smell report' do
```
mvz: You can combine these three expectations into one.

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def casgn_sequence(count: default_max_consts, consts: ('A'..'Z').to_a)
+    consts.first(count).map.with_index { |name, i| "#{name}=#{i}" }.join("\n")
+  end
+
+  context 'counting constants' do
+    it 'should not report for non-excessive constants' do
+      src = <<-EOS
+        class Dummy
+          #{casgn_sequence}
```
mvz: Having this indirection makes the test less readable I think, since it is not immediately clear how many constants are created here. Just writing it all out makes the tests a bit longer but a lot clearer.

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      src = <<-EOS
+        class Dummy
+          #{casgn_sequence(count: too_many_consts)}
+        end
+      EOS
+
+      expect(src).to reek_of(:TooManyConstants)
+    end
+
+    xit 'should report for excessive constants anywhere' do
```
mvz: This case is outside of any of the contexts you're checking. What *may* work is adding `:root` to the list of contexts, but I don't think we check within that context in any of the other detectors, so you'll be a bit in uncharted territory there.

lib/reek/smells/too_many_constants.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      #
+      # Checks +klass+ for too many constants.
+      #
+      # @return [Array<SmellWarning>]
+      #
+      def inspect(ctx)
+        max_allowed_consts = value(MAX_ALLOWED_CONSTS_KEY, ctx)
+
+        count = ctx.each_node(:casgn).map! { |casgn| casgn.children.last }.uniq.length
```
mvz: This looks very complex when I compare it to the specs.

spec/reek/smells/too_many_constants_spec.rb
mvz: Alternatively, you can at least pull the definition of `src` out into a `let` statement.

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    it 'reports a readable message' do
+      src = <<-EOS
+        module Moduly
+          class Klass
+            #{casgn_sequence(count: too_many_consts)}
+          end
+        end
+      EOS
+
+      expect(src).to reek_of(:TooManyConstants, message: "has at least #{too_many_consts} constants")
```
mvz: Be sure to also check at that it reports on the context `Moduly::Class`, and not on the context `Moduly`.

mvz:
> Also I want to add Exception Handling into the discussion because It's pretty common to find code like this:

> ```
> MyException = Class.new(StandardError)
> ```

I think that code should count as a class definition and not as a constant in the regular sense.

troessner:
Gonna review this bad boy as well in the next days!
@waldyr I just added you as "collaborateur" to this repo so no more need for forking the repo, you can create directly pull requests ;)

spec/reek/smells/too_many_constants_spec.rb
waldyr: I will need at least an if clause to treat things that is only inside the `:root` context. Feasible, but viable?

spec/reek/smells/too_many_constants_spec.rb
waldyr: Not a `let`'s fan but I'll improve this expectations

lib/reek/smells/too_many_constants.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      #
+      # Checks +klass+ for too many constants.
+      #
+      # @return [Array<SmellWarning>]
+      #
+      def inspect(ctx)
+        max_allowed_consts = value(MAX_ALLOWED_CONSTS_KEY, ctx)
+
+        count = ctx.each_node(:casgn).delete_if(&:constant_definition).length
```
waldyr: This makes me want to change the [`constant_definition`](https://github.com/waldyr/reek/blob/master/lib/reek/ast/sexp_extensions/module.rb#L90)

defaults.reek
```diff
@@ -115,3 +119,4 @@ UnusedPrivateMethod:
 UtilityFunction:
   enabled: true
   exclude: []
+  public_methods_only: false
```
troessner: This doesnt belong here but should go away after a rebase.

docs/Too-Many-Constants.md
```diff
@@ -0,0 +1,37 @@
...
+  CONST_3 = :dummy
+  CONST_4 = :dummy
+end
+```
+
+Reek would emit the following warning:
+
+```
+test.rb -- 5 warnings:
+  [1]:TooManyConstants: Smelly has at least 4 instance variables
```
troessner: Instance variables? C&p error?

troessner: Also given the configuration above Reek would **not** emit any warnings.

docs/Too-Many-Constants.md
```diff
@@ -0,0 +1,37 @@
...
+  CONST_2 = :dummy
+  CONST_3 = :dummy
+  CONST_4 = :dummy
+end
+```
+
+Reek would emit the following warning:
+
+```
+test.rb -- 5 warnings:
```
troessner: Lets redact this to say "1", otherwise it is confusing for the reader without seeing all the other  warnings.

features/smells/too_many_constants.feature
```diff
@@ -0,0 +1,19 @@
...
+        D = 4
+        E = 5
+        F = 6
+      end
+      """
+    When I run `reek my_class.rb`
+    Then it reports:
+    """
+    my_class.rb -- 1 warning:
+      [1]:TooManyConstants: MyClass has at least 6 constants [https://github.com/troessner/reek/blob/master/docs/Too-Many-Constants.md]
```
troessner: "At least"? That is confusing - why not "has X constants"?

lib/reek/report/code_climate/code_climate_configuration.yml
```diff
@@ -525,6 +525,37 @@ TooManyInstanceVariables:
...
+      CONST_3 = :dummy
+      CONST_4 = :dummy
+    end
+    ```
+
+    Reek would emit the following warning:
+
+    ```
+    test.rb -- 5 warnings:
+      [1]:TooManyConstants has at least 4 constants (TooManyConstants)
```
troessner: Same here, lets redact the number.

lib/reek/smells/too_many_constants.rb
```diff
@@ -0,0 +1,51 @@
...
+    # instance variables, methods, constants or lines of code.
+    #
+    # +TooManyConstants' reports classes having more than a
+    # configurable number of constants.
+    #
+    # See {file:docs/Too-Many-Constants.md} for details.
+    class TooManyConstants < SmellDetector
+      # The name of the config field that sets the maximum number
+      # of constants permitted in a class.
+      MAX_ALLOWED_CONSTS_KEY = 'max_constants'.freeze
```
troessner: We dont abbreviate this anywhere so lets rather spell it out....

lib/reek/smells/too_many_constants.rb
```diff
@@ -0,0 +1,51 @@
...
+      DEFAULT_MAX_CONSTS = 5
+
+      def self.contexts
+        [:class, :module]
+      end
```
troessner: Maybe worth investigating: i think "module" covery class as well.

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    it 'reports the number of consts' do
+      expect(src).to reek_of(:TooManyConstants, count: 6)
+    end
+
+    it 'reports the line where it occurs' do
+      expect(src).to reek_of(:TooManyConstants, lines: [1])
+    end
+
+    it 'reports the line where it occurs' do
```
mvz: -> reports the module.

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      expect(src).to reek_of(:TooManyConstants, context: 'Moduly::Klass')
```
mvz: Please also check no smell is reported on Moduly.

lib/reek/smells/too_many_constants.rb
waldyr: If a class is inside the module it will cover. However we might have classes without namespace like:
```
class Test
end
```
And in this case we should look into class context.

spec/reek/smells/too_many_constants_spec.rb
waldyr: As the class is also inside module both the class and the module is going to smell.

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+        class Dummy
+          A = B = C = D = E = 1
+        end
+      EOS
+
+      expect(src).not_to reek_of(:TooManyConstants)
+    end
+
+    it 'should count each constant only once' do
+      src = <<-EOS
```
mvz: If you put these three classes together in a module, my bet is the module will smell. I don't think this is what we want.

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    it 'reports the number of consts' do
+      expect(src).to reek_of(:TooManyConstants, count: 6)
+    end
+
+    it 'reports the line where it occurs' do
+      expect(src).to reek_of(:TooManyConstants, lines: [1])
+    end
+
+    it 'reports the context where it occurs' do
+      expect(src).to reek_of(:TooManyConstants, context: 'Moduly::Klass')
```
mvz: Only `Moduly::Klass` should smell, not `Moduly`.

spec/reek/smells/too_many_constants_spec.rb
waldyr: Got it.

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,174 @@
...
+          class Dummy
+          end
+        end
+      EOS
+
+      expect(src).to reek_of(:TooManyConstants, context: 'Dummiest')
+    end
+  end
+
+  context 'smell report' do
```
waldyr: @mvz I'd rather repeat it inside the `it` block as many times as needed instead of using this approach or `let`. The reason behind it is even though repeating it goes against DRY principle, in an `it` block I will have everything that is required to understand what is being tested and how.

For instance, let's suppose we start irresponsibly adding more `context`s and `it`s blocks inside this context. Even though I know there's a `src` it will become more and more difficult to find where it is defined. In every line addition the `src` variable will turn into a even more mysterious guest.

Even though the big block of text I'm fine with any possible conclusion. What do you think?

waldyr:
@mvz @troessner 

What about this one? Am I missing something here?

mvz:
Oooh ... this one totally went off the radar. I'll try to have a look at it this weekend.

lib/reek/smells/too_many_constants.rb
```diff
@@ -0,0 +1,57 @@
...
+    #
+    # +TooManyConstants' reports classes having more than a
+    # configurable number of constants.
+    #
+    # See {file:docs/Too-Many-Constants.md} for details.
+    class TooManyConstants < SmellDetector
+      # The name of the config field that sets the maximum number
+      # of constants permitted in a class.
+      MAX_ALLOWED_CONSTANTS_KEY = 'max_constants'.freeze
+      DEFAULT_MAX_CONSTS = 5
```
troessner: "Constants" is nowhere abbreviated so let's use "DEFAULT_MAX_CONSTANTS" here.

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+      src = <<-EOS
+        class Dummy
+          A = B = C = D = E = 1
+        end
+      EOS
+
+      expect(src).not_to reek_of(:TooManyConstants)
+    end
+
+    it 'should not report when increasing default' do
```
troessner: Nice spec!

spec/reek/smells/too_many_constants_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+          end
+        end
+      EOS
+
+      expect(src).to reek_of(:TooManyConstants, context: 'Dummiest')
+    end
+  end
+
+  context 'smell report' do
+    it 'reports the number of consts' do
```
troessner: "constants" please everywhere.

lib/reek/smells/too_many_constants.rb
```diff
@@ -0,0 +1,57 @@
...
+          EXCLUDE_KEY => [])
+      end
+
+      #
+      # Checks +klass+ for too many constants.
+      #
+      # @return [Array<SmellWarning>]
+      #
+      def inspect(ctx)
+        max_allowed_consts = value(MAX_ALLOWED_CONSTANTS_KEY, ctx)
```
troessner: And here :)

troessner:
Couple of small things and travis as usual :)
Then LGTM and put a stop to further detectors until we refactored our specs.

spec/reek/smells/too_many_constants_spec.rb
waldyr: Props to mutant. I really tried to make Travis green.

waldyr:
@troessner 

> Couple of small things and travis as usual :)

Addressed!

> Then LGTM and put a stop to further detectors until we refactored our specs.

Okay. Going to jump in the discussion.

troessner:
Ok, the remaining mutant failures are "ok" I guess. This PR is good to go.
Awesome job @waldyr !

troessner:
And I just released 4.2.0!

PR #958, opened 2016-06-24
Upgrade RuboCop

mvz:
This upgrades RuboCop to version 0.40 and fixes newly detected offences.

I have chosen to configure Style/MultilineMethodCallBraceLayout with the style same_line since that matches most of our code and looks better to me.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -52,6 +52,10 @@ Style/FrozenStringLiteralComment:
 Style/MultilineBlockChain:
   Enabled: false
 
+# Always put the closing brace on the last line
+Style/MultilineMethodCallBraceLayout:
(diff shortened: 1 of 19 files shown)


troessner:
Changes are looking good but seems like you missed an offense:

>> Offenses:
spec/reek/report/code_climate_formatter_spec.rb:37:121: C: Metrics/LineLength: Line is too long. [121/120]
      /\"body\":\"A _Utility Function_ is any instance method that has no dependency on the state of the instance.\\n\"/)

mvz:
That's what you get for rebasing and not checking the result :disappointed:.

mvz:
I fixed the remaining offense, cleaning up the CodeClimateFormatter specs in the process.

Issue #633, opened 2015-08-10
IVar Assumption


mvz:
Follow-up from #483:
- Don't access a superclass' instance variables directly in a subclass.
- Don't assume presence of ivars in a module.

Generalized: Do not assume ivars are set or present outside the current class/module definition.

It may be the case that the detector for #632 will automatically also detect this one:

``` ruby
class Foo
  def initialize
    @bar = :foo
  end
end

class Bar < Foo
  def foo?
    @bar == :foo
  end
end
```

troessner:
Cleaning up our issues: Linked this issue in #556 - closing this one.

mvz:
But I opened all of these based on #556, so we could track them separately.

Since we'll always have more ideas than time to implement them, why not keep them open and use tags to separate urgent from non-urgent stuff?

mvz:
Oh, no I didn't :-)

mvz:
Second point still stands, though.

troessner:
>> Since we'll always have more ideas than time to implement them, why not keep them open and use tags to separate urgent from non-urgent stuff?

My motivation was to cut down our open issues since we had almost 50 and for quite a few "nice to haves" it seemed like nobody was ever going to tackle those. I didn't anticipate that this was a controversial move, if I had, I would have started a quick "should we close old issues" poll.
-> Apologies for that.

>> Since we'll always have more ideas than time to implement them, why not keep them open and use tags to separate urgent from non-urgent stuff?

Yes, we can do that for sure! I'm re-opening this one and all others that are kind of "just an idea" issues.

troessner:
@mvz I did re-open all issues I could find that were used for idea tracking, feel free to re-open anything I had missed. We should have discussed this first, sorry for that.

mvz:
@troessner Apologies accepted, thanks! I'll have another look at old issues as well and add my comments.

waldyr:
This is all part of a greater smell often times referred as Mystery Guest.

mvz:
This has been implemented in #1017.

Issue #632, opened 2015-08-10
Uninitialized IVar


mvz:
Follow-up from #483:

**An instance variable must at least be set in the constructor, or be accessed through a method with lazy initialization.**

Good:
```ruby
class Foo
  def initialize
    @bar = :foo
  end

  def foo?
    @bar == :foo
  end
end
```

Bad:

```ruby
class Foo
  def go_foo!
    @bar = :foo
  end

  def foo?
    @bar == :foo
  end
end
```

Good:

```ruby
class Foo
  def foo?
    bar == :foo
  end

  def bar
    @bar ||= :foo
  end
end

troessner:
Cleaning up our issues: Linked this issue in #556 - closing this one.

troessner:
Seems like we had a misunderstanding, re-opening, see my last comment on #633

troessner:
Isn't this kind of expressed via #633 as well?
I'd then rather close this one. Right now it's confusing because we have 2 separate issues and this one feels like a subset #633 
I know there are some subtle differences, but I believe the user doesn't care for them that much as we apparently do :)

wdyt?

mvz:
Yes, I agree. See also https://github.com/troessner/reek/pull/959#issuecomment-232974279 :smile:.

PR #959, opened 2016-06-26
Add Instance Variable Assumption smell detector

waldyr:
#632 #633

This one might be difficult to have in practice.

This is a work in progress but review and comments are welcome :)


waldyr:
@troessner @mvz 

I think most Rails' actions will suffer from this smell 😆 

However help is needed here: Some cucumber scenarios are failing and there is no output other than the exit status. It's probably encountering smells but I can't see them. Is there a way to evidence those smells?

mvz:
@waldyr if you run just the failing scenario (`bundle exec cucumber features/the_scenario.feature:line`), Aruba will leave its set up in `tmp/aruba`. Then you can `cd` into that directory and run the scenario's Reek incantation there.

Aruba should really have a feature to print the output if an exit code step fails. Maybe it already does :smile:.

mvz:
> I think most Rails' actions will suffer from this smell :laughing:

Yes, we should add that to the Rails section of our documentation.

troessner:
>> if you run just the failing scenario (bundle exec cucumber features/the_scenario.feature:line), Aruba will leave its set up in tmp/aruba. Then you can cd into that directory and run the scenario's Reek incantation there.

Boyscout rule: Let's add this information to our [contributing guide](https://github.com/troessner/reek#developing-reek--contributing), would be great if you could add that in the course of this PR as well @waldyr  :)

waldyr:
Sure

waldyr:
Still have to do:
- [] Contributing guide update
- [] IVar Assumption on Rails section

But I'd like ideas on solving those utility functions.

lib/reek/smells/control_parameter.rb
```diff
@@ -42,6 +42,8 @@ module Smells
     # the source code.
     #
     # See {file:docs/Control-Parameter.md} for details.
+    #
+    # :reek:InstanceVariableAssumption
```
mvz: I think we should fix this smell instead.

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,184 @@
+require_relative '../../spec_helper'
+require_lib 'reek/smells/instance_variable_assumption'
+
+RSpec.describe Reek::Smells::InstanceVariableAssumption do
+  context 'report' do
+    context 'smell line' do
+      it 'should report the core class in the message' do
```
mvz: This message does not match what the spec tests.

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+            end
+          end
+        EOS
+
+        expect(src).to reek_of(:InstanceVariableAssumption, context: 'Dummy')
+      end
+    end
+
+    context 'smell message' do
+      it 'should report the class and ivars in the message' do
```
mvz: I'm not sure this one should smell, since the only use here is an actual assignment, so it doesn't assume the variable is set.

lib/reek/ast/sexp_extensions/methods.rb
```diff
@@ -60,6 +60,10 @@ def full_name(outer)
         def depends_on_instance?
           ReferenceCollector.new(self).num_refs_to_self > 0
         end
+
+        def instance_variable_assignments
+          find_nodes([:ivasgn, :ivar], [:or_asgn]).map(&:name)
```
mvz: I'd move this to the smell detector, since its *very* specific for that.

mvz:
> But I'd like ideas on solving those utility functions.

I'm not sure what you mean?

waldyr:
> I'm not sure what you mean?

I was just asking for review since this smell detector changed a lot of files :)

troessner:
I am confused @waldyr. Is this one up for final review now or are you still tweaking it?

waldyr:
I think I'm pretty much finished here. There is definitely things that I'll clean up (such as those suppressed smells) but I think this detector is surely up for a final review.

troessner:
Great, I'll review tonight :)

features/command_line_interface/basic_usage.feature
```diff
@@ -9,8 +9,9 @@ Feature: The Reek CLI maintains backwards compatibility
     Then the exit status indicates smells
     And it reports:
       """
-      demo.rb -- 8 warnings:
+      demo.rb -- 9 warnings:
         [3]:BooleanParameter: Dirty#awful has boolean parameter 'log' [https://github.com/troessner/reek/blob/master/docs/Boolean-Parameter.md]
+        [1]:InstanceVariableAssumption: Dirty has @screen instance variable assumptions [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
```
troessner: That

>> Dirty has @screen instance variable assumptions

doesn't read like English - how about 

>> Dirty assumes too much for instance variable @screen

?

lib/reek/ast/sexp_extensions/methods.rb
troessner: 👍

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,64 @@
...
+        ivar_assumptions = class_ivars(ctx) - initialized_ivars(ctx)
+
+        return [] if ivar_assumptions.empty?
+
+        build_smell_warning(ctx, ivar_assumptions.uniq)
+      end
+
+      private
+
+      def build_smell_warning(ctx, ivar_assumptions)
```
troessner: Nothing that's blocking this PR but for the future we really should refactor and abstract away the "building of smell warnings"  because right now this is more or less duplicated in almost every detector but in a slightly different way 😆

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,64 @@
...
+          context: ctx,
+          lines: [ctx.exp.line],
+          message: "has #{ivar_assumptions.join(', ')} instance variable assumptions",
+          parameters: { assumptions: ivar_assumptions })]
+      end
+
+      # Collection of ivars used by class
+      # :reek:UtilityFunction
+      def class_ivars(ctx)
+        ctx.node_instance_methods.map(&:instance_variable_assignments).flatten
```
troessner: Why not use the ".uniq" from above here right away? To ask differently: "Does it matter for the caller to know about duplicates?"
I assume not. In this case we should strip them out from the result right away.

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,64 @@
...
+        initialize_exp = initialize_exp(ctx)
+
+        return [] unless initialize_exp
+
+        initialize_exp.instance_variable_assignments
+      end
+
+      # Expression of initialize method
+      # :reek:UtilityFunction
+      def initialize_exp(ctx)
```
troessner: The method name is very confusing. How about renaming it to "initialize_method"?

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,64 @@
...
+      def initialized_ivars(ctx)
+        initialize_exp = initialize_exp(ctx)
+
+        return [] unless initialize_exp
+
+        initialize_exp.instance_variable_assignments
+      end
+
+      # Expression of initialize method
+      # :reek:UtilityFunction
```
troessner: This is lacking method documentation. We should get into the habit of always doing this.
A great example is [here](https://github.com/troessner/reek/blob/master/lib/reek/examiner.rb#L18)

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,64 @@
...
+          parameters: { assumptions: ivar_assumptions })]
+      end
+
+      # Collection of ivars used by class
+      # :reek:UtilityFunction
+      def class_ivars(ctx)
+        ctx.node_instance_methods.map(&:instance_variable_assignments).flatten
+      end
+
+      # Collection of ivars found on initialize
```
troessner: Same as below. Method documentation is missing. "ctx" doesn't have to be explained (otherwise you'd have to do it in all smell detectors and it would be redundant) but we need to be crystal clear about the return types.

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,64 @@
...
+      def initialized_ivars(ctx)
+        initialize_exp = initialize_exp(ctx)
+
+        return [] unless initialize_exp
```
troessner: The [NullObject](https://robots.thoughtbot.com/rails-refactoring-example-introduce-null-object) pattern would be far cleaner than having those checks everywhere but right now this is unfortunately idiomatic Reek code.
So this shouldn't be blocking this PR but for sure should be something that we eye in the future.

lib/reek/spec/should_reek_of.rb
```diff
@@ -18,6 +18,7 @@ def initialize(smell_type_or_class,
         @smell_details = smell_details
         @configuration = configuration
         @configuration.load_values(smell_type => { Smells::SmellConfiguration::ENABLED_KEY => true })
+        @matching_smell_types = nil
```
troessner: How is this change related to the PR?

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+
+    it 'should report any initialization outside initialize' do
+      src = <<-EOS
+        class Dummy
+          def initialize
+            @a = 1
+          end
+
+          def test
+            @a = 1
```
troessner: Why do you even have "@a" here? That's confusing. Can't we just remove it here and the "initialize" completely?

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+            @b = 1
+          end
+        end
+      EOS
+
+      expect(src).to reek_of(:InstanceVariableAssumption)
+    end
+
+    context 'on inner classes' do
```
troessner: I don't feel that this context adds a lot of value. I'd rather remove it completely. @mvz wdyt?

spec/samples/three_clean_files/clean_two.rb
```diff
@@ -1,5 +1,9 @@
 # clean class for testing purposes
 class Clean
+  def initialize(sub)
+    @sub = sub
+  end
+
```
troessner: Yep, that's what I meant with: https://github.com/troessner/reek/issues/978
We really need to address this before writing further detectors (this one is ok, talking about the ones after that).

lib/reek/spec/should_reek_of.rb
waldyr: Adding the ivar assignment fix this class' ivar assumption. Should I just suppress the warning?

lib/reek/smells/instance_variable_assumption.rb
mvz: That's what `smell_warning` was meant for. Perhaps we can indeed extract some common code.

lib/reek/smells/instance_variable_assumption.rb
mvz: For private methods, I disagree. I think it's better to document the structure of the solution on, e.g., `#inspect`, and be diligent in giving private methods good names.

lib/reek/spec/should_reek_of.rb
mvz: We should consider whether this smell is valid and adjust the detector accordingly if it is not (hint: I think it is not valid).

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,184 @@
+require_relative '../../spec_helper'
+require_lib 'reek/smells/instance_variable_assumption'
+
+RSpec.describe Reek::Smells::InstanceVariableAssumption do
```
mvz: All of these specs test assignment to ivars. I think the important part of this smell is detection of use, like:

```
  def foo
    @foo
  end
```

spec/reek/smells/instance_variable_assumption_spec.rb
mvz: I'm confused what's being tested here.

spec/reek/smells/instance_variable_assumption_spec.rb
mvz: Yes, I agree.

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+
+            class Dummiest
+            end
+          end
+        EOS
+
+        expect(src).to reek_of(:InstanceVariableAssumption)
+      end
+
+      it 'should report even if outer class initialize the variable' do
```
mvz: Well, this one may be a sensible test.

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+              end
+
+              def test
+                @a = 1
+              end
+            end
+          end
+        EOS
+
+        expect(src).to reek_of(:InstanceVariableAssumption)
```
mvz: You should add a matcher for the context here.

spec/samples/three_clean_files/clean_two.rb
mvz: Yes, let's make this one smaller instead of larger.

lib/reek/spec/should_reek_of.rb
troessner: >> Adding the ivar assignment fix this class' ivar assumption.

Now here's the problem: I'd consider initialising instance variables to nil as not fixing anything. 
There is no semantic difference between:

```Ruby
class Omg
  def initialize; @bar = nil; end
  def hola(arg); @bar = arg; end
end
```

and

```Ruby
class Omg
  def hola(arg); @bar = arg; end
end
```

Hence, it should still be a smell. 
@mvz @chastell wdyt?

I think initialising instance variables to nil is a strong hint that.....this shouldn't be an instance variable at all.
I'm fine with leaving the instance variable here to not block this pull request (after all, it was there before) but I think we need to discuss my question above since it is relevant for the detector.

lib/reek/smells/instance_variable_assumption.rb
troessner: 👍

lib/reek/smells/instance_variable_assumption.rb
troessner: Fine by me - i still think it's a nice add-on though :)

spec/reek/smells/instance_variable_assumption_spec.rb
troessner: That makes 2 of us 😆

spec/reek/smells/instance_variable_assumption_spec.rb
troessner: True, I agree.

lib/reek/spec/should_reek_of.rb
troessner: After thinking about it this:

>> I think initialising instance variables to nil is a strong hint that.....this shouldn't be an instance variable at all.

is a separate smell for which we'd need a separate detector. That is, if we all agree on that it is a smell 😆 
So please ignore my comment here, I'll create a separate issue for this.

troessner: Follow up discussion [here](https://github.com/troessner/reek/issues/979)

mvz: Yes, I think this case is not an instance of this smell, because no-one is **assuming** anything about the ivar.

mvz:
It seems this implementation basically detects **setting** an ivar in methods other than the initializer. This is not what I had in mind in #633 and #632: The point there is detecting usage of the value of an ivar without setting it in the initializer. In many cases a way around this smell is using a reader instead.

waldyr:
Really liking this detector. Too many questions raised.

Also I'm going to address all the comments today.

waldyr:
@mvz @troessner 

Code and I ready for another round of reviews 👍

waldyr:
@mvz @troessner 

Code and I ready for another round of reviews 👍

README.md
```diff
@@ -496,6 +496,8 @@ Just add this to your configuration file:
     max_allowed_nesting: 2
   UnusedPrivateMethod:
     enabled: false
+  InstanceVariableAssumption:
+    enabled: false
```
troessner: Yep, good call.

features/todo_list.feature
```diff
@@ -50,6 +54,9 @@ Feature:
     And the file ".todo.reek" should contain:
       """
       ---
+      InstanceVariableAssumption:
+        exclude:
+        - Smelly
```
troessner: Not blocking for this PR but we really need to fix our [test set up](https://github.com/troessner/reek/issues/978) before going on with the next smell detector. This is barely maintainable right now.

lib/reek/context/ghost_context.rb
```diff
@@ -8,6 +8,8 @@ module Context
     # context tree. This context will not be part of the resulting tree, but
     # will track context and visibility separately while building is in
     # progress.
+    #
+    # :reek:InstanceVariableAssumption
```
troessner: It's ok for me to disable this here and above.

lib/reek/smells/control_parameter.rb
troessner: I kind of agree but I believe it's more important to finish this PR and get this merged.
There are a couple of things we should refactor before new features anyway.

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,76 @@
...
+        end
+
+        def variables_from_initialize
+          initialize_exp = @method_expressions.detect do |method_exp|
+            method_exp.name == :initialize
+          end
+
+          return [] unless initialize_exp
+
+          initialize_exp.find_nodes([:ivasgn, :ivar], [:or_asgn]).map(&:name)
```
troessner: Let's extract this:

```Ruby
[:ivasgn, :ivar], [:or_asgn]
````

as an aptly named constant at the top of the class so we can remove the duplication with above and make this more intention revealing.

**Comment by troessner** [comment]
Just one [tiny thing](https://github.com/troessner/reek/pull/959/files#diff-4e583f4c70a9c84110c06f72a7dd1718R71) and this PR is ready to merge code wise from my side!

travis is failing on mutant which is ok to ignore this time AFAIK.

File: lib/reek/smells/instance_variable_assumption.rb (hunk 22 above)
Comment (mvz) [inline lib/reek/smells/instance_variable_assumption.rb]: Let's extract it to method then to avoid confusion.

File: lib/reek/context/ghost_context.rb (hunk 21 above)
Comment (mvz) [inline lib/reek/context/ghost_context.rb]: I don't agree that this class has this smell. We either need to change the smell name or change the implementation.

File: lib/reek/smells/instance_variable_assumption.rb (hunk 22 above)
Comment (mvz) [inline lib/reek/smells/instance_variable_assumption.rb]: My vote goes to an aptly-named method instead.

However, why would you remove the `||=` expressions when looking for ivars in the initializer?

File: spec/reek/smells/instance_variable_assumption_spec.rb [hunk 23]
```diff
@@ -0,0 +1,176 @@
...
+          end
+        EOS
+
+        expect(src).to reek_of(:InstanceVariableAssumption, message: message)
+      end
+    end
+  end
+
+  describe 'reporting' do
+    it 'should not report for using ivars from initialize' do
```
mvz: This spec description makes no sense anymore.

mvz:
> travis is failing on mutant which is ok to ignore this time AFAIK.

We should at least look at what mutant has to say about this detector. For the rest, I agree, since this PR touches many files without updating their logic.

mvz:
> We should at least look at what mutant has to say about this detector.

@waldyr let me know if you need any assistance with that.

mvz:
Mutant in fact only complains about the new detector class, so it should be possible to make it green.

lib/reek/smells/instance_variable_assumption.rb
waldyr: > My vote goes to an aptly-named method instead.

I agree.

> why would you remove the ||= expressions when looking for ivars in the initializer?

Well, I don't see advantages in using an `||=` in an object's initialization. For instance
```
class AClass
  def initialize
    @something ||= # something
  end  
end
```
This code makes me wonder why would `@something` have a previous value?

mvz: Makes sense. I think you can limit it to `:ivasgn` then. In fact, I think the mutant run on Travis is saying this as well.

waldyr: Good point

waldyr:
@mvz addressed almost all mutant complaints except one. Can you help me out?

screen shot 2016-07-13 at 11 01 35 pm

Should I move this into one of the contexts or is it good the way it is?

troessner:
>> @mvz addressed almost all mutant complaints except one. Can you help me out?

Yeah, those "eql" vs "==" complaints are something I think is ok to ignore. 
@mvz any objections?
If not, let's merge this one!

lib/reek/context/ghost_context.rb
mvz: Due to the changes in the detector, this class no longer has this smell so these lines can be removed.

lib/reek/smells/control_parameter.rb
mvz: Due to the changes in the detector, this class no longer has this smell so these lines can be removed.

lib/reek/context/code_context.rb
```diff
@@ -15,6 +15,7 @@ module Context
     #
     # :reek:TooManyMethods: { max_methods: 19 }
     # :reek:TooManyInstanceVariables: { max_instance_variables: 8 }
+    # :reek:InstanceVariableAssumption
```
mvz: Due to the changes in the detector, this class no longer has this smell so these lines can be removed.

lib/reek/spec/should_reek_of.rb
```diff
@@ -8,6 +8,7 @@ module Spec
     # An rspec matcher that matches when the +actual+ has the specified
     # code smell.
     #
+    # :reek:InstanceVariableAssumption
```
mvz: When I look at the code of this class, I think this class does not have this smell.

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+          def test
+            @a = 1
+          end
+
+          class Dummiest
+          end
+        end
+      EOS
+
+      expect(src).to reek_of(:InstanceVariableAssumption, context: 'Dummy')
```
mvz: IMHO, this example should *not* reek, because the class assumes nothing about `@a`: It is only assigned to.

mvz:
Nearly there :smile:. I think there's one small point left on what this smell should detect.

mvz:
I'm not sure why I made two separate tickets #632 and #633 :confused:.

spec/reek/smells/instance_variable_assumption_spec.rb
waldyr: So assigning ivars, with `=` and not `||=`, out of initialize's scope should be allowed?

lib/reek/spec/should_reek_of.rb
waldyr: Is there a specific reason for `@matching_smell_types = nil` not live in `initialize`?

mvz: Yes, every time the matcher is used it needs to re-initialize this value. This is because RSpec allows you to do this:

```ruby
match = reek_of(:UnicornPoo, color: 'pink')
expect(forest).to match
expect(city).not_to match
```

spec/reek/smells/instance_variable_assumption_spec.rb
mvz: Yes, I think so. I do like the lazily immutable style where you don't do this, but that's not what we're trying to enforce here.

lib/reek/spec/should_reek_of.rb
waldyr: Would not it be a `Temporary Field` smell?

mvz: Well, it is nearly always set, not just under odd circumstances. I suppose we could set it directly rather than using the memoization in line 52. Anyway, I don't like the code in this class much, but its design is a bit forced by RSpec's API.

troessner:
Lets get this one going!

I'm fine with mutant failing the "==" mutations.

So AFAIS this bad boy can be merged and released!

Any objections @mvz ?

waldyr:
I still have to address some comments.

waldyr:
Closed in favor of #1017

PR #967, opened 2016-07-06
Allow longer timeout for debugging purposes

mvz:
Stops specs from failing while debugging with pry.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -85,7 +85,7 @@ RSpec.configure do |config|
 
   # Avoid infinitely running tests. This is mainly useful when running mutant.
   config.around(:each) do |example|
-    Timeout.timeout(5, &example)
+    Timeout.timeout(60, &example)
   end
 end
 


troessner:
I ran into the same problem but this solution feels very hacky.
How about using a CLI flag:

```DEBUG=1 bundle exec rspec spec/.....```

and then doing:

```Ruby
Timeout.timeout(5, &example) unless ENV['DEBUG']
```

wdyt?

mvz:
Yeah, that is much nicer!

troessner:
Great! And let's also not forget to document this [here](https://github.com/troessner/reek#developing-reek--contributing) (we should mention the 5 seconds there anyways, regardless of this PR)

backus:
Could also put something like this at the top of your spec_helper:

```ruby
source =
  Pathname
    .glob('**/*.rb')
    .map(&:read)
    .join

DEBUG_MODE = (source =~ /binding\.pry/)
```

troessner:
@backus super smart solution but i think a too little smart for my taste for sth mundane like this ;)

troessner:
I think #969 is the way to go

backus:
@troessner alright. If you don't mind I'm curious what specifically you mean. Do you mean that the presence of a `binding.pry` disabling the timeout is too magic for something like this? I understand the judgement call I just want to make sure I understand your motivations.

For context, I have a test timeout setup (for mutant) in some of my projects and I've been meaning to add something like this `binding.pry` solution.

troessner:
@backus dont get me wrong, i really like your solution. It is just that I'd prefer something more explicit to avoid confusion (e.g. for new team members).
But thats really just a slight preference, if the other team members prefer your idea i am happy to roll with it as well.

backus:
@troessner okie doke. I was just curious. Wasn't trying to apply any pressure for you to take my solution. I don't even contribute to reek (yet). I just watch this repo because it is interesting and I think you guys are good developers :smile:

troessner:
>> Wasn't trying to apply any pressure for you to take my solution. 

Wasn't even remotely perceived like that. Was just a regular discussion - we absolutely appreciate those kind of suggestions :)

troessner:
Closing this one with #969 merged.

PR #969, opened 2016-07-07
Allow disabling spec Timeout for debugging

mvz:
Alternative for #967.

I'll still add the README bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,6 +461,23 @@ bundle exec rake console
 => Reek::Examiner
 ```
 
+You can also use Pry while running the tests by adding the following at the
+point where you want to start debugging:
+
+```ruby
+require 'pry'
+binding.pry
+```
+
+If you do this, you need to also run the specs with `DEBUG=1` set, e.g.:
+
+```
+DEBUG=1 bundle exec rspec spec/your/file_spec.rb:23
+```
+
+This is necessary because normally all specs run with a timeout of 5 seconds,
+which isn't much if you're busy using Pry.
+
 Have a look at our [Developer API](docs/API.md) for more inspiration.
 
 From then on you should check out:
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -84,8 +84,12 @@ RSpec.configure do |config|
   end
 
   # Avoid infinitely running tests. This is mainly useful when running mutant.
-  config.around(:each) do |example|
-    Timeout.timeout(5, &example)
+  # Set the DEBUG environment variable to something truthy like '1' to disable
+  # this and allow using pry without specs failing.
+  unless ENV['DEBUG']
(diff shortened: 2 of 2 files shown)


troessner:
Codr looks good!

mvz:
Documentation added.

README.md
```diff
@@ -461,6 +461,23 @@ bundle exec rake console
...
+binding.pry
+```
+
+If you do this, please run the specs with `DEBUG=1` set, e.g.:
+
+```
+DEBUG=1 bundle exec rspec spec/your/file_spec.rb:23
+```
+
+This is necessary because normaly all specs run with a timeout of 5 seconds,
```
troessner: s/normaly/normally

README.md
```diff
@@ -461,6 +461,23 @@ bundle exec rake console
...
 
+You can also use Pry while running the tests by adding the following at the
+point where you want to start debugging:
+
+```ruby
+require 'pry'
+binding.pry
+```
+
+If you do this, please run the specs with `DEBUG=1` set, e.g.:
```
troessner: Let's not use words like please in this context, otherwise we would have to use them all over the place ;)

README.md
```diff
@@ -461,6 +461,23 @@ bundle exec rake console
 => Reek::Examiner
 ```
 
+You can also use Pry while running the tests by adding the following at the
+point where you want to start debugging:
+
+```ruby
+require 'pry'
```
troessner: Shouldn't this be 

```Ruby
require 'pry-byebug'
```

?

troessner:
Bunch of small things and then LGTM!

README.md
mvz: No, just requiring pry will automatically load available plugins like pry-byebug.

troessner: Nice, didnt know that!

mvz:
Updated.

spec/spec_helper.rb
```diff
@@ -84,8 +84,11 @@ def sexp(type, *children)
   end
 
   # Avoid infinitely running tests. This is mainly useful when running mutant.
-  config.around(:each) do |example|
-    Timeout.timeout(5, &example)
+  # Set DEBUG to disable this and allow using pry without specs failing.
```
troessner: "Set DEBUG to..." - there is a part of the sentence missj g, isnt it?

mvz: Nope, you just need to set it. As in, 'to' here is *not* part of a "Set DEBUG to '1' to disable this". Maybe it's too ambiguous :smile:.

troessner: How about "to something truthy like 1 to disable it"?

mvz: Done.

Issue #953, opened 2016-06-16
Let's have Reek catch all uncatched exceptions at top level


troessner:
Originating from #942:

Let's have Reek catch all uncatched exceptions at top level so that we can 1) debug user issues like in #942  easier and 2.) Improve overall user experience. 

I'll come up with a first suggestion of how this could look like in this PR in the next couple of days and weeks.

mvz:
We can probably take a look at how RuboCop handles this, since they're pretty consistent about showing an error for the offending file and then continuing with the others.

mvz:
@troessner any sneak preview of how you want to handle this?

troessner:
So here's the idea:

### ReekException

* We have our own exception that is the **only** `RuntimeException` `Reek` should ever propagate upstream
* This will allow clients - be it CLI clients, RSpec matcher users or API users to react to `Reek` failures in a consistent, predictable way like this: "If you get a ReekException something is wrong with Reek and there's not much you can do about. So exit gracefully and report a bug"
* The `ReekException` always wraps the original exception to still allow for debugging

### Adapt all ports a client might use

Right now we have 3 integration possibilities:

* the binary that uses `CLI::Application`
* our RSpec matchers
* API users: Now that's more difficult since they could use literally everything we offer (marked "public" or not) but I think it's safe to say that the `Examiner` is the main touchpoint

Now the question becomes "How can update all of those integrations so that whatever happens the user gets back a ReekException?".

**I see 2 options:**

1.) We update every single integration option from above so that it:

* catches all exceptions < RuntimeException
* wraps them in a `ReekException`

2.) We just update the `Examiner` to do so.

(1) would be kind of bulletproof but also seems a little over the top and will also imply some slight code duplication.
(2) would not be 100% bulletproof in the sense that there might be some edge cases where a `RuntimeException` slips through that gets not wrapped into a `ReekException`. But it would be a minimal change compared to (1).

I think we should go with (2)

### SmellDetector

Problem is that `ReekException` from above won't fix it because it'll still miss the detailed information about the parsed source code that actually triggered the error.
Hence I suggest to make `SmellDetector` catch all RuntimeErrors as well and enrich them with information like "which file, which line" and so on.
It then re-raises the exception where it will be caught by the `Examiner` properly and wrapped into a `ReekException`.

WDYT?

troessner:
First spec draft up in #971

mvz:
I think we're trying to achieve some conflicting purposes:

[...]

PR #971, opened 2016-07-08
Catch, enrich and re-raise all exceptions as ReekException.

troessner:
Spec draft for #953

### DIFF
--- a/lib/reek/reek_exception.rb
+++ b/lib/reek/reek_exception.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+module Reek
+  #
+  # @public
+  #
+  # ReekException is the only exception that Reek should give back
+  # to the user in case something goes wrong.
+  class ReekException < RuntimeError
+  end
+end
--- a/spec/reek/smells/smell_detector_spec.rb
+++ b/spec/reek/smells/smell_detector_spec.rb
@@ -2,6 +2,18 @@ require_relative '../../spec_helper'
 require_lib 'reek/smells/smell_detector'
 
 RSpec.describe Reek::Smells::SmellDetector do
+  describe 'exception handling' do
+    describe 'catches all exceptions, enriches them and re-raises them' do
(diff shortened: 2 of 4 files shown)


spec/reek/smells/smell_detector_spec.rb
```diff
@@ -2,6 +2,18 @@
...
+  describe 'exception handling' do
+    describe 'catches all exceptions, enriches them and re-raises them' do
+      it 'adds source file line number to the original exception' do
+        skip
+      end
+
+      it 'adds source file name to the original exception' do
+        skip
+      end
+    end
```
troessner: Just a bunch of first thoughts here. We can also directly print out the offending code,

lib/reek/reek_exception.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+module Reek
+  #
+  # @public
+  #
+  # ReekException is the only exception that Reek should give back
+  # to the user in case something goes wrong.
+  class ReekException < RuntimeError
```
backus: Any reason you choose repeating `Reek` (`Reek::ReekException`) instead of something more concise like `Reek::Error`?

troessner: I have seen this pattern in a couple of libraries (can't remember which though right now unfortunately) and I kind of like it since I find it very intention revealing. `Error` in the context of `Reek` doesn't tell me much as an outsider since it's pretty vague. `ReekException` (or maybe better: `ReekError`?) kind of signals that this is the exception `Reek` will propagate to the caller.

backus: Interesting! Thanks for explaining.

troessner:
Closing this one since we discussed  in #953 that the changes proposed here are too ambitious and that we'll start smaller by just handling crashing smell detectors in a first step. 
I'll update this PR accordingly as soon as I get around to it.

troessner:
Closing now in favor of #996

PR #976, opened 2016-07-09
Remove guard exceptions from Reek::Report

mvz:
- `Hash#fetch` already raises if the key is not found, which is enough for a non-public API.
- These exceptions were untested

To allow mutant to check this properly, I had to remove `module_function`. See https://github.com/mbj/mutant/issues/459.

### DIFF
--- a/lib/reek/report.rb
+++ b/lib/reek/report.rb
@@ -6,8 +6,6 @@ require_relative 'report/heading_formatter'
 module Reek
   # Reek reporting functionality.
   module Report
-    module_function
-
     REPORT_CLASSES = {
       yaml: YAMLReport,
       json: JSONReport,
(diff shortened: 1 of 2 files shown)


lib/reek/report.rb
```diff
@@ -39,28 +37,38 @@ module Report
...
-    def warning_formatter_class(warning_format)
-      WARNING_FORMATTER_CLASSES.fetch(warning_format) do
-        raise "Unknown warning format: #{warning_format}"
-      end
+    # Map warning format symbol to a report class.
+    #
+    # @param [Symbol] warning_format The format to map
+    #
+    # @return The mapped warning class
+    #
```
troessner: Nice, appreciate the comments a lot!

troessner:
Nice job! Merged!

Issue #980, opened 2016-07-16
Add rubocop style progress reporting to reek


jhubert:
I'm a big fan of the way that RuboCop reports on the progress of an inspection:

![image](https://cloud.githubusercontent.com/assets/3384/16897438/8a25bb14-4b66-11e6-950f-f6e0b156e99b.png)

It's especially great when using RuboCop with something like Guard that runs every time a file is changed.

That style of progress display would be really nice in Reek.

troessner:
>> I'm a big fan of the way that RuboCop reports on the progress of an inspection:

Great initiative and I fully agree :)

I did check out both of your pull requests but I think it makes sense if we discuss the specifics of this feature here before going into code since the same questions pop up and this will be pretty hard to follow and discuss on 2 different pull requests.

**The following questions and discussion items come to my mind:**

* Should the "progress bar" be a property of the "text format" or it's own format? You modelled it [here](https://github.com/troessner/reek/pull/983/files#diff-ccf18235057177600222fe6eedf2f132R111) as its own format (albeit directly inheriting from TextReport) which I personally find a little confusing. I'd rather make it just a property of TextReport which means it would be a separate flag like [in your other pull request](https://github.com/troessner/reek/pull/982/files#diff-b5b9b45d56bae1cd97fb9ee6745139d4R49).
* Looking at [making it part of the ReportCommand](https://github.com/troessner/reek/pull/982/files#diff-9ce866437f22b59ed4a1995bf2c65aadR23) I believe the right way to go is indeed to put this behaviour into our reports
* Given that the "progress bar" is only available for text reports, wouldn't it make sense to enable it by default?

cc @mvz @waldyr

mvz:
Looking at it from a functional level, I would have it appear as an option for the text report, unless we have it appear on STDERR. In the latter case, we can have it everywhere and it won't interfere with the report output.

I see that RuboCop has it as a separate report format, but that is indeed a bit odd.

If we go for the option of putting the behavior in TextReport, we can always model it as a decorator or put it in the superclass if we want to make it available for more than one format.

It makes sense to enable the progress reporter by default eventually, but I think that should not be done in the 4.x series.

TL;DR: My vote is to make it an non-default option on TextReport and implement it there.

troessner:
>> It makes sense to enable the progress reporter by default eventually, but I think that should not be done in the 4.x series.

Why? I really don't think we need to be so anal about semver for Reek. This feature is a clear improvement and I think we should ship as fast as possible.

[...]

PR #982, opened 2016-07-16
Optionally show the examiner progress

jhubert:
This adds an option to the cli that shows a `.` or a `S` for each file as they are examined. There may be a more elegant way to accomplish this, or to have it as a reporter, but this accomplished my needs and I haven't spent too much time understanding all of the complexities of the reek reporters.

image

Per #980

### DIFF
--- a/features/command_line_interface/options.feature
+++ b/features/command_line_interface/options.feature
@@ -46,6 +46,9 @@ Feature: Reek can be controlled using command-line options
       Generate a todo list:
           -t, --todo                       Generate a todo list
 
+      Show progress:
+          -p, --progress                   Show progress of each file
+
       Report format:
           -f, --format FORMAT              Report smells in the given format:
                                              html
--- a/features/command_line_interface/show_progress.feature
+++ b/features/command_line_interface/show_progress.feature
@@ -1,0 +1,30 @@
+Feature: Show progress
+  In order to see the progress of the examiners
+  As a developer
+  I want to be able to selectively activate progress reporting
+
+  Scenario: --progress activates progress output on clean files
+    Given a directory called 'clean_files' containing some clean files
+    When I run reek --progress clean_files
+    Then it succeeds
+    And it reports:
+      """
+      Inspecting 3 file(s):
+      ...
+
+      0 total warnings
+      """
+  Scenario: --progress activates progress output on smelly files
+    Given a smelly file called 'smelly.rb'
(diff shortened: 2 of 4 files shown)


jhubert:
A second take at this that uses a report format has been done as #983

jhubert:
I could use some help with figuring out how to deal with the test suite. I've never dealt with mutations. The cucumber feature passes.

features/command_line_interface/show_progress.feature
```diff
@@ -0,0 +1,30 @@
...
+    And it reports:
+      """
+      Inspecting 1 file(s):
+      S
+
+      smelly.rb -- 3 warnings:
+        [4, 5]:DuplicateMethodCall: Smelly#m calls @foo.bar 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
+        [4, 5]:DuplicateMethodCall: Smelly#m calls puts @foo.bar 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
+        [3]:UncommunicativeMethodName: Smelly#m has the name 'm' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Method-Name.md]
+      """
```
troessner: How does it look like when it is mixed? E.g. 1 smelly file and 3 non-smelly ones?
It would be great if we could add another scenario.

features/command_line_interface/options.feature
```diff
@@ -46,6 +46,9 @@ Feature: Reek can be controlled using command-line options
       Generate a todo list:
           -t, --todo                       Generate a todo list
 
+      Show progress:
+          -p, --progress                   Show progress of each file
```
troessner: I kind of wonder if this shouldn't be the default. @mvz @waldyr wdyt?

mvz: Eventually yes, but I would prefer not to make it so for the 4.x series.

waldyr: I also think this should be the default.

features/command_line_interface/show_progress.feature
mvz: Or just extend this one to cover several files.

troessner:
I think we all agree that #983 is the way to go and not change the Examiner.
Closing this one and lets continue in #983

jhubert:
:+1:

PR #983, opened 2016-07-17
Add a progress report format on top of the text format

jhubert:
This is an alternate take on #982 that uses a format built on top of the text format to display the progress as it runs. The downside is that it only works for the specific format instead of being independent of the format choice, but it seems to fit more closely with how the reporting has been designed.

Per #980

### DIFF
--- a/features/command_line_interface/options.feature
+++ b/features/command_line_interface/options.feature
@@ -54,12 +54,12 @@ Feature: Reek can be controlled using command-line options
                                              json
                                              xml
                                              code_climate
-                                             progress
 
       Text format options:
               --[no-]color                 Use colors for the output (default: true)
           -V, --[no-]empty-headings        Show headings for smell-free source files (default: false)
           -U, --[no-]wiki-links            Show link to related wiki page for each smell (default: true)
+          -P, --[no-]progress              Show progress of each file (default: true)
           -n, --[no-]line-numbers          Show line numbers in the output (default: true)
           -s, --single-line                Show location in editor-compatible single-line-per-smell format
               --sort-by SORTING            Sort reported files by the given criterium:
--- a/features/command_line_interface/options.feature
+++ b/features/command_line_interface/options.feature
@@ -54,6 +54,7 @@ Feature: Reek can be controlled using command-line options
                                              json
                                              xml
                                              code_climate
+                                             progress
 
       Text format options:
               --[no-]color                 Use colors for the output (default: true)
--- a/features/command_line_interface/show_progress.feature
+++ b/features/command_line_interface/show_progress.feature
@@ -3,28 +3,17 @@ Feature: Show progress
   As a developer
   I want to be able to selectively activate progress reporting
 
-  Scenario: --progress activates progress output on clean files
-    Given a directory called 'clean_files' containing some clean files
-    When I run reek -f progress clean_files
-    Then it succeeds
-    And it reports:
-      """
-      Inspecting 3 file(s):
-      ...
-
-      0 total warnings
-      """
-  Scenario: --progress activates progress output on smelly files
-    Given a smelly file called 'smelly.rb'
-    When I run reek -f progress smelly.rb
+  Scenario: --progress activates progress output on mixed files
+    Given a directory called 'mixed_files' containing some clean and smelly files
+    When I run reek -f progress mixed_files
     Then the exit status indicates smells
     And it reports:
       """
-      Inspecting 1 file(s):
-      S
+      Inspecting 2 file(s):
+      .S
 
-      smelly.rb -- 3 warnings:
-        [4, 5]:DuplicateMethodCall: Smelly#m calls @foo.bar 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
-        [4, 5]:DuplicateMethodCall: Smelly#m calls puts @foo.bar 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
-        [3]:UncommunicativeMethodName: Smelly#m has the name 'm' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Method-Name.md]
+      mixed_files/dirty.rb -- 2 warnings:
+        [1]:IrresponsibleModule: Dirty has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]
+        [2]:UncommunicativeMethodName: Dirty#a has the name 'a' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Method-Name.md]
+      2 total warnings
       """
--- a/features/command_line_interface/show_progress.feature
+++ b/features/command_line_interface/show_progress.feature
@@ -1,0 +1,30 @@
+Feature: Show progress
+  In order to see the progress of the examiners
+  As a developer
+  I want to be able to selectively activate progress reporting
+
+  Scenario: --progress activates progress output on clean files
+    Given a directory called 'clean_files' containing some clean files
+    When I run reek -f progress clean_files
+    Then it succeeds
+    And it reports:
+      """
+      Inspecting 3 file(s):
+      ...
+
+      0 total warnings
+      """
+  Scenario: --progress activates progress output on smelly files
+    Given a smelly file called 'smelly.rb'
+    When I run reek -f progress smelly.rb
+    Then the exit status indicates smells
+    And it reports:
+      """
+      Inspecting 1 file(s):
+      S
+
+      smelly.rb -- 3 warnings:
+        [4, 5]:DuplicateMethodCall: Smelly#m calls @foo.bar 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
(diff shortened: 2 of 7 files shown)


troessner:
I'll hold of with the review of this one for now. Let's discuss this first in #980

features/command_line_interface/options.feature
```diff
@@ -54,6 +54,7 @@ Feature: Reek can be controlled using command-line options
                                              json
                                              xml
                                              code_climate
+                                             progress
```
troessner: As discussed in #980 let's rather make this an option for the TextFormat.

features/command_line_interface/show_progress.feature
```diff
@@ -0,0 +1,30 @@
...
+    When I run reek -f progress clean_files
+    Then it succeeds
+    And it reports:
+      """
+      Inspecting 3 file(s):
+      ...
+
+      0 total warnings
+      """
+  Scenario: --progress activates progress output on smelly files
```
troessner: Let's add a scenario where smelly and non-smelly files are mixed or update this to reflect this case.

lib/reek/report/report.rb
```diff
@@ -121,6 +123,41 @@ def total_smell_count_message
     end
 
     #
+    # Displays the status of each file as it is examined
+    # Displays the text report when complete
+    #
+    # @public
+    class ProgressReport < TextReport
```
troessner: Do you feel that modelling the option (see above) as separate Report class?
If yes, we can go with this way as well. I just think that it's important that we show it as an option to the user to not confuse him.

jhubert: There is a lot of duplication between this and the TextReport class, so it felt cleaner to subclass instead of creating an entirely independent class.

It's not clear to me how options would be passed down to the Report class, so if I was to re-implement this I could rename the class to be `TextReportWithProgress` and still subclass the `TextReport`. Then I would use the options to decide if we call `TextReport` or `TextReportWithProgress`.

Do you think the duplication would be worth it?

jhubert: Also, please keep in mind that I've only spent about an hour looking at this code base, so if there is a better way to implement it I'm listening. 😄

features/command_line_interface/show_progress.feature
jhubert: Addressed in 441ec88

lib/reek/report/report.rb
mvz: Options are passed down as named arguments to the initializer. See for instance `sort_by_issue_count`.

jhubert:
I've just revised this PR to make progress the default behavior of the TextReport class. Let me know if that's what you had in mind. I think this is pretty straight forward.

I'm going to need some help understanding the mutation tests before travis passes.

lib/reek/cli/options.rb
```diff
@@ -136,6 +139,9 @@ def set_up_verbosity_options
                   'Show link to related wiki page for each smell (default: true)') do |show_links|
           self.show_links = show_links
         end
+        parser.on('-P', '--[no-]progress', 'Show progress of each file (default: true)') do |show_progress|
```
jhubert: I debated whether or not this could be in it's own method (like colored output), but it felt well placed within the verbosity options.

lib/reek/report/report.rb
```diff
@@ -63,14 +67,19 @@ def smells
         examiners.map(&:smells).flatten
       end
 
+      def progress(msg)
```
jhubert: I put this here as a nod towards the idea that other Reporters could easily add progress reporting. Happy to move it if you disagree.

lib/reek/report/report.rb
jhubert: @mvz Thanks! I just pushed up a new version that follows that pattern.

troessner:
Ok, I KNOW that this is very erratic and I'm sorry for that but comparing your [first commit](https://github.com/troessner/reek/pull/983/commits/24ffef02b181c95df81125db337b3faf8d4b1eef) that uses ProgressReport as a separate format with the latter solution (yes, the one that I pushed for 😆 ) I actually think using ProgressReport looks way cleaner and elegant.
Additionally I have come to the conclusion that it is not counter-intuitive to label this as "format". We define what a format is in Reeks context and ProgressReport resonates well with me.

So I'd suggest to:

- Kill commit [nr 3](https://github.com/troessner/reek/pull/983/commits/c531564db084ef000b01eb8e7710bb358943a723) and [nr 4](https://github.com/troessner/reek/pull/983/commits/6861d7d37708617fdb9988e13726e37b63365f07)
- Give this one a final review
- merge it!

@mvz wdyt?
@jhubert sorry again for being so jumpy!

mvz:
@troessner I need to ponder this for a bit.

mvz:
Functionally, I think it would be odd to provide progress as a separate report format. It would also block the option of providing it for the other formats.

I think we should focus on the end-user-visible result here. We can always refactor the code later.

So, my full vote here is:
* Progress should be an option, not a format: `-p` seems nice.
* Progress should ideally go to STDERR to allow pipes to do the right thing.
* Progress should first be only available for TextReport, and be on by default there.

There is an alternative for the STDERR point: If progress is only on for TTYs (like color is), it can go to STDOUT. I may like that option even better now that I think of it.

Oh, finally: I would like to play with this option a bit before we release, so let's leave a little time after merging for that.

troessner:
>> It would also block the option of providing it for the other formats.

But right now this wouldn't be a problem for **any** of our existing formats and it's not like us adding new formats every day.

>> Progress should ideally go to STDERR to allow pipes to do the right thing.

But the progress bar has nothing to do with **errors** so I can't see why it should go to STDERR.

It seems @mvz and me are disagreeing here big time. But at the same time this should not block this feature.
So I can live with it being an option for the sake of getting this rascal on the road ;)

@jhubert can you please squash this to one commit so we can merge this bad boy?

mvz:
> But right now this wouldn't be a problem for any of our existing formats and it's not like us adding new formats every day.

Changing our minds about this later would mean another major version bump (a format going away), so I'd rather have the discussion right now :smile:.

I'm still not sure why you think this should be a separate report format from the standpoint of the user?

> But the progress bar has nothing to do with errors so I can't see why it should go to STDERR.

I agree. Unix should have a STDLOG or STDNOTIMPORTANT output stream. I'm liking the idea of handling this like color more and more. That can be a next step though. Let's merge this first!

troessner:
>> I'm still not sure why you think this should be a separate report format from the standpoint of the user?

Well, it IS a separate format in the way that information is now formatted differently 😆
Would you only count different mime-type subtypes as format?

mvz:
> Well, it IS a separate format in the way that information is now formatted differently :laughing:

LOL. You mean like reporting with and without colour, line numbers and wiki links :stuck_out_tongue:?

> Would you only count different mime-type subtypes as format?

I wouldn't go that far :smile:.

troessner:
Maybe we should publish a Reek specific dictionary where we extensively discuss the word the exact meaning of "format" 😆

troessner:
@jhubert we have another api breaking PR [here](https://github.com/troessner/reek/pull/997) that is close to finish. Would be great if we could get this one here going. AFAIS you only need to fix the specs and squash to one commit, then I'm happy to merge this one!

jhubert:
@troessner Apologies for the slowness. Been a crazy week at work. Taking a look at the PR tonight to fix those specs.

mvz:
@jhubert did you manage to get any further on those specs?

jhubert:
@mvz @troessner I finally got a break and sat down to solve this. I realized that everything was prettry straight forward, with the exception of the stdin specs. stdin defaults to silent so that you can pipe the output to various places, so I believe we should *not* show progress by default if stdin is being used.

In other words, [this feature spec](https://github.com/troessner/reek/blob/master/features/command_line_interface/stdin.feature#L6) should stay as is. But the [verbose feature spec](https://github.com/troessner/reek/blob/master/features/command_line_interface/stdin.feature#L16) should include progress.

I'm not entirely sure how to handle this in an elegant way.. any suggestions?

troessner:
Fully agree, lets ignore stdin completely for this feature.
Lets make this feature only work for the regular case., i.e. `reek my_files/`.

jhubert:
@troessner @mvz Well, after dabbling with this one for a while I decided that I wasn't a fan of the implementation so I took a crack at #1096 with all of the specs passing. 

I prefer that implementation, but if you don't like it let me know and I'll get to work on using similar logic to get this one passing.

Sorry again for the delay on this. I hate to slow down progress.

troessner:
Closing this one with #1096 up ;)

mvz:
I have looked at the plugins that use Reek's output directly, and they seem to completely ignore any output that is not a smellwarning. Therefore, I'd like to not mark this as a breaking change and release it with Reek 4.5 (since it's a new feature :tada:).

jhubert:
🎉

troessner:
>> I have looked at the plugins that use Reek's output directly, and they seem to completely ignore any output that is not a smellwarning. Therefore, I'd like to not mark this as a breaking change and release it with Reek 4.5 (since it's a new feature

Sounds good to me, thanks for checking. Are you going to cherry-pick the single differing commit in this branch into a new PR against master? Happy to merge then :)

mvz:
Ok, I'll set up a new pull request.

mvz:
@troessner see #1121.

PR #988, opened 2016-07-19
Fix TooManyConstants for outer module when inner module suppress

### LINKED ISSUE #987: TooManyConstants Affects Parent Modules Even When Ignored
Hi, I just upgraded to `4.2.0` from an older version and noticed the new `TooManyConstants` smell which, to me, has some unexpected behavior.

If I have a scenario like this

```ruby
module Foo
  class Bar
    A = 1
    B = 2
    C = 3
    D = 4
    E = 4
    F = 4
  end
end
```

and run `reek`, i get:
>   [2]:TooManyConstants: Foo::Bar has 6 constants [https://github.com/troessner/reek/blob/master/docs/Too-Many-Constants.md]

which is all well and good. If I change `Bar` to a module:

```ruby
module Foo
  module Bar
    A = 1
    B = 2
    C = 3
    D = 4
    E = 4
    F = 4
  end
end
```

I get
>   [1]:TooManyConstants: Foo has 6 constants [https://github.com/troessner/reek/blob/master/docs/Too-Many-Constants.md]
  [2]:TooManyConstants: Foo::Bar has 6 constants [https://github.com/troessner/reek/blob/master/docs/Too-Many-Constants.md]

Even when I add an ignore on `Bar`

```ruby
module Foo
  # ignore :reek:TooManyConstants:
  module Bar
    A = 1
    B = 2
    C = 3
    D = 4
    E = 4
    F = 4
  end
end
```

I still get
>   [1]:TooManyConstants: Foo has 6 constants [https://github.com/troessner/reek/blob/master/docs/Too-Many-Constants.md]

This extends for as many modules as you nest. Adding the ignore above `Foo` instead of above `Bar` suppresses both warnings in this case.

Is this a bug? Or am I missing some module-specific purpose here?

waldyr:
Fixes #987

### DIFF
--- a/lib/reek/smells/too_many_constants.rb
+++ b/lib/reek/smells/too_many_constants.rb
@@ -17,6 +17,7 @@ module Reek
       # of constants permitted in a class.
       MAX_ALLOWED_CONSTANTS_KEY = 'max_constants'.freeze
       DEFAULT_MAX_CONSTANTS = 5
+      IGNORED_NODES = [:module, :class].freeze
 
       def self.contexts
         [:class, :module]
@@ -36,7 +37,7 @@ module Reek
       def inspect(ctx)
         max_allowed_constants = value(MAX_ALLOWED_CONSTANTS_KEY, ctx)
 
-        count = ctx.each_node(:casgn, [:class]).delete_if(&:defines_module?).length
+        count = ctx.each_node(:casgn, IGNORED_NODES).delete_if(&:defines_module?).length
 
         return [] if count <= max_allowed_constants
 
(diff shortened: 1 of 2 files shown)


waldyr:
@troessner

lib/reek/smells/too_many_constants.rb
```diff
@@ -52,6 +52,10 @@ def build_smell_warning(ctx, count)
           message: "has #{count} constants",
           parameters: { count: count })]
       end
+
+      def ignored_nodes
```
troessner: This is not supposed to change so lets rather make this a constant.

troessner:
One small thing the lgtm ;)

waldyr:
Done.

troessner:
Awesome speed, merged! Would be great if you could release a new bugfix version right away ;)

troessner:
Btw you can now issue PRs directly to this repo, no need for your fork anymore ;)

PR #990, opened 2016-07-20
Make mutant task more robust

mvz:
Specifying 'master^' makes sure that in most day-to-day cases, the
mutant task will find some commits to check. The is preferable to
checking for the current branch, which seems to fail on Travis.

An alternative is to use HEAD~n, for some value of n. However, this
could hide problems while a PR is being developed, while potentially
prolonging build failures on master after merging a branch that mutant
does not approve of.

### DIFF
--- a/tasks/mutant.rake
+++ b/tasks/mutant.rake
@@ -1,18 +1,13 @@
 desc 'Runs mutant'
 task :mutant do
   command = <<-EOS
-    branch_name=$(git symbolic-ref HEAD 2>/dev/null | cut -d"/" -f 3)
-    if [ "$branch_name" != "master" ]; then
-      RUBY_THREAD_VM_STACK_SIZE=64000\
-      bundle exec mutant\
-        --include lib\
-        --require reek\
-        --use rspec\
-        --since master\
-        --jobs 4 'Reek*'
-    else
-      echo "!!! Not running mutant on master branch !!!"
-    fi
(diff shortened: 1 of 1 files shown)


troessner:
>> The is preferable to
checking for the current branch, which seems to fail on Travis.

Yeah, that's disappointing. Our current check seems to fail because travis probably creates tmp branches for everything, even master, and then the names of the branches change and thus our check fails (don't know that for sure so that is just conjecture though).

tasks/mutant.rake
```diff
@@ -1,18 +1,13 @@
...
-        --jobs 4 'Reek*'
-    else
-      echo "!!! Not running mutant on master branch !!!"
-    fi
+    RUBY_THREAD_VM_STACK_SIZE=64000\
+    bundle exec mutant\
+      --include lib\
+      --require reek\
+      --use rspec\
+      --since master^\
```
troessner: But couldn't this result in bogus, false positive mutant failures if we run mutant not against the most current commit on master?

mvz: I think there will be little difference in practice compared to just using master. For branches based off latest master, there is one extra diff being taken into account (and that should be mutant-clean, given it's on master).

The only time this will be a problem is when we merge a branch mutant doesn't like.

troessner: Hok, fair enough.

PR #991, opened 2016-07-24
Revamp our badge setup.

### LINKED ISSUE #985: Evaluate badges from the hamster gem
I think that [the badges from the hamster gem](https://github.com/hamstergem/hamster) look pretty awesome. 
Let's evaluate if some of those make sense for us.

troessner:
Fixes #985

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,12 +34,16 @@
 
 ## Overview
 
-[![Build Status](https://secure.travis-ci.org/troessner/reek.svg?branch=master)](https://travis-ci.org/troessner/reek?branch=master)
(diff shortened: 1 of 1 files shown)


troessner:
Ready for review. 

I added badges for:

* total downloads
* downloads for current version
* git tag
* licence

and I also flattened out our badges which is easier on the eyes.

mvz:
Nice! Is there a better dependencies badge? This one still looks blurry.

troessner:
>> Nice! Is there a better dependencies badge? This one still looks blurry.

Good catch!
Changed that as well, looks way better now ;)

Issue #978, opened 2016-07-11
Refactor our specs big time


troessner:
Completely unrelated to @waldyr 's [PR](https://github.com/troessner/reek/pull/959/files) but honest to god, our spec set up is a huge pain in the ass sometimes.
When one isolated change like a new smell detector causes a huge diff already then this means that we have a lot, a lot of "know it all, test it all" specs.

I believe we should start by making our classical "smelly.rb" example that leaks throughout our spec base wayyy smaller so that it actually only has one single offense. 

Then we should remove our shared examples or clean them up significantly. 
Then we should remove all kind of duplicated specs.

Let's work on that in the future. Our spec set up is the thorn in our paw.

mvz:
Yes. The best strategy here is probably to avoid adding complexity to the specs. So any PR should either leave the overall spec structure exactly as is, or it should remove complexity.

One think we cannot avoid is new detectors detecting new smells in our own code base. This does however provide a first test of validity for those new detectors: Do they work in practice on a relatively neat code base?

:+1: on keeping the clean and smelly example code as simple as possible.

Let's keep this issue open and we can chip away at it a bit at a time.

waldyr:
> I believe we should start by making our classical "smelly.rb" example that leaks throughout our spec base wayyy smaller so that it actually only has one single offense.

An empty class reeks of `Irresponsible Module`.

Just enumerating things to do for this issue so I can tackle them later:

- [ ] Make `smelly.rb` only reeks of `Irresponsible Module`.
- [ ] Remove Shared Examples.
- [ ] Seek and destroy all duplicated specs.

Am I missing something?

mvz:
> Make smelly.rb only reeks of Irresponsible Module.

I like this one.

troessner:
>> Make smelly.rb only reeks of Irresponsible Module.
I like this one.

Not a big fan. I don't think this example is very accessible for newbies because in real life you almost never have empty classes.

How about something that is fully functional and still kind of obviously smelly like this one:

```Ruby
# Dummy class
class Dummy
  # This will reek of UncommunicativeMethodName
  def x
  end
end
```

WDYT?

mvz:
> WDYT?

Yes, that one is better.

waldyr:
Going to work on it later.

waldyr:
Are we going to drop all samples files except `smelly.rb`?

troessner:
>> Are we going to drop all samples files except smelly.rb?

[...]

PR #992, opened 2016-07-27
Refactor Specs

waldyr:
WIP for #978 

Beginning refactor our specs so one Smell Detector does not break most tests.


waldyr:
@troessner @mvz @chastell 

Is this how the samples folder should look like?

Can I start fixing those tests?

mvz:
Since cleaning up everything will be a long process, I would highly recommend doing this one sample file at a time. Baby steps.

troessner:
>> Since cleaning up everything will be a long process, I would highly recommend doing this one sample file at a time. Baby steps.

Couldn't agree more. Also "1 file removal = 1 commit" might make the review easier for us :)

waldyr:
Sorry for my hastiness. It's been quite some time since my last contribution and I'm in an urge for developing more detectors.

@mvz @troessner 

Deleted `two_smelly_files` and fixed their specs by pointing them to new `smelly`.

troessner:
>> Deleted two_smelly_files and fixed their specs by pointing them to new smelly.

Yes, that looks good to me! I think that's exactly the way we need to go about this :)

waldyr:
@mvz @troessner 

I've deleted the clean_due_to_masking/dirty_*.rb files.
I also added a configuration/configuration.reek responsible for suppressing the one and only smell smelly.rb has.

Are we also going to create a single clean file? How is it going to be?

troessner:
>> Are we also going to create a single clean file? How is it going to be?

How about

```Ruby
# Well documented class.
class Awesome
  def hello(argument)
    say argument
  end
end
```

?

waldyr:
@troessner @mvz 

So we have three major files:
- Clean.rb
- Dirty.rb
- Configuration.reek

Buuuuut in reek's tests there are some tests that partially mask a smelly file. Should Dirty.rb reek of two smells and we have two configuration files one for both smells and other to mask one?

troessner:
>> We can either add one more smell in Dirty.rb or create a empty configuration file. Wdyt?

Hmmm, good point. Let's go with one additional smell, that's not too much overhead and it is still self-explanatory ;)

waldyr:
@troessner @mvz 

Everything is all right?

troessner:
>> Everything is alright?

I did a preliminary review and this looks great so far!
How about making the filenames constants as well to prevent fat-fingering and remove duplication?
E.g. "clean.rb" appears multiple times across multiple files.

Other than that I think you should just go on the way you did so far ;)

waldyr:
Good idea. I'm also thinking in refactor the examiner spec because the error output wasn't helping me at all with those shared examples. Wdyt?

mvz:
Yeah, looks good so far.

mvz:
Maybe you should do the examiner refactoring in a next pull request. I'd like to get this one merged.

troessner:
Yep, agree with @mvz on the Examiner

waldyr:
Should I put the constants in the spec file? I think if I do that I would be hiding where I'm defining them.

Could you guys tell me what more can I do here? I need guidance :P

troessner:
Yeah, the spec_helper seems to be the right place.
You can then also:

* look into all the spec/samples_with_* directories and remove / streamline the files in there with what you have done so far. E.g. we can get rid of [this one](https://github.com/troessner/reek/blob/master/spec/samples/source_with_non_ruby_files/uncommunicative_parameter_name.rb) and so on. 
* Kill the whole samples/no_config_file directory
* Streamline "dirty.rb" in samples/masked_by_dotfile

spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -25,47 +25,61 @@
...
+        dir = Pathname.new(tempdir)
+        subdir = dir.join('subdir')
+        subsubdir = subdir.join('subsubdir')
+        config = dir.join('.reek')
+
+        FileUtils.mkdir(subdir)
+        FileUtils.mkdir(subsubdir)
+        FileUtils.touch(config)
+
+        found = described_class.find(current: subsubdir, home: subsubdir)
```
waldyr:
```
dir
 |-- .reek
 |-- subdir
       |-- subsubdir
```

Why this test does not pass? Do we traverse up from :home? @troessner @mvz

troessner: Not from home, but from the current directory, check out: https://github.com/troessner/reek#configuration-loading

spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -83,7 +97,7 @@
 
     describe '.load_from_file' do
       let(:sample_configuration_path) do
-        SAMPLES_PATH.join('configuration/simple_configuration.reek')
+        SAMPLES_PATH.join('configuration/full_mask.reek')
```
waldyr: CONFIG_PATH

troessner:
@waldyr lets stop here and review & merge, wdyt?

waldyr:
Okay. I was looking `spec/samples_with_*` but I didn't find any clean way to remove it.

troessner:
I'll do a review of this tomorrow.

spec/samples/configuration/.reek
```diff
@@ -0,0 +1,9 @@
+---
+IrresponsibleModule:
+  enabled: false
+"spec/samples/three_clean_files/":
+  UtilityFunction:
+    enabled: false
+exclude_paths:
+  - "spec/samples/two_smelly_files/"
+  - "spec/samples/source_with_non_ruby_files/"
```
troessner: This file is confusing since it seems like a grabbag for all kind of things. Why do we need it? By whom is it used for what?

spec/reek/examiner_spec.rb
```diff
@@ -47,14 +47,15 @@
                           filter_by_smells: [],
                           configuration: configuration)
     end
-    let(:path) { SAMPLES_PATH.join('all_but_one_masked/masked.reek') }
-    let(:smelly_file) { Pathname.glob(SAMPLES_PATH.join('all_but_one_masked/d*.rb')).first }
+    let(:path) { CONFIG_PATH.join('partial_mask.reek') }
+    let(:smelly_file) { SMELLY_FILE }
```
troessner: Why not just use SMELLY_FILE directly like you do below in the specs as well?

spec/reek/examiner_spec.rb
```diff
@@ -47,14 +47,15 @@
...
+    let(:expected_first_smell) { 'UncommunicativeVariableName' }
 
     it_should_behave_like 'one smell found'
   end
 
   context 'with a fragrant File' do
-    let(:clean_file) { Pathname.glob(SAMPLES_PATH.join('three_clean_files/*.rb')).first }
+    let(:clean_file) { CLEAN_FILE }
```
troessner: Same question here.

troessner:
1 question and 2 small things and then lgtm!

spec/samples/configuration/.reek
waldyr: https://github.com/troessner/reek/pull/992/files#diff-0a2d7be0c2679312a6bcf13434c60172R34

waldyr:
@troessner comments addressed.

@mvz any comments?

spec/samples/configuration/.reek
troessner: Then rather empty pls

troessner:
Pls squash as well :)

waldyr:
Done!

troessner:
Great job @waldyr - merged!

PR #993, opened 2016-07-27
Make our parameter handling consistent in smell detectors.

troessner:
Follow-up to #940 

This PR makes our parameter handling in smell warnings created in our smell detectors (`SmellDetector#inspect` to be precise) consistent.
The rule is simple (quoting @mvz 
):

>> Another way of looking at it, is that :parameters should contain all values that we interpolate into the smell warning message. No more, no less.

That's what this PR implements. On top of that a significant amount of spec changes because of `mutant`.

### DIFF
--- a/lib/reek/smells/control_parameter.rb
+++ b/lib/reek/smells/control_parameter.rb
@@ -52,12 +52,12 @@ module Reek
       # :reek:FeatureEnvy
       def sniff(ctx)
         ControlParameterCollector.new(ctx).control_parameters.map do |control_parameter|
-          name = control_parameter.name.to_s
+          argument = control_parameter.name.to_s
           smell_warning(
             context: ctx,
             lines: control_parameter.lines,
-            message: "is controlled by argument #{name}",
-            parameters: { name: name })
+            message: "is controlled by argument #{argument}",
+            parameters: { argument: argument })
         end
       end
 
--- a/spec/reek/smells/attribute_spec.rb
+++ b/spec/reek/smells/attribute_spec.rb
@@ -7,6 +7,17 @@ RSpec.describe Reek::Smells::Attribute do
 
   it_should_behave_like 'SmellDetector'
 
+  it 'reports the right values' do
+    src = <<-EOS
+      class Klass
+        attr_writer :my_attr
+      end
+    EOS
+    expect(src).to reek_of(:Attribute,
+                           lines: [2],
+                           message: 'is a writable attribute')
+  end
+
   context 'with no attributes' do
     it 'records nothing' do
       src = <<-EOS
@@ -34,7 +45,7 @@ RSpec.describe Reek::Smells::Attribute do
           attr_writer :my_attr
         end
       EOS
-      expect(src).to reek_of(:Attribute, name: 'my_attr')
+      expect(src).to reek_of(:Attribute, context: 'Klass#my_attr')
     end
 
     it 'does not record writer attribute if suppressed with a preceding code comment' do
@@ -53,7 +64,7 @@ RSpec.describe Reek::Smells::Attribute do
           attr_writer :my_attr
         end
       EOS
-      expect(src).to reek_of(:Attribute, name: 'my_attr')
+      expect(src).to reek_of(:Attribute, context: 'Mod#my_attr')
     end
 
     it 'records accessor attribute' do
@@ -62,7 +73,7 @@ RSpec.describe Reek::Smells::Attribute do
           attr_accessor :my_attr
         end
       EOS
-      expect(src).to reek_of(:Attribute, name: 'my_attr')
+      expect(src).to reek_of(:Attribute, context: 'Klass#my_attr')
     end
 
     it 'records attr defining a writer' do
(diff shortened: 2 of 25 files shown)


troessner:
Ready for review!

2 relevant notes:

1.) Mutant is still failing for 2 reasons: 

1.1) Mutations in `FeatureEnvy#inspect` that are unrelated to what I did
1.2) Stuff like "to_s" vs "to_str" that I don't consider relevant

2.) In basically all of our specs for detectors we check the warnings for the right lines, the right message, the right parameters and so on. Unfortunately we do this in a lot of different ways and in different places (e.g. beginning of the spec vs the end vs scattered across the spec).

E.g. like this:

```Ruby

```

or like this

```Ruby

```

or like this:

```Ruby

```

We should fix this and come up with one way. 

Of the top of my head I can think of

* shared_examples 
* separate module we include in our spec_helper
* via our own spec matcher even though that sucks because you get a generic failure message when it fails

and then put this up at the top of each detector.
But this would have blown up this PR so I'd rather do this in a follow-up PR.

lib/reek/smells/control_parameter.rb
```diff
@@ -52,12 +52,12 @@ class ControlParameter < SmellDetector
...
       def inspect(ctx)
         ControlParameterCollector.new(ctx).control_parameters.map do |control_parameter|
-          name = control_parameter.name.to_s
+          argument = control_parameter.name.to_s
           smell_warning(
             context: ctx,
             lines: control_parameter.lines,
-            message: "is controlled by argument #{name}",
-            parameters: { name: name })
+            message: "is controlled by argument #{argument}",
```
troessner: Actually this should rather be "parameter" and not "argument" but I'd like to fix that in a separate PR.

spec/reek/smells/attribute_spec.rb
```diff
@@ -34,7 +45,7 @@ class Klass
           attr_writer :my_attr
         end
       EOS
-      expect(src).to reek_of(:Attribute, name: 'my_attr')
+      expect(src).to reek_of(:Attribute)
```
mvz: Maybe we can match on context here?

spec/reek/smells/attribute_spec.rb
```diff
@@ -53,7 +64,7 @@ module Mod
...
```
mvz: Same here and later on.

spec/reek/smells/irresponsible_module_spec.rb
```diff
@@ -6,12 +6,14 @@
...
-    expect(src).to reek_of :IrresponsibleModule, name: 'BadClass'
+    expect(src).to reek_of :IrresponsibleModule,
+                           lines: [1],
+                           message: 'has no descriptive comment'
   end
 
   it 'reports a module without a comment' do
     src = 'module BadClass; end'
-    expect(src).to reek_of :IrresponsibleModule, name: 'BadClass'
+    expect(src).to reek_of :IrresponsibleModule
```
mvz: Here too we can match on context.

spec/reek/smells/prima_donna_method_spec.rb
```diff
@@ -3,28 +3,19 @@
 require_relative 'smell_detector_shared'
 
 RSpec.describe Reek::Smells::PrimaDonnaMethod do
+  it 'reports the right values' do
+    src = 'class C; def m!; end; end'
+    expect(src).to reek_of :PrimaDonnaMethod,
+                           lines: [1],
+                           message: 'has prima donna method `m!`',
+                           name: :m!
+  end
```
mvz: :+1: This looks like way to go to replace our old specs.

spec/reek/smells/utility_function_spec.rb
```diff
@@ -153,7 +157,7 @@ def m(a) a.to_s; end
 
   context 'with only one call' do
     it 'reports a call to a parameter' do
-      expect('def simple(arga) arga.to_s end').to reek_of(:UtilityFunction, name: 'simple')
+      expect('def simple(arga) arga.to_s end').to reek_of(:UtilityFunction)
```
mvz: We can match on context for this smell, also.

mvz:
Very nice! Couple of small things, then LGTM!

spec/reek/smells/attribute_spec.rb
troessner: good idea!

spec/reek/smells/prima_donna_method_spec.rb
troessner: Yes, let's do that right away after #998 
I guess we can then kill 1000's of LoC.

troessner:
Addressed all comments! Mutant still fails but I believe it's legit to ignore the failures.
LGTM? :)

spec/reek/smells/utility_function_spec.rb
mvz: @troessner this one still has no context matcher?

mvz:
Nearly there :smile:!

troessner:
Alright, addressed all comments! Merge? :)

troessner:
@mvz I believe you reviewed this one already. I just rebased as well. Merge?

PR #996, opened 2016-07-30
Make Reek more resilient on incomprehensible source.

### LINKED ISSUE #995: Make Reek more resilient on incomprehensible source.
Coming from #942 and #953 we agreed on the following:

We have smell detectors sometimes failing for unexpected reasons, and Reek aborts.
In order to be able to drill down to the root of this problem we decided to:

* Print a (detailed) error
* and skip that source.

so that the user can open up a bug report with enough information for us to fix it.

troessner:
Fixes #995, heavily wip.

### DIFF
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -41,12 +41,13 @@ module Reek
     # @public
     def initialize(source,
                    filter_by_smells: [],
-                   configuration: Configuration::AppConfiguration.default)
+                   configuration: Configuration::AppConfiguration.default,
+                   smell_repository_class: Smells::SmellRepository)
       @source           = Source::SourceCode.from(source)
       @collector        = CLI::WarningCollector.new
-      @smell_types      = Smells::SmellRepository.eligible_smell_types(filter_by_smells)
-      @smell_repository = Smells::SmellRepository.new(smell_types: @smell_types,
-                                                      configuration: configuration.directive_for(description))
+      @smell_types      = smell_repository_class.eligible_smell_types(filter_by_smells)
+      @smell_repository = smell_repository_class.new(smell_types: @smell_types,
+                                                     configuration: configuration.directive_for(description))
     end
 
     # FIXME: Should be named "origin"
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -12,6 +12,19 @@ module Reek
   #
   # :reek:TooManyInstanceVariables: { max_instance_variables: 7 }
   class Examiner
+    INCOMPREHENSIBLE_SOURCE_TEMPLATE = <<-EOS.freeze
+      !!!
+      Source %s can not be processed by Reek.
+      This is most likely a Reek bug.
+      It would be great if you could report this back to the Reek team
+      by opening up a corresponding issue at https://github.com/troessner/reek/issues
+      Make sure to include the source in question, the Reek version
+      and the original exception below.
+
+      Original exception:
+      %s
+      !!!
+    EOS
     #
     # Creates an Examiner which scans the given +source+ for code smells.
     #
@@ -77,6 +90,10 @@ module Reek
     # Runs the Examiner on the given source to scan for code smells
     # and returns the corresponding Examiner instance.
     #
+    # In case one of the smell detectors raises an exception we probably hit a Reek bug.
+    # So we catch the exception here, let the user know something went wrong
+    # and continue with the analysis.
+    #
     # @return an instance of Examiner
     #
     # :reek:TooManyStatements: { max_statements: 6 }
@@ -84,12 +101,20 @@ module Reek
       @run ||= begin
         syntax_tree = source.syntax_tree
         return self unless syntax_tree
-        ContextBuilder.new(syntax_tree).context_tree.each do |element|
-          smell_repository.examine(element)
+        begin
+          examine syntax_tree
+        rescue StandardError => exception
+          $stderr.puts format(INCOMPREHENSIBLE_SOURCE_TEMPLATE, source.origin, exception.inspect)
+        else
+          smell_repository.report_on(collector)
         end
-
-        smell_repository.report_on(collector)
         self
+      end
+    end
+
+    def examine(syntax_tree)
+      ContextBuilder.new(syntax_tree).context_tree.each do |element|
+        smell_repository.examine(element)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


troessner:
This one is depending on #997

mvz:
#997 has been merged, so this can be rebased I think.

troessner:
Ready for review!

mvz:
I'll have a look tomorrow.

spec/reek/examiner_spec.rb
```diff
@@ -108,5 +108,42 @@
         expect(examiner.smells).to eq([])
       end
     end
+
+    context 'with an incomprehensible source that causes Reek to crash' do
+      let(:subject) do
+        examiner = described_class.new 'class C; def does_crash_reek; end; end'
+        allow(examiner).to receive(:smell_repository) do
```
mvz: I'm don't like it if the object under test is stubbed. Can we avoid this by either injecting `SmellRepository`, or stubbing `SmellRepository.new` directly?

spec/reek/examiner_spec.rb
```diff
@@ -108,5 +108,42 @@
         expect(examiner.smells).to eq([])
       end
     end
+
+    context 'with an incomprehensible source that causes Reek to crash' do
+      let(:subject) do
```
mvz: Can you replace this with either `subject do ...`, or (preferred) `let(:examiner) do`?

spec/spec_helper.rb
```diff
@@ -3,6 +3,7 @@
 require_relative '../lib/reek/spec'
 require_relative '../lib/reek/ast/ast_node_class_map'
 require_relative '../lib/reek/configuration/app_configuration'
+require 'active_support/core_ext/string/strip'
```
mvz: Why did you move this?

mvz:
Just a few nits, then it's ready to go.

spec/reek/examiner_spec.rb
troessner: >> Can we avoid this by either injecting SmellRepository

I'd prefer that as well but our current Examiner constructor signature is not supporting so we'd have to add this. Which would mean that it would take 4 arguments, which I find quite a lot but still manageable. Are you suggesting to update the Examiner constructor to be like this:

```Ruby

    def initialize(source,
                   filter_by_smells: [],
                   configuration: Configuration::AppConfiguration.default,
                   smell_repository: Smells::SmellRepository.new)
end
```

?

>> or stubbing SmellRepository.new directly?

I can't see the benefit. For me this would be even worse since it is more opaque - you're stubbing something that is not even called directly thus making it harder to build up a mental image. Why do you think this is better?

spec/reek/examiner_spec.rb
troessner: Good point, fixed!

spec/spec_helper.rb
troessner: The overall rationale here was that we use `strip_heredoc` in multiple places so it would make sense to require this in the spec_helper.
But you're right, this really doesn't belong to this PR, I just undid this change.

spec/reek/examiner_spec.rb
mvz: > Are you suggesting to update the Examiner constructor to be like this:

We probably need to inject the class instead of an instance, but yes, something like that.

> > or stubbing SmellRepository.new directly?
> I can't see the benefit. For me this would be even worse since it is more opaque - you're stubbing something that is not even called directly thus making it harder to build up a mental image. Why do you think this is better?

I think `SmellRepository.new` is actually called directly in `Examiner`. But to answer more broadly, I have found again and again that stubbing the object under test is really the most dangerous option, since it tends to hide bugs in the implementation of the object itself.

lib/reek/examiner.rb
```diff
@@ -12,6 +12,19 @@ module Reek
   #
   # :reek:TooManyInstanceVariables: { max_instance_variables: 7 }
   class Examiner
+    INCOMPREHENSIBLE_SOURCE_TEMPLATE = <<-EOS.freeze
+      !!!
+      Source %s can not be processed by Reek.
+      This is most likely a Reek bug.
+      It would be great if you could report this back to the Reek team
+      by opening up a corresponding issue at https://github.com/troessner/reek/issues
+      Make sure to include the source in question, the Reek version
```
Drenmi: Just a thought. Would it be feasible to have all this information ready to be copied and pasted from the error message, rather than having the user go and gather it?

troessner: Awesome idea!
I'll add this, thanks!

spec/reek/examiner_spec.rb
```diff
@@ -108,5 +108,42 @@
         expect(examiner.smells).to eq([])
       end
     end
+
+    context 'with an incomprehensible source that causes Reek to crash' do
+      subject do
+        examiner = described_class.new 'class C; def does_crash_reek; end; end'
+        allow(examiner).to receive(:smell_repository) do
```
mvz: (comment copied so it is not hidden)

> Are you suggesting to update the Examiner constructor to be like this:

We probably need to inject the class instead of an instance, but yes, something like that.

> >  or stubbing SmellRepository.new directly?

>  I can't see the benefit. For me this would be even worse since it is more opaque - you're stubbing something that is not even called directly thus making it harder to build up a mental image. Why do you think this is better?

I think SmellRepository.new is actually called directly in Examiner. But to answer more broadly, I have found again and again that stubbing the object under test is really the most dangerous option, since it tends to hide bugs in the implementation of the object itself.

troessner:
@mvz addressed of all your feedback in 2 separate commits, the first one making SmellRepository injectable and the second one making the specs less brittle.

I'll address 

>> Would it be feasible to have all this information ready to be copied and pasted from the error message, rather than having the user go and gather it?

in a separate pull request to not block this one (see #1019)

LGTM?

mvz:
It looks like you're not using the new dependency injection possibility in the specs?

troessner:
>> It looks like you're not using the new dependency injection possibility in the specs?

Exactly! I started out with adding it but when it came to the spec itself I realized that RSpec's awesome 

```Ruby
class_double('Reek::Smells::SmellRepository').as_stubbed_const
```

works as well for this very spec. I still left the DI in there because a) I think we should start using DI way more in our codebase and b) starting to DI SmellRepository might be useful for subsequent specs. That's why it's 2 separate commits.

I can also remove the DI for SmellRepository but I think it makes sense to keep it for above reasons.

WDYT?

mvz:
Ok, clear.

PR #997, opened 2016-07-30
Decouple initialization from action in Examiner.

troessner:
I feel that the architecture of our Examiner is not where it could be.

**I see two problems:**

1.) We do **everything** in the initializer which couples initialization to action which is not best practice
2.) Regarding [1]: I find this counter-intuitive and always felt uneasy about it. I always thought "but wait, don't I have to call anything on that?"
I also believe it is not idiomatic Ruby compared with other gems (e.g. `mutant`)
3.) It makes testing the Examiner very hard and awkward which I realized working on #996 

I think we should fix this by decoupling initialization from our `run` action, with this PR being the first draft for it.

I wanted to discuss this first before going deeper ( there are still some failing examiner specs).
WDYT @waldyr && @mvz ?

### DIFF
--- a/spec/reek/examiner_spec.rb
+++ b/spec/reek/examiner_spec.rb
@@ -28,13 +28,13 @@ RSpec.describe Reek::Examiner do
   let(:expected_first_smell) { 'NestedIterators' }
 
   context 'with a fragrant String' do
-    let(:examiner) { described_class.new('def good() true; end') }
+    let(:examiner) { described_class.run('def good() true; end').run }
 
     it_should_behave_like 'no smells found'
   end
 
   context 'with a smelly String' do
-    let(:examiner) { described_class.new('def fine() y = 4; end') }
+    let(:examiner) { described_class.run('def fine() y = 4; end') }
     let(:expected_first_smell) { 'UncommunicativeVariableName' }
 
     it_should_behave_like 'one smell found'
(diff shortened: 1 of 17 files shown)


mvz:
Yes, this makes a lot of sense.

waldyr:
I agree. In the case I want to run the method after the initialization I tend to create a class method with the same name the method I'm willing to run. For example:
```
def self.run(*args)
  new(*args).run
end
```
But it might not be needed

troessner:
@waldyr good idea!
I was actually eyeing [Procto](https://github.com/snusnu/procto) here as since we're basically talking about method objects, which I particularly like. But that's another discussion, let's have the `run` method on class level like you suggested and the `run` method on instance level for power users like us.

troessner:
I updated the PR and introduced `Examiner.run`.
Left to do:

- Fix the specs (the goddamn shared example, something we really should kill)
- Adds docs (see TODO)
- Polish here and there
- Add this change to our upgrade section when we bump Reek to 5

spec/reek/examiner_spec.rb
```diff
@@ -55,15 +55,38 @@
...
+    context 'returns an Examiner' do
+      it 'has been run on the given source' do
+        expect(examiner.description).to eq('string')
+      end
+
+      it 'has the right smells' do
+        smells = examiner.smells
+        expect(smells[0].message).to eq('has no descriptive comment')
+        expect(smells[1].message).to eq("has the name 'f'")
+        expect(smells[2].message).to eq("has the name 'C'")
```
troessner: The reason why I'm doing it like this is #999

mvz: :crying_cat_face:

troessner:
Closing this one in favor of #1008

PR #1001, opened 2016-07-31
Fix our rspec matcher warnings to include all details.

### LINKED ISSUE #994: Make checking the warnings in smell detectors consistent across our specs
Follow-up from #993 

In basically all of our specs for detectors we check the warnings for the right lines, the right message, the right parameters and so on. Unfortunately we do this in a lot of different ways and in different places (e.g. beginning of the spec vs the end vs scattered across the spec).

E.g. like this:

```Ruby
    context 'when a smells is reported' do
      let(:warning) do
        src = <<-EOS
        ...
        EOS
        Reek::Examiner.new(src, filter_by_smells: ['UtilityFunction']).smells[0]
      end

      it 'reports the line number of the method' do
        expect(warning.lines).to eq([1])
      end
  end
```

or like this (the worst in my mind)

```Ruby
      let(:warning) do
        src = 'def cc(arga = true) end'
        ctx = Reek::Context::MethodContext.new(nil, Reek::Source::SourceCode.from(src).syntax_tree)
        detector.inspect(ctx).first
      end

      it 'reports the right parameter' do
        expect(warning.parameters[:parameter]).to eq('arga')
      end
```

or like this (the most concise way but the error message isn't really helpful in case the spec fails):

```Ruby
  it 'reports a class without a comment' do
    src = 'class BadClass; end'
    expect(src).to reek_of :IrresponsibleModule,
                           lines: [1],
                           message: 'has no descriptive comment'
  end
```

We should fix this and come up with one way. 

Of the top of my head I can think of

* shared_examples 
* separate module we include in our spec_helper
* via our own spec matcher even though that sucks because you get a generic failure message when it fails

and then put this up at the top of each detector.

troessner:
Fixes #994

### DIFF
--- a/spec/reek/spec/should_reek_of_spec.rb
+++ b/spec/reek/spec/should_reek_of_spec.rb
@@ -1,6 +1,8 @@
 require 'pathname'
 require_relative '../../spec_helper'
 require_lib 'reek/spec'
+require 'active_support/core_ext/string/strip'
+require 'active_support/core_ext/hash/except'
 
 RSpec.describe Reek::Spec::ShouldReekOf do
   describe 'smell type selection' do
@@ -99,8 +101,8 @@ RSpec.describe Reek::Spec::ShouldReekOf do
     end
 
     context 'smell type is matching but smell details are not' do
-      let(:smelly_code) { 'def dummy() y = 4; end' }
-      let(:matcher) { Reek::Spec::ShouldReekOf.new(:UncommunicativeVariableName, name: 'x') }
+      let(:smelly_code) { 'def double_thing() @other.thing.foo + @other.thing.foo end' }
+      let(:matcher) { Reek::Spec::ShouldReekOf.new(:DuplicateMethodCall, name: 'foo', count: 15) }
 
       it 'is falsey' do
         expect(matcher.matches?(smelly_code)).to be_falsey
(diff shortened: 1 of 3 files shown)


mvz:
Awesome, you're on a roll!

troessner:
There are a couple of mutant failures I don't understand, e.g. this one:

>>
@@ -1,7 +1,4 @@
 def to_hash
-  stringified_params = Hash[parameters.map do |key, val|
-    [key.to_s, val]
-  end]
   base_hash.merge(stringified_params)
 end

This one doesn't make any sense to me. When I just go and delete this block and then run the specs there are failing specs. @mvz any idea what's up with that?
Other than that I believe this PR is mergeable as it is.
Let's get this going as soon as possible so we can start to kill hundreds of lines of bad specs ;)

troessner:
Ready for review, lgtm! :)

spec/reek/spec/should_reek_of_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'pathname'
 require_relative '../../spec_helper'
 require_lib 'reek/spec'
+require 'active_support/core_ext/string/strip'
+require 'active_support/core_ext/hash/except'
```
mvz: I think these lines should go into `lib`.

mvz: Ah no, only the second line.

troessner: I dont follow. Why would i put that into lib and thus make activesupport a runtime dependency?

mvz: You're using `except` in the matcher, aren't you?

troessner: F**k. Thanks for paying attention (because I clearly didn't), my bad, pull request to fix this incoming.

troessner: See #1013, lgtm

PR #1002, opened 2016-08-02
Enable bundler in test

soutaro:
I did `git clone` the repo and tried to run test as `rake`. The result was an error because I have another version of JSON library installed:

```
/Users/soutaro/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/runtime.rb:35:in `block in setup': You have already activated json 2.0.2, but your Gemfile requires json 1.8.3. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```

I tried `bundle exec rake` as recommended, but did not solve the issue.

This PR is to make test setup Bundler. I'm not very sure this fix makes sense (but it works!).

### DIFF
--- a/tasks/test.rake
+++ b/tasks/test.rake
@@ -4,13 +4,13 @@ require 'rspec/core/rake_task'
 namespace 'test' do
   RSpec::Core::RakeTask.new('spec') do |t|
     t.pattern = 'spec/reek/**/*_spec.rb'
-    t.ruby_opts = ['-rsimplecov -Ilib -w']
+    t.ruby_opts = ['-rbundler/setup -rsimplecov -Ilib -w']
   end
 
   desc 'Tests code quality'
   RSpec::Core::RakeTask.new('quality') do |t|
     t.pattern = 'spec/quality/**/*_spec.rb'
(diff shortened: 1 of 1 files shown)


mvz:
Hi, this is a known bug in Bundler: https://github.com/bundler/bundler/issues/4758.

mvz:
This is clever workaround though! :heart:

troessner:
I don't see anything that speaks against merging this PR. Objections anyone?

mvz:
If it solves the Bundler problem for now, I'm ok with merging.

soutaro:
I tried with ruby-2.2.5 and `bundle exec rake` on `mater` runs.
I'm not sure yet, but the issue should be because of JSON version in Gemfile.

mvz:
I think @troessner and I agree here. Merging....

soutaro:
I am confusing... The reason it run with ruby-2.2.5 is just because I have no JSON.gem installed...

Thanks for the review! I'm happy to see this got merged 🎆

troessner:
@soutaro just released 4.2.2 with your bugfix, thanks a lot!

PR #1004, opened 2016-08-04
Add SexpExtensions::CSendNode for safe navigation operator

soutaro:
Currently `:csend` node is not supported, and the following Ruby code may make reek crash:

```rb
foo&.bar()
```

This PR is to fix the issue by adding `CSendNode`.

### DIFF
--- a/spec/reek/ast/sexp_extensions_spec.rb
+++ b/spec/reek/ast/sexp_extensions_spec.rb
@@ -124,6 +124,26 @@ RSpec.describe Reek::AST::SexpExtensions::DefNode do
   end
 end
 
+RSpec.describe Reek::AST::SexpExtensions::CSendNode do
+  let(:node) { Reek::Source::SourceCode.from('1&.foo(true)').syntax_tree }
+
+  it 'has a class including SendNode' do
+    expect(node.class.included_modules).to include(Reek::AST::SexpExtensions::SendNode)
+  end
+
+  it 'has a receiver' do
+    expect(node.receiver).to eq(sexp(:int, 1))
+  end
+
+  it 'has a name' do
+    expect(node.name).to eq(:foo)
+  end
+
+  it 'has arguments' do
+    expect(node.args).to eq([sexp(:true)])
+  end
+end
+
 RSpec.describe Reek::AST::SexpExtensions::DefsNode do
   context 'with no parameters' do
     let(:node) { sexp(:defs, sexp(:lvar, :obj), :hello, sexp(:args)) }
--- a/lib/reek/ast/sexp_extensions/send.rb
+++ b/lib/reek/ast/sexp_extensions/send.rb
@@ -53,6 +53,7 @@ module Reek
       end
 
       Op_AsgnNode = SendNode
+      CSendNode = SendNode
     end
   end
 end


troessner:
What Reek version are you using?
The current version of Reek does **not** crash on this code:

>>   
( $ (master)) echo "class C; def m; foo&.bar(); end; end" | reek

STDIN -- 3 warnings:
  [1]:IrresponsibleModule: C has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]
  [1]:UncommunicativeMethodName: C#m has the name 'm' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Method-Name.md]
  [1]:UncommunicativeModuleName: C has the name 'C' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Module-Name.md]

and

>>
( $ (master)) reek -v
reek 4.2.2

soutaro:
Oops, I'm sorry for wrong example. This is the shortest Ruby program to reproduce I found:

```rb
def f
  x&.tap {|_| }
end
```

and following is the trace:

```
/Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:138:in `block in ignored_iterator?': undefined method `name' for "s(:csend,\n  s(:send, nil, :x), :tap)":#<Class:0x007fee34813a38> (NoMethodError)
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:138:in `any?'
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:138:in `ignored_iterator?'
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:129:in `increment_depth'
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:109:in `block in scout'
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:108:in `each'
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:108:in `flat_map'
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:108:in `scout'
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:85:in `find_candidates'
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:73:in `find_violations'
	from /Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:46:in `inspect'
	from /Users/soutaro/src/reek/lib/reek/smells/smell_detector.rb:47:in `run_for'
	from /Users/soutaro/src/reek/lib/reek/smells/smell_repository.rb:47:in `block in examine'
	from /Users/soutaro/src/reek/lib/reek/smells/smell_repository.rb:46:in `each'
	from /Users/soutaro/src/reek/lib/reek/smells/smell_repository.rb:46:in `examine'
	from /Users/soutaro/src/reek/lib/reek/examiner.rb:81:in `block in run'
	from /Users/soutaro/src/reek/lib/reek/context/code_context.rb:85:in `each'
	from /Users/soutaro/src/reek/lib/reek/context/code_context.rb:87:in `block in each'
	from /Users/soutaro/src/reek/lib/reek/context/code_context.rb:86:in `each'
	from /Users/soutaro/src/reek/lib/reek/context/code_context.rb:86:in `each'
	from /Users/soutaro/src/reek/lib/reek/examiner.rb:80:in `run'
	from /Users/soutaro/src/reek/lib/reek/examiner.rb:37:in `initialize'
	from /Users/soutaro/src/reek/lib/reek/cli/command/report_command.rb:24:in `new'
	from /Users/soutaro/src/reek/lib/reek/cli/command/report_command.rb:24:in `block in populate_reporter_with_smells'
	from /Users/soutaro/src/reek/lib/reek/cli/command/report_command.rb:23:in `each'
	from /Users/soutaro/src/reek/lib/reek/cli/command/report_command.rb:23:in `populate_reporter_with_smells'
	from /Users/soutaro/src/reek/lib/reek/cli/command/report_command.rb:15:in `execute'
	from /Users/soutaro/src/reek/lib/reek/cli/application.rb:28:in `execute'
	from bin/reek:13:in `<main>'
```

The motivating example to me is one script of my app. The trace looks the same, but error message is a bit different.

```
/Users/soutaro/src/reek/lib/reek/smells/nested_iterators.rb:138:in `block in ignored_iterator?': undefined method `name' for #<#<Class:0x007fc9141413d0>:0x007fc914141358> (NoMethodError)
```

Anyway, the error looks like because of `:csend` node.

soutaro:
I checked out latest master branch, and run `bin/reek` in the directory.

```
$ git rev-parse HEAD
39919a9514c4495b74e16080aaeb769f08506f07
$ bin/reek --version
reek 4.2.2
```

troessner:
>> Oops, I'm sorry for wrong example. This is the shortest Ruby program to reproduce I found:

Acknowledged - I'll review this bad boy tonight!

spec/reek/ast/sexp_extensions_spec.rb
```diff
@@ -124,6 +124,26 @@
...
 end
 
+RSpec.describe Reek::AST::SexpExtensions::CSendNode do
+  let(:node) { Reek::Source::SourceCode.from('1&.foo(true)').syntax_tree }
+
+  it 'has a class including SendNode' do
+    expect(node.class.included_modules).to include(Reek::AST::SexpExtensions::SendNode)
+  end
+
+  it 'has receiver method' do
```
troessner: Lets just say "has a receiver" to make this less specific.

spec/reek/ast/sexp_extensions_spec.rb
```diff
@@ -124,6 +124,26 @@
...
+    expect(node.receiver).to eq(sexp(:int, 1))
+  end
+
+  it 'has name method' do
```
troessner: Lets go with "has a name" :)

spec/reek/ast/sexp_extensions_spec.rb
```diff
@@ -124,6 +124,26 @@
...
+    expect(node.name).to eq(:foo)
+  end
+
+  it 'has args method' do
```
troessner: "has arguments" :)

troessner:
Just some tiny comments and then lgtm!
I'd also release a new version right away

mvz:
I'd like to also see that sample code that crashes turned into a spec. Otherwise, it's not clear why we're having a `CSendNode` in the first place.

soutaro:
I fixed spec descriptions, and squashed!

spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -271,6 +271,15 @@ def bad(fred)
...
     end
   end
+
+  it 'should analyze safe-navigation-operator without error' do
+    src = '
+      def f
+        x&.tap {|_| }
+      end
+    '
+    expect(src).not_to reek_of(:NestedIterators, count: 2)
```
soutaro: When the analysis raises an error (without the fix), this spec fails as `F`. However, I don't think this `expect` is good for one in my mind.

I want something like:

```rb
  expect {
    # run reek against src
  }.not_to raise_error
```

Any advice? (or is this good enough?)

troessner: I don't think this test makes much sense here. The error just happened there coincidentally. Having this test here is going to be super-confusing for us in 6 months.

How about:

* creating a directory in the [samples](https://github.com/troessner/reek/tree/master/spec/samples) directory called "super_smelly"
* put "inline.rb", "optparse.rb" and "redcloth.rb" in there
* create a new file called "syntax_checks.rb"
* Put the code above in it
* Update our [samples feature](https://github.com/troessner/reek/blob/master/features/samples.feature) to include this as well

WDYT? cc @mvz

mvz: I'd prefer to have some actual tests for NestedIterators with safe navigation that show it detecting actual smells. I vaguely recall having some of those tests lying around from earlier when making them pass was blocked by an issue in the parser gem.

TL;DR: I'm ok with merging this one without a test for NestedIterators right now (so basically removing this these lines from the PR), and then I will add a set of tests this weekend.

mvz: That said, it may be useful to to have some newer code in our smelly samples feature.

soutaro: Thanks, I completely agree your idea. I also think there is better place to have the example in. So, I will just drop this example.

> I will add a set of tests this weekend.

Sounds great 👏

troessner:
Merged! Thanks a lot @soutaro - I just released 4.2.3 with your bugfix!

soutaro:
Thanks! Now, I can use reek for my project running on ruby-2.3 🎆

PR #1006, opened 2016-08-05
[Fix #999] Rename `#inspect` method to `#sniff`

### LINKED ISSUE #999: Weird edge case bug in our specs
I stumbled across this around a year ago in a new spec I was working on but managed to work around it. Now it happened again so I thought I'd document it here.
I happened while working [on this PR](https://github.com/troessner/reek/pull/997) with this innocent looking spec:

```Ruby
  describe '.run' do
    let(:examiner) do
      described_class.run('class C; def f; end; end')
    end

    context 'returns an Examiner' do
      it 'has the right smells' do
        expect(examiner.smells).to eq([])
      end
    end
  end
end
```

Running the spec like this:

````
DEBUG=1 bx rspec spec/reek/examiner_spec.rb:73
```

failed with:

```
Run options: include {:focus=>true, :locations=>{"./spec/reek/examiner_spec.rb"=>[73]}}
F

Failures:

  1) Reek::Examiner.run returns an Examiner has the right smells
     Failure/Error:
       def inspect(ctx)
         return [] if descriptive?(ctx) || ctx.namespace_module?
         expression = ctx.exp
         [smell_warning(
           context: ctx,
           lines: [expression.line],
           parameters: { name: expression.name })]

     ArgumentError:
       wrong number of arguments (given 0, expected 1)
     # ./lib/reek/smells/irresponsible_module.rb:22:in `inspect'
     # ./spec/reek/examiner_spec.rb:76:in `block (4 levels) in <top (required)>'

Finished in 0.0037 seconds (files took 0.65824 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/reek/examiner_spec.rb:73 # Reek::Examiner.run returns an Examiner has the right smells
```

The error itself doesn't make any sense at this point. Also, look at the weird stacktrace which basically goes from IrresponsibleModule to the spec directly.

I think this has somehow something to do with rspec, because when I do it like this:

```Ruby
      it 'has the right smells' do
        smells = examiner.smells
        expect(smells).to eq([])
      end
```

the exception is raised in **this** line:

```Ruby
expect(smells).to eq([])
```

and not in this:

```Ruby
smells = examiner.smells
```

I can even insert a breakpoint before

```Ruby
expect(smells).to eq([])
```
and "cd" / walk around in `smells` itself with everything looking as expected.

Additionally the whole spec works if I do it like this:

```Ruby
it 'has the right smells' do
  smells = examiner.smells
  expect(smells[0].message).to eq('has no descriptive comment')
  expect(smells[1].message).to eq("has the name 'f'")
  expect(smells[2].message).to eq("has the name 'C'")
end
```

Drenmi:
It has been said:

> There are only two hard things in Computer Science:
> cache invalidation and naming things.
>
> -- Phil Karlton

I hereby humbly offer my suggestion. 😁 

I found it fitting, because not only is *sniff* the imperative verb associated with detecting a smell, it also has the informal meaning of discovering something through investigation.

### DIFF
--- a/lib/reek/smells/attribute.rb
+++ b/lib/reek/smells/attribute.rb
@@ -30,7 +30,7 @@ module Reek
       #
       # @return [Array<SmellWarning>]
       #
-      def inspect(ctx)
+      def sniff(ctx)
         attributes_in(ctx).map do |attribute, line|
           smell_warning(
             context: ctx,
--- a/lib/reek/smells/boolean_parameter.rb
+++ b/lib/reek/smells/boolean_parameter.rb
@@ -20,7 +20,7 @@ module Reek
       # @return [Array<SmellWarning>]
       #
       # :reek:FeatureEnvy
-      def inspect(ctx)
+      def sniff(ctx)
         ctx.default_assignments.select do |_param, value|
           [:true, :false].include?(value.type)
         end.map do |parameter, _value|
--- a/lib/reek/smells/class_variable.rb
+++ b/lib/reek/smells/class_variable.rb
@@ -24,7 +24,7 @@ module Reek
       #
       # @return [Array<SmellWarning>]
       #
-      def inspect(ctx)
+      def sniff(ctx)
         class_variables_in(ctx.exp).map do |variable, lines|
           smell_warning(
             context: ctx,
--- a/lib/reek/smells/control_parameter.rb
+++ b/lib/reek/smells/control_parameter.rb
@@ -50,7 +50,7 @@ module Reek
       # @return [Array<SmellWarning>]
       #
       # :reek:FeatureEnvy
-      def inspect(ctx)
+      def sniff(ctx)
         ControlParameterCollector.new(ctx).control_parameters.map do |control_parameter|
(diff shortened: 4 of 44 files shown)


troessner:
Very, very awesome!
LGTM. Only mutant is failing right now and that is ok-ish, since you just renamed the method.
I also like "sniff". @mvz @waldyr wdyt?

Drenmi:
Yes. Been staring at the mutant errors for 35 minutes, but still can't make sense of it since I, as you said, just renamed the method. 😞

mvz:
Mutant considers all the classes that use the method as channged, so will be checking all of them, I guess.

troessner:
Lets merge this - thanks a lot @Drenmi

mvz:
Uhh ... this means master is reek 5 now.

mvz:
Oh, no it's not :smile:.

troessner:
Why? `inspect` was never intended to be used by api users directly.

mvz:
> Why? inspect was never intended to be used by api users directly.

I was confused.

PR #1008, opened 2016-08-05
[Breaking change] Decouple initialization from action in Examiner.

troessner:
The same as #997 but pointing to the reek-5 branch this time. 
Ready for review!

There are quite a few mutant failures I don't understand. 
E.g.:

````
     ContextBuilder.new(syntax_tree).context_tree.each do |element|
-      smell_repository.examine(element)
+      smell_repository.examine(nil)
     end
````

When I add this change manually literally **all** the specs for Examiner break. 

Same goes for 

```
     ContextBuilder.new(syntax_tree).context_tree.each do |element|
       smell_repository.examine(element)
     end
-    smell_repository.report_on(collector)
+    nil
     self
```

I don't understand what is happening here. But let's ignore this, this is a separate issue. 

I do however feel that this change has been spec'ed enough and I don't want to address the [mutant issue](https://github.com/troessner/reek/issues/1007) here.

Speedy review would be great so we can get this on the road :)

### DIFF
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -10,7 +10,7 @@ module Reek
   #
   # @public
   #
-  # :reek:TooManyInstanceVariables: { max_instance_variables: 6 }
+  # :reek:TooManyInstanceVariables: { max_instance_variables: 7 }
   class Examiner
     #
     # Creates an Examiner which scans the given +source+ for code smells.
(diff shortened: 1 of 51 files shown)

### CODE REVIEW

File: lib/reek/examiner.rb [hunk 1]
```diff
@@ -10,7 +10,7 @@ module Reek
   #
   # @public
   #
-  # :reek:TooManyInstanceVariables: { max_instance_variables: 6 }
+  # :reek:TooManyInstanceVariables: { max_instance_variables: 7 }
```
troessner: That's because of the memoization in "#run" below. We really should fix the detector to not count this as instance variables.

docs/API.md
```diff
@@ -32,7 +32,7 @@ source = <<-EOS
 EOS
 
 reporter = Reek::Report::TextReport.new
-examiner = Reek::Examiner.new source
+examiner = Reek::Examiner.run(source)
```
mvz: I'm not finding it very intuitive that `Reek::Examiner.run` or even `#run` should return an examiner. Why not write this out like so:

```
examiner = Reek::Examiner.new(source)
examiner.run
```

It's a little more verbose, but I think it's clearer.

mvz:
Perhaps an alternative way to decouple these things is to only 'run' when `Examiner#smells` is called. That way, we don't run in initialization, but we still don't require an explicit call to `run` (and why would the user of this class decide when the actual sniffing takes place?).

troessner:
>> Perhaps an alternative way to decouple these things is to only 'run' when Examiner#smells is called. That way, we don't run in initialization, but we still don't require an explicit call to run (and why would the user of this class decide when the actual sniffing takes place?).

Excellent idea! I'll update the PR respectively tomorrow.

troessner:
Closing this one in favor of #1015

PR #1009, opened 2016-08-06
Update default config for UtilityFunction

mvz:
This is an automatically generated change.

### DIFF
--- a/defaults.reek
+++ b/defaults.reek
@@ -119,3 +119,4 @@ UnusedPrivateMethod:
 UtilityFunction:
   enabled: true
   exclude: []
+  public_methods_only: false


troessner:
Where is this coming from all of a sudden?

mvz:
We have a task that regenerates `defaults.reek` based on the defaults in the code. I don't know why this wasn't triggered before, but now it really wants to add this line.

troessner:
>> I don't know why this wasn't triggered before

Yep, that's what I don't understand. Looking [at it](https://github.com/troessner/reek/blob/master/tasks/configuration.rake) I can't see the reason for this behaviour. Mysterious, isn't it? 😆

troessner:
Mutant complains [a lot](https://travis-ci.org/troessner/reek/jobs/150290362) because were doing

>> --since master^

and not

>> --since master

because otherwise our build on master would fail.....
@mbj any idea how to proceed here?
@mvz I think we should disable mutant for now, fix our setup and then re-enable it. wdyt?

mvz:
I think if you merge this mutant will shut up.

After that, I think:
* I may look into submitting a patch for mutant to not fail when using `--since master` on master.
* We should revert to using `--since master`.

troessner:
>> I think if you merge this mutant will shut up.

Sure, my point was rather that our mutant set up is still not right.

>>
After that, I think:
I may look into submitting a patch for mutant to not fail when using --since master on master.
We should revert to using --since master.

Now that sounds like a plan!

mbj:
> Sure, my point was rather that our mutant set up is still not right.

`--since master` in theory should result in an empty diff (so empty test selection) on master. Can you confirm this happened? (AFAIK empty mutation test in master is fine, as your branch was green before merge). If you are not happy with that you likely need to add some special code that finds the pre merge master head and special case this for builds on master.

OT: I typically recommend using `--since HEAD~1` as the better workflow strategy, mostly because it does not result in ever increasing mutant selection on bigger branches, and it lets the programmer care more about "each commit part of a PR should be green".

troessner:
I am not sure I follow :)

* HEAD^1 made this PR red
* HEAD breaks our master build

So i am not sure how to continue now. Dont you have the same problem in the mutant repo?

mbj:
> HEAD^1 made this PR red

Because there are alive mutations in the working set you should have killed or explicitly ignored. Thats the intend workflow.

troessner:
>> Because there are alive mutations in the working set you should have killed or explicitly ignored. Thats the intend workflow.

Ok, fair enough. But if you wonder why mutant adoption is slow, here is a big reason for this. Just imagine it would not have been @mvz issuing this pull request, but just a casual guy who thought he'd make a contribution to an OSS project. So he submits an innocent change like above and all of sudden his PR is red and mutant blows up right in his face. What impression is that gonna leave on this guy? What is he supposed to do?

I'm pretty sure he'll never come back and rightfully so.

Seriously, HEAD^1 is NOT an option. Not for this team or for any other team except for the mutant team maybe. It seems to be a workflow that only works in that rare situation that you have a very tight-knit OSS team that has a stellar test coverage and code that is almost completely non-redundant. So fine for mutant itself, but literally inapplicable for 99.99% of all other projects out there.

If we integrate mutant properly (and I'm in favor of that for sure since I like the idea a lot) it should be via 

>> --since master

And for that we should fix that it then breaks on our master build.

What do you think?

mvz:
I agree that *at the present time* our mutant setup should make master green even if we choose to merge branches where mutant fails. This means using `--since master`. In turn, this means we should figure out why `--since master` fails on master at the moment. I'm planning to look into this soonish.

mbj:
> What is he supposed to do?

In prefered order descending:

* Improve his code (refactor the mutation away),
* Improve his tests to cover the alive mutaiton
* Add a subject exclude.

> I'm pretty sure he'll never come back and rightfully so.

If you do not want feedback from semantic coverage tools, do not use semantic coverage tools. But this eliminates bug reports the same way the absence of QA depeartments eliminates them.

> Seriously, HEAD^1 is NOT an option. Not for this team or for any other team except for the mutant team maybe.

I use it sucessfully on right now 5 commercial teams. Does not imply it works for you and sorry I proposed it.

> So fine for mutant itself, but literally inapplicable for 99.99% of all other projects out there.

Add subject excludes or build up a whitelist for the part of the code that is 100% mutation covered and should stay there. This works very decently to adapt for different teams needs.

> I agree that at the present time our mutant setup should make master green even if we choose to merge branches where mutant fails.

Sounds like you want the whitelist approach where you only run mutant on code to be currently known tobe 100% covered.

PR #1010, opened 2016-08-06
Add specs for NestedIterators with safe navigation

mvz:
Follow-up for #1004.

### DIFF
--- a/spec/reek/smells/nested_iterators_spec.rb
+++ b/spec/reek/smells/nested_iterators_spec.rb
@@ -193,6 +193,30 @@ RSpec.describe Reek::Smells::NestedIterators do
     end
   end
 
+  it 'reports nested iterators called via safe navigation' do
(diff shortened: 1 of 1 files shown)


spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -193,6 +193,30 @@ def foo
...
+          bar&.each do |bottle|
+            puts bottle
+          end
+        end
+      end
+    EOS
+    expect(source).to reek_of(:NestedIterators)
+  end
+
+  it 'does not reports unnested iterators called via safe navigation' do
```
troessner: "report" and then lgtm!

mvz: Done.

troessner:
And the same mutant problem. As far as i am concerned we need to fix this, fast.

PR #1013, opened 2016-08-06
Remove accidentally added ActiveSupport dependency.

troessner:


troessner:
You have a tendency to not press "delete branch" 😆

mvz:
> You have a tendency to not press "delete branch" :laughing:

It doesn't seem to be an option in the mobile interface :disappointed:.

troessner:
Which is exactly the reason why i use the desktop view instead of githubs abysmal mobile UI :)

PR #1017, opened 2016-08-09
Add Instance Variable Assumption smell detector

waldyr:
#632 #633

This one might be difficult to have in practice.

This is a work in progress but review and comments are welcome :)


waldyr:
@troessner @mvz Closed the last PR because I deleted the repository with the branch. I've addressed the last comments from #959. Any more comments?

troessner:
I'll review this one the next days, promised.

CONTRIBUTING.md
```diff
@@ -105,6 +105,9 @@ git rebase -i master
 # squash squash squash
 git push -f origin
 ```
+## Failing Cucumber Scenarios
+
+If there is a failing scenario and you can not figure out why it is failing, just run the failing scenario: `bundle exec cucumber features/failing_scenario.feature:line`. By doing so Aruba will leave its set up in the `tmp/aruba` directory. Then you can run the scenario's Reek incantation there and have a better debugging.
```
troessner: >> Then you can run the scenario's Reek incantation there and have a better debugging.

I am not so sure this sentence actually makes sense. Can you describe what you are trying to say here? Maybe we'll find an easier explanation :)

features/configuration_loading.feature
```diff
@@ -44,8 +45,9 @@ Feature: Offer different ways how to load configuration
     Then the exit status indicates smells
     And it reports:
       """
-      smelly.rb -- 3 warnings:
+      smelly.rb -- 4 warnings:
         [4, 5]:DuplicateMethodCall: Smelly#m calls @foo.bar 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
         [4, 5]:DuplicateMethodCall: Smelly#m calls puts @foo.bar 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
+        [2]:InstanceVariableAssumption: Smelly assumes too much for instance variable @foo [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
         [3]:UncommunicativeMethodName: Smelly#m has the name 'm' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Method-Name.md]
```
troessner: Can we rather adjust the example code we use here and above to **not** cause the InstanceVariableAssumption warnings?
Did you already rebase against master after #992 has been merged?
Adding a new smell detector should **not** trigger so many changes in unrelated cucumber features. Ideally, it wouldn't cause any.

features/programmatic_access.feature
```diff
@@ -19,6 +19,7 @@ Feature: Using Reek programmatically
       """
       calls @foo.bar 2 times
       calls puts @foo.bar 2 times
+      assumes too much for instance variable @foo
```
troessner: Same here, this feature shouldn't change.

features/reports/json.feature
```diff
@@ -39,6 +39,15 @@ Feature: Report smells using simple JSON layout
...
           },
           {
+              "context": "Smelly",
+              "lines": [2],
+              "message": "assumes too much for instance variable @foo",
+              "smell_type": "InstanceVariableAssumption",
+              "source": "smelly.rb",
+              "assumptions": ["@foo"],
+              "wiki_link": "https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md"
+          },
```
troessner: This should also not change.

features/samples.feature
```diff
@@ -32,6 +32,8 @@ Feature: Basic smell detection
       DuplicateMethodCall: Inline::C#initialize calls stack.empty? 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
       DuplicateMethodCall: Module#inline calls Inline.const_get(lang) 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
       DuplicateMethodCall: Module#inline calls options[:testing] 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
+      InstanceVariableAssumption: Inline::C assumes too much for instance variable @types, @module_name, @so_name [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
+      InstanceVariableAssumption: Module assumes too much for instance variable @options [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
```
troessner: Now those changes are fine with me - that's exactly why we're having inline.rb and friends.

features/step_definitions/sample_file_steps.rb
```diff
@@ -53,6 +53,10 @@ def a
   contents = <<-EOS.strip_heredoc
     # clean class for testing purposes
     class Clean
+      def initialize(sub)
+        @sub = sub
+      end
+
```
troessner: Wait, couldn't we rather improve the code in our "clean class" to not have instance variables at all?
Why not just use "samples/clean.rb" here?

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,84 @@
...
+          lines: [ctx.exp.line],
+          message: message,
+          parameters: { assumptions: assumptions })]
+      end
+
+      #
+      # The InstanceVariableAssumptionsExtractor is the class responsible for
+      # extracting the assumptions from a method expression
+      #
+      class InstanceVariableAssumptionsExtractor
```
troessner: This feels a little overengineered here. I mean, the way it is right now the smell detector itself is not much but an empty shell not doing much. Do we really need this class? How about dissolving it into the smell detector?

troessner:
I did a first review. I think we need to clean up this PR in terms of feature changes. It shouldn't change much if anything at all in terms of unrelated cucumber features. I also left some other comments that I'd love to see addressed. Let's work on this first and then I'm happy to give this another review :)

features/step_definitions/sample_file_steps.rb
waldyr: Should I move the path's constants definition to a different file and require them in the `spec_helper` and also the `sample_file_steps`?

troessner: I think its ok to just reference the existing files in our sample steps. Its not like our specs are going anywhere :)

CONTRIBUTING.md
waldyr: This was based on this comment https://github.com/troessner/reek/pull/959#issuecomment-231314306.

Could you please come up with a better wording?

lib/reek/smells/instance_variable_assumption.rb
waldyr: The solution before this class reeked of many smells. `ctx` causing Feature Envy or `sniff` being a big method are the ones that I can recall. But I can move the logic to the `sniff` method and we improve it step by step. Wdyt?

mvz: Yes, I think it's better to go for the simple, if smelly, solution for now. We need to rethink the structure of the smell detectors anyway. Just add a good dose of smell suppression comments :smile:.

troessner: Ok, then let's keep it like that.
But we really should streamline our smell detectors **before** writing new ones.

mvz: Uh .. I was suggesting changing it to something simpler for now.

troessner: Ok, I'm fine with that as well 😆

CONTRIBUTING.md
troessner: Hmm, how about:

>>
You can then `cd` into this directory and run Reek the same way the cucumber scenario actually ran it. This way you can debug scenario failures that can be very opaque sometimes.
 Then you can run the scenario's Reek incantation there and have a better debugging.

waldyr:
@troessner @mvz Wdyt?

mvz:
Having a final look right now.

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,71 @@
...
+
+        return [] if assumptions.empty?
+
+        build_smell_warning(ctx, assumptions)
+      end
+
+      private
+
+      def build_smell_warning(ctx, assumptions)
+        message = "assumes too much for instance variable #{assumptions.join(', ')}"
```
mvz: I thing we should report each ivar separately.

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+          [@a, @b]
+        end
+      end
+    EOS
+
+    expect(src).to reek_of(:InstanceVariableAssumption)
+  end
+
+  context 'inner classes' do
+    it 'should not report inner class' do
```
mvz: I don't think this negated test is relevant.

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+      EOS
+
+      expect(src).not_to reek_of(:InstanceVariableAssumption)
+    end
+
+    it 'should report outter class' do
+      src = <<-EOS
+        class Dummy
+          def test
+            @a = 1
```
mvz: This code doesn't really assume anything about `@a`, so it should **not** reek IMHO.

spec/reek/smells/instance_variable_assumption_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+        def test
+          @a
+        end
+      end
+    EOS
+
+    expect(src).to reek_of(:InstanceVariableAssumption)
+  end
+
+  it 'should report any initialization outside initialize' do
```
mvz: Maybe 'reports variable even if others are initialized' describes better what is being tested here?

lib/reek/smells/instance_variable_assumption.rb
```diff
@@ -0,0 +1,71 @@
...
+          method_exp.name == :initialize
+        end
+
+        return [] unless initialize_exp
+
+        initialize_exp.each_node(:ivasgn).map(&:name)
+      end
+
+      def variables_from_context(instance_methods)
+        instance_methods.map do |mexp|
```
troessner: "mexp"? "method expression"? How about just "method"?

features/configuration_files/directory_specific_directives.feature
```diff
@@ -163,11 +166,13 @@ Feature: Directory directives
...
-    other/projects_controller.rb -- 1 warning:
+    other/projects_controller.rb -- 2 warnings:
+      [1]:InstanceVariableAssumption: ProjectController assumes too much for instance variable @project [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
       [4]:NestedIterators: ProjectController#show contains iterators nested 2 deep [https://github.com/troessner/reek/blob/master/docs/Nested-Iterators.md]
-    web_app/app/controllers/users_controller.rb -- 1 warning:
+    web_app/app/controllers/users_controller.rb -- 2 warnings:
+      [1]:InstanceVariableAssumption: UsersController assumes too much for instance variable @user [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
       [1]:IrresponsibleModule: UsersController has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]
-    2 total warnings
+    4 total warnings
```
troessner: Can we rather update the code samples for this features so that it doesn't smell of this, meaning this feature and stdin.feature dont show up at all in this PR?

mvz: Let's finalize the implementation first, then update this.

spec/reek/smells/instance_variable_assumption_spec.rb
waldyr: Makes sense. I'll change it.

lib/reek/context/code_context.rb
```diff
@@ -15,6 +15,7 @@ module Context
     #
     # :reek:TooManyMethods: { max_methods: 19 }
     # :reek:TooManyInstanceVariables: { max_instance_variables: 8 }
+    # :reek:InstanceVariableAssumption
```
mvz: This one can go now \o/.

lib/reek/context/ghost_context.rb
```diff
@@ -8,6 +8,8 @@ module Context
     # context tree. This context will not be part of the resulting tree, but
     # will track context and visibility separately while building is in
     # progress.
+    #
+    # :reek:InstanceVariableAssumption
```
mvz: And so can this one :tada:.

lib/reek/spec/should_reek_of.rb
```diff
@@ -8,6 +8,7 @@ module Spec
     # An rspec matcher that matches when the +actual+ has the specified
     # code smell.
     #
+    # :reek:InstanceVariableAssumption
```
mvz: And this one :balloon:.

features/configuration_files/directory_specific_directives.feature
mvz: Oh, I see the the code for this is actually in the feature file, so it will show up either way. Looking at it, I *think* we may want to suggest disabling InstanceVariableAssumption for Rails controllers.

troessner: Which @waldyr did in this PR as well by updating our [README](https://github.com/troessner/reek/pull/1017/files#diff-04c6e90faac2675aa89e2176d2eec7d8R513) ;)

waldyr:
@troessner @mvz updated!

troessner:
Awesome job @waldyr !

Would you like to have to honor of releasing (minor version)?
(create a PR like [this](https://github.com/troessner/reek/pull/1034) and after that is merge pull on master followed by `rake release`).

waldyr:
Sure things 👍

prpetten:
I saw @mvz comment about disabling this for Rails Controllers.  I was wondering if there was a pattern you might suggest we use within the context of Rails Controllers that would 'deodorize' this smell, rather than disabling it for the directory.

troessner:
I'm not sure I understand what you mean with 'deodorize' in this context 😆 
If you want to be more specific than the shotgun approach you can do something like this:

```Yaml
InstanceVariableAssumption:
  exclude:
    - "UsersController#index"
    - "OtherController#show"
````

Or you could use our [configuration comments](https://github.com/troessner/reek#source-code-comments).
Is this what you had in mind?

waldyr:
If 'deodorize' means make the smell go away then you can't use variables that are defined in other places like before_filters or concerns. In other words: `Use the controller solely for instantiating variables`.

prpetten:
@waldyr Ah, I didn't realize it was the before_filter that was triggering it.  I like the DRY-ness of the before filter to set the instance variables across actions, so I'll likely keep it, and then just add the recommended exclusion.

@troessner Thanks for the note on the exclude, I didn't realize that you could exclude individual methods within the controller, that's actually really cool.

PR #1022, opened 2016-08-16
Update yard dependency

mvz:
Nothing's changed, really.


troessner:
Still amazing to me that there is no proper service that can issue those PRs for us....

PR #1028, opened 2016-08-22
Add ManualDispatch smell

### LINKED ISSUE #673: Detect more type checking
We have NilCheck, but we should detect more type checking, e.g.:

```ruby
if foo.respond_to? :bar
   foo.bar
end
```

```ruby
case foo
when Bar
  foo.bar
else
  puts 'no bar!'
end
```

### LINKED ISSUE #1031: Report ManualDispatch smell for code that uses `respond_to?`
Extracted from #673:

Example of bad code:

```ruby
if foo.respond_to? :bar
   foo.bar
end
```

Part of the "simulated polymorphism" family

backus:
Aims to fix #1031


backus:
@waldyr do you know why running cucumber for `features/samples.feature:12` results in zero smells being found for only `samples/redcloth.rb`?

waldyr:
Unfortunately I don't. That output is really messed up though it ignored the whole redcloth file.

backus:
@waldyr ahh running it manually shows that it outputs an exception to stderr which is not caught be cucumber

mvz:
@backus that is our new exception catcher kicking in: If Reek encounters an internal error when processing a file, instead of aborting, it now prints a message and goes on to the next file.

backus:
@mvz thanks. Could you clarify how complex you think this smell detector should be? I started with something more complex that would flag something like this:

```ruby
def foo
  bar.baz if bar.respond_to?(:baz)
end
```

but not this

```ruby
def foo
  bar.something_else if bar.respond_to?(:baz)
end
```

After finishing that implementation I figured that, if I'm checking the duck type of some object, then I probably want to flag any method call on that receiver inside of a conditional. I ended up following that logic a bit further and just reduced the code to flagging any `respond_to?` usage which I think makes sense. WDYT?

features/smells/manual_dispatch.feature
```diff
@@ -0,0 +1,23 @@
+Feature: Smell - ManualDispatch
```
mvz: We're trying to reduce the number of cucumber scenario's, and I don't think a separate scenario for an individual smell is very useful. So, as far as I'm concerned, this one can go. @troessner wdyt?

backus: Oh ok. My first smell contribution so I'll just remove it

troessner: >> We're trying to reduce the number of cucumber scenario's, and I don't think a separate scenario for an individual smell is very useful.

Fully agree, let's rather remove this and focus on the specs.

troessner:
>> After finishing that implementation I figured that, if I'm checking the duck type of some object, then I probably want to flag any method call on that receiver inside of a conditional. I ended up following that logic a bit further and just reduced the code to flagging any respond_to? usage which I think makes sense. WDYT?

Yes, that makes sense for me as well :)

docs/Manual-Dispatch.md
```diff
@@ -0,0 +1,30 @@
+## Introduction
+
+Reek reports a _Manual Dispatch_ smell if it finds source code that manually checks whether an object responds to a method before that method is called.
+
```
troessner: Let's explain why this is bad as well :)

spec/reek/smells/manual_dispatch_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        class Dummy
+          def call
+            foo.respond_to?(:bar) && foo.bar
+          end
+        end
+      EOS
+
+      expect(src).to reek_of(:ManualDispatch)
+    end
+  end
```
troessner: Nice specs!

lib/reek/smells/manual_dispatch.rb
```diff
@@ -0,0 +1,32 @@
...
+      #
+      # @return [Array<SmellWarning>]
+      #
+      # :reek:FeatureEnvy
+      def sniff(ctx)
+        ctx.each_node(:send).flat_map do |node|
+          next unless node.name.equal?(:respond_to?)
+
+          smell_warning(context: ctx, lines: [node.line], message: MSG)
+        end.compact
```
troessner: Nice and easy, like it a lot.

lib/reek/smells/manual_dispatch.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Smells
+    #
+    # A Manual Dispatch occurs when a method is only called after a
+    # manual check that the method receiver is of the correct type.
+    #
+    # The +ManualDispatch+ checker reports any invocation of +respond_to?+
+    #
+    # See {file:docs/Manual-Dispatch.md} for details.
+    class ManualDispatch < SmellDetector
+      MSG = 'manually dispatches method call'.freeze
```
troessner: We have the convention in our codebase to spell things out (with the exception of "ctx" and "exp") so let's rather go with MESSAGE

backus: Will change

spec/reek/smells/manual_dispatch_spec.rb
backus: Thanks!

backus: 100% mutant covered too :)

docs/Manual-Dispatch.md
backus: Added an additional sentence

backus:
@troessner updated!

troessner:
Looks good to me, lgtm! @mvz any objections from your side?

backus:
Hm maybe this should handle cases where the condition body is an exception?  I have the following code in a private codebase

```ruby
fail InvalidReturn, outcome unless outcome.respond_to?(:success?)
```

which helps enforce a contract. Not sure I would agree with this being a smell.

troessner:
>> Not sure I would agree with this being a smell.

I agree, but that's exactly what our Reek configuration is for, isn't it? You can even exclude [whole directories](https://github.com/troessner/reek#configuration-file) so I don't think we should add complex conditions to our source but rather let the user configure it.

backus:
@troessner yup I agree with that (I figured this might just be a case for disabling). I haven't contributed to reek before so I wanted to see how fine grained you prefer the smell detectors be

docs/Manual-Dispatch.md
```diff
@@ -0,0 +1,30 @@
...
+    foo.bar if foo.respond_to?(:bar)
+  end
+end
+```
+
+Reek would emit the following warning:
+
+```
+test.rb -- 1 warning:
+  [9]: MyManualDispatcher manually dispatched method call (ManualDispatch)
```
mvz: dispatched -> dispatches

mvz:
Just one nit :smile:.

backus:
@mvz good eye! I've fixed the inconsistency

backus:
(Github sometimes closes PRs when I rebase and [push-each](https://github.com/dkubb/git-tools/blob/master/git-push-each) them. Ignore the close)

troessner:
Awesome job @backus !

Happy to merge this one! I'll also release a new minor version right away to get this bad boy on the road :)

troessner:
>>   ( $ (master)) rake release
reek 4.3.0 built to pkg/reek-4.3.0.gem.
Tagged v4.3.0.
Pushed git commits and tags.
Pushed reek 4.3.0 to rubygems.org

backus:
Thanks @troessner!

Issue #1011, opened 2016-08-06
Introduce our smell matcher to all our detector specs


troessner:
Now with #1001 merged I'd like to introduce this in all our detector specs.
I have a 7 hour bus ride coming up next Wednesday and that's when I'll start cracking on this one ;)

troessner:
Closing with #1056 merged!

PR #1032, opened 2016-08-22
Remove memoized instance variables from accounting of TooManyInstanceVariables smell

waldyr:
#807

### DIFF
--- a/lib/reek/context/code_context.rb
+++ b/lib/reek/context/code_context.rb
@@ -68,8 +68,9 @@ module Reek
       # @param type [Symbol] the type of the nodes we are looking for, e.g. :defs.
       # @yield block that is executed for every node.
       #
-      def local_nodes(type, &blk)
-        each_node(type, [:casgn, :class, :module], &blk)
+      def local_nodes(type, ignored = [], &blk)
+        ignored += [:casgn, :class, :module]
+        each_node(type, ignored, &blk)
       end
 
       # Iterate over `self` and child contexts.
--- a/lib/reek/smells/too_many_instance_variables.rb
+++ b/lib/reek/smells/too_many_instance_variables.rb
@@ -35,13 +35,28 @@ module Reek
       #
       def sniff(ctx)
         max_allowed_ivars = value(MAX_ALLOWED_IVARS_KEY, ctx)
-        count = ctx.local_nodes(:ivasgn).map { |ivasgn| ivasgn.children.first }.uniq.length
+
+        count = count_instance_variables(ctx)
+
         return [] if count <= max_allowed_ivars
+
         [smell_warning(
           context: ctx,
           lines: [ctx.exp.line],
           message: "has at least #{count} instance variables",
           parameters: { count: count })]
+      end
+
+      private
+
+      def count_instance_variables(ctx)
+        ctx.each_node(:ivasgn, ignored_nodes).map do |node|
+          node.children.first
+        end.uniq.size
+      end
+
+      def ignored_nodes
+        [:or_asgn, :class, :module]
       end
     end
   end
--- a/lib/reek/smells/too_many_instance_variables.rb
+++ b/lib/reek/smells/too_many_instance_variables.rb
@@ -35,28 +35,14 @@ module Reek
       #
       def sniff(ctx)
         max_allowed_ivars = value(MAX_ALLOWED_IVARS_KEY, ctx)
-
-        count = count_instance_variables(ctx)
-
+        variables = ctx.local_nodes(:ivasgn, [:or_asgn]).map(&:name)
+        count = variables.uniq.size
         return [] if count <= max_allowed_ivars
-
         [smell_warning(
           context: ctx,
           lines: [ctx.exp.line],
           message: "has at least #{count} instance variables",
           parameters: { count: count })]
-      end
-
-      private
-
-      def count_instance_variables(ctx)
-        ctx.each_node(:ivasgn, ignored_nodes).map do |node|
-          node.children.first
-        end.uniq.size
-      end
-
-      def ignored_nodes
-        [:or_asgn, :class, :module]
       end
     end
   end
--- a/spec/reek/smells/too_many_instance_variables_spec.rb
+++ b/spec/reek/smells/too_many_instance_variables_spec.rb
@@ -13,12 +13,10 @@ RSpec.describe Reek::Smells::TooManyInstanceVariables do
         end
       EOS
 
-      message = 'has at least 5 instance variables'
-
-      expect(src).to reek_of(:TooManyInstanceVariables,
+      expect(src).to reek_of(described_class,
                              lines: [1],
                              count: 5,
-                             message: message,
+                             message: 'has at least 5 instance variables',
                              context: 'Empty')
     end
   end
@@ -32,7 +30,7 @@ RSpec.describe Reek::Smells::TooManyInstanceVariables do
           end
         end
       EOS
-      expect(src).not_to reek_of(:TooManyInstanceVariables)
+      expect(src).not_to reek_of(described_class)
     end
 
     it 'has a configurable maximum' do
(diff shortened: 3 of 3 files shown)


features/samples.feature
```diff
@@ -133,7 +133,7 @@ Feature: Basic smell detection
       SubclassedFromCoreClass: OptionParser::CompletingHash inherits from a core class (Hash) [https://github.com/troessner/reek/blob/master/docs/Subclassed-From-Core-Class.md]
       SubclassedFromCoreClass: OptionParser::OptionMap inherits from a core class (Hash) [https://github.com/troessner/reek/blob/master/docs/Subclassed-From-Core-Class.md]
       TooManyConstants: OptionParser has 16 constants [https://github.com/troessner/reek/blob/master/docs/Too-Many-Constants.md]
-      TooManyInstanceVariables: OptionParser has at least 6 instance variables [https://github.com/troessner/reek/blob/master/docs/Too-Many-Instance-Variables.md]
+      TooManyInstanceVariables: OptionParser has at least 7 instance variables [https://github.com/troessner/reek/blob/master/docs/Too-Many-Instance-Variables.md]
```
mvz: This change is surprising ...

waldyr: After:
[:@stack, :@program_name, :@banner, :@summary_width, :@summary_indent, :@default_argv, :@optparse]

Before:
[:@stack, :@program_name, :@banner, :@summary_width, :@summary_indent, :@default_argv]

waldyr: The new code isn't ignoring whatever is inside a inner module. Should it also be ignored?

waldyr: Went ahead and didn't change any old behavior.

lib/reek/smells/too_many_instance_variables.rb
```diff
@@ -35,14 +35,29 @@ def self.default_config
       #
       def sniff(ctx)
         max_allowed_ivars = value(MAX_ALLOWED_IVARS_KEY, ctx)
-        count = ctx.local_nodes(:ivasgn).map { |ivasgn| ivasgn.children.first }.uniq.length
+
+        count = count_instance_variables(ctx)
+
```
troessner: How about not having those newlines here and above? For me they are not increasing readability but rather decreasing it.

lib/reek/smells/too_many_instance_variables.rb
```diff
@@ -35,14 +35,29 @@ def self.default_config
...
           context: ctx,
           lines: [ctx.exp.line],
           message: "has at least #{count} instance variables",
           parameters: { count: count })]
       end
+
+      private
+
+      def count_instance_variables(ctx)
+        ctx.each_node(:ivasgn, ignored_nodes).map do |node|
```
mvz: It's a pity we can't re-use `local_nodes` here, which automatically stops at inner class/module boundaries.

lib/reek/smells/too_many_instance_variables.rb
```diff
@@ -35,14 +35,29 @@ def self.default_config
...
+          node.children.first
+        end.uniq.size
+      end
+
+      def ignored_nodes
+        [:or_asgn, :class, :module]
```
troessner: Lets make this a constant.

lib/reek/smells/too_many_instance_variables.rb
troessner: I agree, right now its confusing because of that. Maybe we can revise our approach here and "make it work"?

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -1,32 +1,79 @@
...
+        end
+      EOS
+
+      message = 'has at least 5 instance variables'
 
-  def ivar_sequence(count: default_max_ivars, alphabet: ('a'..'z').to_a)
-    alphabet.first(count).map do |name|
-      "@#{name}=#{rand}"
-    end.join(',')
+      expect(src).to reek_of(:TooManyInstanceVariables, message: message)
```
troessner: I know that the "purist" TDD philosophy advises to test all relevant attributes separately, but this really creates a lot of noise in this case and seems unnecessary.
Let's rather pull all of this together into one succinct example:

```Ruby
expect(src).to reek_of(:TooManyInstanceVariables, 
  message: message,
  context: ctx, 
  ...)
```

We do this in our specs already and I intend to even double down on this in the course of #1011

lib/reek/smells/too_many_instance_variables.rb
waldyr: Wdyt in adding the ignored logic into the local_nodes method?

spec/reek/smells/too_many_instance_variables_spec.rb
waldyr: In my case it's more false-negative avoidance than a TDD thing. However I agree it definitely is unnecessary and #1011 rocks.

waldyr:
@mvz @troessner Wdyt?

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -1,32 +1,47 @@
...
+            @a = @b = @c = @d = 1
+            @e = 1
+          end
+        end
+      EOS
 
-  def too_many_ivars
-    default_max_ivars + 1
-  end
+      message = 'has at least 5 instance variables'
```
troessner: I don't think we need this one-off variable - how about just putting the literal in the expectation below?

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -1,32 +1,47 @@
...
 
-  def ivar_sequence(count: default_max_ivars, alphabet: ('a'..'z').to_a)
-    alphabet.first(count).map do |name|
-      "@#{name}=#{rand}"
-    end.join(',')
+      expect(src).to reek_of(:TooManyInstanceVariables,
+                             lines: [1],
+                             count: 5,
+                             message: message,
+                             context: 'Empty')
```
troessner: Yes, waaaayyy conciser already ;)

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -37,62 +52,106 @@ def ivars
...
+            @e = 1
           end
         end
       EOS
-      ctx = Reek::Context::CodeContext.new(nil, Reek::Source::SourceCode.from(src).syntax_tree)
-      detector.sniff(ctx).first
+      expect(src).to reek_of(:TooManyInstanceVariables)
     end
 
-    it_should_behave_like 'common fields set correctly'
```
troessner: After #1011 we hopefully can throw out this one completely :)

troessner:
Just one tiny thing and then this is ready for merge from my POV!

lib/reek/context/code_context.rb
```diff
@@ -68,8 +68,9 @@ def initialize(_parent, exp)
       # @param type [Symbol] the type of the nodes we are looking for, e.g. :defs.
       # @yield block that is executed for every node.
       #
-      def local_nodes(type, &blk)
-        each_node(type, [:casgn, :class, :module], &blk)
+      def local_nodes(type, ignored = [], &blk)
+        ignored += [:casgn, :class, :module]
```
backus: Maybe write this as `ignore = ignored + [:casgn, :class, :module]`. This method probably shouldn't mutate the array it receives

troessner: But it doesn't. The original reference will still be intact since `+=` will mutate a locally scoped array:

>>[1] pry(main)> foo = [:class]
=> [:class]
[2] pry(main)> def bar(arg); arg += [:module]; end
=> :bar
[3] pry(main)> bar foo
=> [:class, :module]
[4] pry(main)> foo
=> [:class]

backus: Huh... I didn't realize it worked like that. Sorry for the noise :/

troessner: Absolutely no need to apologise, learning and discussing is what pull requests are for :)

waldyr:
@troessner @mvz updated!

troessner:
You really are on a roll :)

Merged! Let's put that in our next release as well!

PR #1035, opened 2016-08-24
User our Spec matcher everywhere.

### LINKED ISSUE #1011: Introduce our smell matcher to all our detector specs

troessner:
** Update **

Ok, so this got WAY out of hand. The actual title of this PR should now rather be "Refactor and simplify all our smell detector specs and make them consistent."

I'll create a separate PR out of this once I finished up everything with a better branch name and commit message.

Fixes #1011 

What I did:

* Killed all the shared examples
* Introduced our Spec matcher everywhere. This and the bullet point above also have the nice side effect that now all requires at the top of a smell detector spec file look like this:

```Ruby
require_lib 'reek/smells/irresponsible_module'
````

And that's it. No more references to `Examiner`, `CodeContext` and god knows what else.

* Removed silly specs. Some areas of Reek are shockingly under-speced whilst others are over-speced. If we have a threshold of 4 x it is just silly to have a spec that 3 x does not smell. And 2 x does not smell. And 1 x does not smell. That's just bloated code.

* Removed redundant specs. Bigger smell detector specs were full of duplication.

* Removed foo / bar because I honestly can't see that anymore and replaced it with "alpha/bravo/charlie" as discussed.

* Made our syntax consistent. Introduced heredocs everywhere and removed multi line strings. Introduced newlines. And so on.

* Simplified a lot, a lot of our examples. For instance:

```Ruby
# Old
src = 'def simple(arg) args = {}; if arg then args.merge(\'a\' => \'A\') end end'
# New
src = 'def alpha(bravo); if bravo then charlie end; end'
```

The meaning stays the same but the example is less confusing and better readable.

* Introduced the same general examples for every spec at the beginning. Every smell detector spec now starts out like this:

```Ruby
RSpec.describe Reek::Smells::IrresponsibleModule do
  it 'reports the right values' do
    src = <<-EOS
      class Dummy
      end
    EOS

                           lines:   [1],
                           context: 'Dummy',
                           source:  'string')
  end

  it 'does count all occurences' do
    src = <<-EOS
      class Outer
        # Method is necessary because we don't count empty classes.
        def something; end
        class Inner
        end
      end

    EOS

                               lines:   [1],
                               context: 'Outer')
                           lines:   [4],
                           context: 'Outer::Inner')
  end
end
```

The 2nd spec 'does count all occurences' is only added when it actually makes sense, so there are some detectors that only start with the first spec above.

* Removed contexts that didn't make any sense. E.g. I removed non-nested contexts with only example and just updated the example description. Same goes for specs that were covered by one giant context.


troessner:
I'd like to discuss several proposal before going on further:

* Get rid of the damn shared examples. I believe Rspec's shared examples are bad in any case, not just in Reek. I don't feel they offer any value at all (if you need to share behaviour, rather use a module and include it in your spec_helper) but do a lot of damage since the additional abstraction and indirection is confusing and tedious. 
I'd love to drop them completely throughout our spec base, but first of all in our smell detectors. 
I started in this PR by dropping 

```Ruby
it_should_behave_like 'SmellDetector'
```

* I'd like to introduce one example at the beginning of each smell detector that basically looks like this:

```Ruby
  it 'reports the right values' do
    src = <<-EOS
    class Klass
      attr_writer :my_attr
    end
    EOS

    expect(src).to reek_of(:Attribute,
                           lines: [2],
                           context: 'Klass#my_attr',
                           message: 'is a writable attribute')
  end
```

And yes, I'd repeat this pattern in every smell detector spec. I take this little amount of duplication (it is little since the code in question and the reported attributes will be different in every detector) over shared examples any day.

WDYT?

troessner:
Ok, so I refactored the first 3 smell detector specs. I'll stop right here so we can discuss it first. If everybody is on board with the changes I'll continue to go down that road ;)

waldyr:
The message could eventually trigger Rubocop's long line, couldn't it?

troessner:
>> Isn't described_class better than writing the smell name?

I'm not sure I follow - I'm using `described_class` everywhere with one exception [here](https://github.com/troessner/reek/pull/1035/files#diff-88b49d22aa7fd6d7107689d24008e8fcR14) that I missed and will fix later.

waldyr:
I was looking at the example at that time.

mvz:
I agree completely with your plan, @troessner.

troessner:
>> I was looking at the example at that time.

Ah, ok 😆 
I'll fix that :)

troessner:
Update of what I've done so far:

* Killed all the shared examples
* Introduced our Spec matcher everywhere. This and the bullet point above also have the nice side effect that now all requires at the top of a smell detector spec file look like this:

```Ruby
require_lib 'reek/smells/irresponsible_module'
````

And that's it. No more references to `Examiner`, `CodeContext` and god knows what else.

* Removed silly specs. Some areas of Reek are shockingly under-speced whilst others are over-speced. If we have a threshold of 4 x it is just silly to have a spec that 3 x does not smell. And 2 x does not smell. And 1 x does not smell. That's just bloated code.

* Made our syntax consistent. Removed foo / bar because I honestly can't see that anymore. Introduced heredocs everywhere and removed multi line strings. Introduced newlines. And so on.

* Introduced the same general examples for every spec at the beginning. Every smell detector spec now starts out like this:

```Ruby
RSpec.describe Reek::Smells::IrresponsibleModule do
  it 'reports the right values' do
    src = <<-EOS
      class Dummy
      end
    EOS

                           lines:   [1],
                           context: 'Dummy',
                           source:  'string')
  end

  it 'does count all occurences' do
    src = <<-EOS
      class Outer
        # Method is necessary because we don't count empty classes.
        def something; end
        class Inner
        end
      end

    EOS

                               lines:   [1],
                               context: 'Outer')
                           lines:   [4],
                           context: 'Outer::Inner')
  end
end
```

Only a dozen smell detectors left to go now :)

backus:
Have you considered adding a support method like this

```ruby
module ReekDescription
  def reek(options)
    reek_of(described_class, options
  end
end
```

and mixing it into all of your smell detect specs so you could just write `expect(src).to reek(lines: [4], ...)`

troessner:
>> and mixing it into all of your smell detect specs so you could just write expect(src).to reek(lines: [4], ...)?

Yes, I have. But then I would - again - introduce a layer of indirection to something that is now very readable. I believe it is fine to have this duplication **especially** given that our syntax here is very unlikely to change (and if it would, it would be a major upgrade).

mvz:
> Removed foo / bar because I honestly can't see that anymore.

What have you got against foo, bar, etc.? I rather like them because they stop you from trying to find a hidden meaning, and picking names becomes a lot easier.

We can pick some other list of course: https://en.wikipedia.org/wiki/Metasyntactic_variable

troessner:
>> What have you got against foo, bar, etc.? I rather like them because they stop you from trying to find a hidden meaning, and picking names becomes a lot easier.

I'm not trying to change this everywhere, so I don't have an agenda.
But when I have the option of doing

```Ruby
def foo(bar); end
```

or this

```Ruby
def x(y); end
```

or 

```Ruby
def a(b); end
```

I'd always go with the last 2 because it's conciser.

spec/reek/smells/control_parameter_spec.rb
```diff
@@ -4,34 +4,36 @@
...
+        puts 'c' if argc
+        puts 'd' if argd
       end
     EOS
 
     expect(src).to reek_of(described_class, lines: [2], argument: 'arga')
-    expect(src).to reek_of(described_class, lines: [6], argument: 'argb')
+    expect(src).to reek_of(described_class, lines: [3], argument: 'argb')
+    expect(src).to reek_of(described_class, lines: [7], argument: 'argc')
+    expect(src).to reek_of(described_class, lines: [8], argument: 'argd')
```
mvz: Isn't two occurrences enough for this test?

spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -43,7 +75,7 @@ def chained
         @sig.keys.sort_by { |xray| xray.to_s }.each { |min| md5 << min.to_s }
       end
     EOS
-    expect(src).not_to reek_of(:NestedIterators)
+    expect(src).not_to reek_of(described_class)
```
mvz: I think the examples are easier to read when using the symbol.

spec/reek/smells/boolean_parameter_spec.rb
```diff
@@ -1,91 +1,103 @@
...
+                           context: 'm1',
+                           parameter: 'b')
+    expect(src).to reek_of(described_class,
+                           lines:   [4],
+                           context: 'm2',
+                           parameter: 'x')
+    expect(src).to reek_of(described_class,
+                           lines:   [4],
+                           context: 'm2',
+                           parameter: 'y')
```
mvz: I think using very short names for things makes it harder to check that these four expectations are actually correct.

spec/reek/smells/control_parameter_spec.rb
```diff
@@ -1,103 +1,132 @@
...
 
-  it_should_behave_like 'SmellDetector'
+  it 'does count all occurences' do
+    src = <<-EOS
+      def m(arga, argb)
+        arga ? @one : 5
+        argb ? @two : 7
+      end
+
+      def nested(argc, argd)
```
mvz: I'm not sure why this method is named `nested` anymore?

spec/reek/smells/data_clump_spec.rb
```diff
@@ -1,178 +1,126 @@
...
-        def c_raw_singleton (src, options) end
-      end
-    EOS
-    expect(src).to reek_of(:DataClump, count: 5)
-  end
+    it 'does not care about the order of arguments' do
+      src = <<-EOS
+        #{scope} Dummy
+          def x(y1,y2); end
+          def y(y2,y1); end # <- This is the swapped one!
```
mvz: Again, this is easier to spot when using readable names I think.

mvz:
Overall, I think this is going in the right direction and really makes our tests cleaner and more consistent :+1:. I think we do need to have a discussion about:
* Names used in the code samples.
* Using `described_class` versus `:SmellSymbol` in the `reek_of` matcher.

spec/reek/smells/control_parameter_spec.rb
troessner: Good point, I removed the last 2.

spec/reek/smells/nested_iterators_spec.rb
troessner: Hmm, I kind of agree but given that we use this consistently everywhere it still should be obvious, don't you think?
Overall I'm not hung up on that and I can switch it back if you feel that this decreases readability signficantly.

spec/reek/smells/boolean_parameter_spec.rb
troessner: I agree. But, and that is a big but, those are the core specs. They are never intended to change and if it is, it's a really nasty bug.
I'd prefer x over foo and y over bar anytime. Also, is foo / bar more readable? I doubt it.
Or did you have other names in mind?

spec/reek/smells/control_parameter_spec.rb
troessner: Removed it completely due to your other comment above.

troessner:
Only 9 more specs to go, wooohooo!

troessner:
Ready for review! See the initial PR description above for a summary of what I did.
The diff is sometimes very badly readable, apologies for that!
In hindsight it would have been smarter to split this up into multiple PR's, one just dealing with indentation, one with renames, one with spec removals and so on.
Hope you can still review it ;)

mvz:
> I agree. But, and that is a big but, those are the core specs. They are never intended to change and if it is, it's a really nasty bug.

The intent may not change, but it looks to me they are actually changed regularly. Also, I think they should be very readable, because they reveal the intent of our smell detectors.

> I'd prefer x over foo and y over bar anytime. Also, is foo / bar more readable? I doubt it.
Or did you have other names in mind?

It seems my brain processes the names better if they're easily pronounceable and maximally different. When it encounters stuff like `m1`, it translates it into 'yada yada yada'. Other people's brains may function differently. I may also have been trained to detect uncommunicative *whatever* name automatically, which adds to the noise when reading :smile:.

In conclusion: For one-line examples, `x` and `y` are fine, but not `x1`, `x2`. For longer examples, I prefer something more pronounceable and readily distinguished. Could be from the foo/bar group, could be something else (the Japanese use 'hoge').

troessner:
Actually it's kind of funny that we never had this discussion. I mean, if you think about, in "normal" code bases you don't have this problem because most of the time whatever you test is coming from the domain you're covering. And that has proper names.

Reek is the opposite, at least when it comes what we give as input to smell detectors: We almost never have good names because we they are not only unnecessary but they would also be misleading by implying that they would be relevant.

What I'm going for here is that I believe that it actually is very important that we have one consistent, unambiguous standard for how we "name things, that should not have a good name".
Right now this is a complete chaos in most of our specs and it decreases readability and introduces  additional brain work all the time (brain time I'd rather allocate on other problems).

So how about we settle for a standard now?

"foo" / "bar" / "baz" is not good enough for me because quite often we need more than 3 names.

How about just using the military alphabet? alpha, bravo, charlie, delta, echo, foxtrot.....
And then just in descending order like this:

```Ruby
class Alpha
  def bravo(charlie, delta, echo)
    charlie.foxtrot
  end
end
```

That way the name would also give you a clue where the thing you're talking about is. E.g. alpha is always at the beginning, foxtrot probably somewhere at the end.

WDYT?

mvz:
> So how about we settle for a standard now?

Yes!

> "foo" / "bar" / "baz" is not good enough for me because quite often we need more than 3 names.

Well, the list goes on (qux, quuz, ..), but the clarity suffers.

> How about just using the military alphabet? alpha, bravo, charlie, delta, echo, foxtrot.....

That would be fine and gives us a nice large set to work with.

spec/reek/smells/nested_iterators_spec.rb
mvz: I have a slight preference for using the symbol, because it means you need a little less context to understand the spec.

Also, I prefer the idea of code reeking of a smell over it reeking of a detector class. OTOH, I wouldn't mind specs written like:

```
let(:detector) { described_class.new }

it 'something' do
  expect(detector).to detect_smell_in(source)
end
```

Just my 2 cents of the evening.

mvz:
@waldyr and @chastell, do you have any input for the discussions @troessner and I are having in this PR?

troessner:
@mvz given there is no further input: I'll update the naming everywhere, update our contributing section (!!!) and then lgtm. Deal?

mvz:
@troessner deal!

spec/reek/smells/nested_iterators_spec.rb
chastell: I also have a slight preference for the explicit `expect(src).not_to reek_of(:NestedIterators)` – especially in this particular example the _very_ English-like flow of RSpec has the most value, you just read the sentence and don’t have to do the (granted, tiny) mental jump of dereferencing `described_class`. But it’s a slight preference and I’ll be fine with `described_class` everywhere.

chastell:
Thanks for pinging me – this summer for me is streaks of holidays (with a toddler…) interspersed with streaks of very intensive work; thanks for keeping me on the team. :)

I have a slight preference for explicit smell names (over `described_class`), because they read a tiny bit better (the brain does not have to do the dereferencing), which is a win over a more _proper_ constant extraction. But I’m OK with either.

As for names, I’d love to have (a) consistency and (b) something better than both `x`/`y` and `foo`/`bar` – and I really like the `alpha`/`bravo`/`charlie` scheme for some reason. ;)

Master of Orion espionage

troessner:
>> 
I have a slight preference for explicit smell names (over described_class), because they read a tiny bit better (the brain does not have to do the dereferencing), which is a win over a more proper constant extraction. But I’m OK with either.

@mvz and @chastell you both are in favour of not using described_class -> I'll change it :)

troessner:
Closed in favor of #1056

PR #1038, opened 2016-08-24
Bump version to 4.4.0

waldyr:


waldyr:
@troessner Is that right? Now merge and `rake release`?

troessner:
>> @troessner Is that right? Now merge and rake release?

Just merged it. Now you just to do 

>>
git co master
git pl
rake release

(I could do that myself but you're on the team so you should get into the habit of doing this as well)

waldyr:
I just wanted to make sure I was doing the right thing. 

I did `rake release` and released it. Thank you for the honor 😄

PR #1039, opened 2016-08-25
Do not hide exceptions in reek_of matcher

### LINKED ISSUE #1027: Handle or fail tests when smell detectors return nil
When `#sniff` returns `nil` instead of `[]`, our tests still pass but we output our please-report-an-exception text. We should either allow `nil` output or fail tests in this case.

I found this with mutant when implementing #1026.

mvz:
Fixes #1027.

An alternative is to make Examiner not catch exceptions, but introduce a SafeExaminer that does and use that from the CLI.

### DIFF
--- a/lib/reek/spec.rb
+++ b/lib/reek/spec.rb
@@ -38,8 +38,8 @@ module Reek
   # To check for specific smells, use something like this:
   #
   #   ruby = 'def double_thing() @other.thing.foo + @other.thing.foo end'
-  #   ruby.should reek_of(:Duplication, /@other.thing[^\.]/)
-  #   ruby.should reek_of(:Duplication, /@other.thing.foo/)
+  #   ruby.should reek_of(:DuplicateMethodCall, name: '@other.thing')
+  #   ruby.should reek_of(:DuplicateMethodCall, name: '@other.thing.foo', count: 2)
   #   ruby.should_not reek_of(:FeatureEnvy)
   #
   # @public
--- a/lib/reek/spec/should_reek_of.rb
+++ b/lib/reek/spec/should_reek_of.rb
@@ -9,6 +9,13 @@ module Reek
     # code smell.
     #
     class ShouldReekOf
+      # Variant of Examiner that doesn't swallow exceptions
+      class UnsafeExaminer < Examiner
+        def run
+          examine_tree
+        end
+      end
+
       attr_reader :failure_message, :failure_message_when_negated
 
       def initialize(smell_type_or_class,
@@ -16,13 +23,15 @@ module Reek
                      configuration = Configuration::AppConfiguration.default)
         @smell_type = normalize smell_type_or_class
         @smell_details = smell_details
+        configuration.load_values(smell_type => { Smells::SmellConfiguration::ENABLED_KEY => true })
         @configuration = configuration
-        @configuration.load_values(smell_type => { Smells::SmellConfiguration::ENABLED_KEY => true })
       end
 
       def matches?(source)
         @matching_smell_types = nil
-        self.examiner = Examiner.new(source, configuration: configuration)
+        self.examiner = UnsafeExaminer.new(source,
+                                           filter_by_smells: [smell_type],
+                                           configuration: configuration)
         set_failure_messages
         matching_smell_details?
       end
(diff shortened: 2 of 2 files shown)


mvz:
I'm not entirely happy with this implementation: We make Examiner and SmellRepository jump through hoops to match only one smell, plus the replacement of Examiner with UnsafeExaminer is hard to test.

I'm going to try some refactoring.

troessner:
Great, then I can hold off with my review ;)

mvz:
Dammit, I can't make the change easy. There's just too much coming together in Examiner.

mvz:
@troessner please review as-is. Also, wdyt about the other option? A third option is moving the exception handling up into the Reporter class.

troessner:
Will review both this one and #1041 tomorrow :)

troessner:
"Was du heute kannst besorgen, das verschiebe nicht auf morgen" as we say in Germany.

Which roughly translates to "F*ck it, I'm reviewing now"

lib/reek/smells/smell_repository.rb
```diff
@@ -25,6 +25,7 @@ def self.smell_types
       # @return [Array<Reek::Smells::SmellDetector>] All SmellDetectors that we want to filter for
       #         e.g. [Reek::Smells::Attribute].
       def self.eligible_smell_types(filter_by_smells = [])
+        filter_by_smells = Array(filter_by_smells)
```
troessner: Hmmm. Hmmm. Well. It seems like you're trying to make up for Ruby's lack of static typing.
If you're doing this here, shouldn't we then consequently update ALL methods in Reek?
This change seems arbitrary - Why do we do this?

lib/reek/spec.rb
```diff
@@ -38,8 +38,8 @@ module Reek
   # To check for specific smells, use something like this:
   #
   #   ruby = 'def double_thing() @other.thing.foo + @other.thing.foo end'
-  #   ruby.should reek_of(:Duplication, /@other.thing[^\.]/)
-  #   ruby.should reek_of(:Duplication, /@other.thing.foo/)
+  #   ruby.should reek_of(:DuplicateMethodCall, name: '@other.thing')
+  #   ruby.should reek_of(:DuplicateMethodCall, name: '@other.thing.foo', count: 2)
```
troessner: Good catch!

lib/reek/spec/should_reek_of.rb
```diff
@@ -9,20 +9,30 @@ module Spec
     # code smell.
     #
     class ShouldReekOf
+      # Variant of Examiner that doesn't swallow exceptions
+      class UnsafeExaminer < Examiner
+        def run
+          examine source.syntax_tree
+          smell_repository.report_on(collector)
```
troessner: This will break with #1041 merged - we probably should get #1041 merged first and then update this one.

troessner:
Reviewed, looks good overall!

lib/reek/smells/smell_repository.rb
mvz: I do this to ensure we have an array, so the `#include?` later on uses the Array version, rather than the String version. This stops mutant from suggesting removal of quite a bit of code elsewhere. However, this is a bit of a leftover from an earlier direction I took in implementing this, and it can go now.

lib/reek/spec/should_reek_of.rb
mvz: Yeah either way I will need to rewrite this bit in one of the PRs.

lib/reek/smells/smell_repository.rb
troessner: >> I do this to ensure we have an array, so the #include? later on uses the Array version, rather than the String version. 

Ok, maybe I should phrased that better :)
My question was of provocative nature. I understand why you're doing this but it feels erratic. Here we would do this but there are another 100 places in our code where we're not doing this. 
I'd prefer to remove it.

mvz: Yes, I'll remove it. It is unfortunate that Ruby's String and Array methods often have the same names.

mvz:
Rebased and updated.

Issue #1040, opened 2016-08-25
Kill the `examine` / `report_on` architecture.


mvz:
The nested calling of `examine` and `report_on` is making it very hard to tease apart the Examiner / SmellRepository / WarningCollector classes. Also, it means we're continually updating the state of some collector class.

I'd like to go from this to a more functional style, where `#examine` just returns a collection of smells.

As part of this, I'd like to also clean up the division of labour between `Examiner#examine` and `SmellRepository#examine`: For some reason they interface by passing individual context objects, making at lease `SmellRepository#examine` very hard to re-use.

mvz:
This will probably be a breaking change :smile:.

troessner:
Yes, I couldn't agree. The current state of this is really bad. Quite some methods we only call for sideeffects and / or they can update an object that is buried down deep deep deep in some other objects. 
👍 for a more functional style.

troessner:
Closing this one with #1041 merged!

PR #1041, opened 2016-08-25
Refactor examiner

mvz:
Implements #1040. i don't *think* I broke any APIs, so this PR goes to master :tada:.


lib/reek/examiner.rb
```diff
@@ -95,28 +97,24 @@ def smelly?
...
-          smell_repository.report_on(collector)
-        end
-        self
+      syntax_tree = source.syntax_tree
+      return [] unless syntax_tree
+      begin
+        examine syntax_tree
+      rescue StandardError => exception
+        $stderr.puts format(INCOMPREHENSIBLE_SOURCE_TEMPLATE, source.origin, exception.inspect)
+        []
```
troessner: Yep, way nicer already.

lib/reek/examiner.rb
```diff
@@ -95,28 +97,24 @@ def smelly?
...
       end
     end
 
     def examine(syntax_tree)
+      result = []
       ContextBuilder.new(syntax_tree).context_tree.each do |element|
```
troessner: How about writing this even conciser without the ugly initialization and explicit return below via:

```Ruby
def examine(syntax_tree)
  ContextBuilder.new(syntax_tree).context_tree.each_with_object([]) do |element, result|
  result += smell_repository.examine(element)
end
```

?

lib/reek/examiner.rb
```diff
@@ -44,28 +42,32 @@ def initialize(source,
...
     end
 
     #
     # @return [Array<SmellWarning>] the smells found in the source
     #
     # @public
     def smells
-      run
-      @smells ||= collector.warnings
+      @smells ||= run.sort.uniq
```
troessner: YES.

lib/reek/examiner.rb
```diff
@@ -44,28 +42,32 @@ def initialize(source,
                    configuration: Configuration::AppConfiguration.default,
                    smell_repository_class: Smells::SmellRepository)
       @source           = Source::SourceCode.from(source)
-      @collector        = CLI::WarningCollector.new
```
troessner: I quickly checked and this seems to be the only place where we use WarningCollector, this means the whole class can be removed as well.

troessner:
Sooooooooo awesome! This makes Reek a lot simpler! 2 suggestions and then lgtm!

lib/reek/examiner.rb
mvz: Done.

lib/reek/examiner.rb
```diff
@@ -86,35 +87,31 @@ def smelly?
...
     end
 
-    def examine(syntax_tree)
-      ContextBuilder.new(syntax_tree).context_tree.each do |element|
+    def syntax_tree
+      @syntax_tree ||= source.syntax_tree
+    end
+
+    def examine_tree
+      ContextBuilder.new(syntax_tree).context_tree.each.flat_map do |element|
```
mvz: This looks a little odd, but because `context_tree` is not an `Enumerable`, I cannot call `#flat_map` on it directly. I'm toyed with the idea of including `Enumerable` in `CodeContext`, but I'm not sure it's a good idea since it is so much more than a container.

troessner: But it is **also** a container. I'd vote for including Enumerable ;)

lib/reek/examiner.rb
```diff
@@ -86,35 +87,31 @@ def smelly?
...
-    # Runs the Examiner on the given source to scan for code smells
-    # and returns the corresponding Examiner instance.
+    # Runs the Examiner on the given source to scan for code smells and returns
+    # the resulting smells.
     #
     # In case one of the smell detectors raises an exception we probably hit a Reek bug.
     # So we catch the exception here, let the user know something went wrong
     # and continue with the analysis.
     #
-    # @return an instance of Examiner
```
troessner: But let's document the new return type here.

troessner:
One tiny thing + squash and then lgtm!

mvz:
Updated and squashed.

troessner:
🎉

PR #1037, opened 2016-08-24
[WIP / DEMO] Add rubocop-rspec

backus:
I'm opening this in response to #1036 to demonstrate what rubocop-rspec flags and what it can help with

Issue #1036, opened 2016-08-24
Introduce rubocop-rspec


troessner:
@backus nudged me regarding [rubocop-rspec](https://github.com/nevir/rubocop-rspec) - looks interesting and relevant for us. Let's give it a shot - wdyt?

backus:
Happy to help with the decision and with introducing the gem. Just like rubocop there is a lot of configuration options and features that can be overwhelming at first but over time help you strongly calibrate your preferences. I'm also looking for feedback on how the gem can be improved further (I've been acting as the primary maintainer for the last month or so).

backus:
See the build for https://github.com/troessner/reek/pull/1037 to see what is flagged. Most obvious cop I expect you would disable is the `ExpectActual` cop since right now your rspec matcher basically makes it impossible to avoid this. Basically this cop says that you should not say `expect([LITERAL]).to ...` instead you should have `expect([thing under test]).to ...` but the `reek_of` matcher forces a literal into `expect(...)`

mvz:
@backus thanks for the demo. I can see how this will be very useful in making our specs consistent. One thing I see is that there are several cops that mention some exceeded maximum, without mentioning what the value actually is ('Maximum example group nesting exceeded', and 'Too many expectations.'). It would be useful to have the values shown, as 'Example has too many lines. [10/5]' does.

We could work around the ExpectActual issue by putting the value in a variable, but that seems like cheating. Perhaps this cop is too simplistic, since it would also flag stuff like `expect(3).to be_odd`, which seems like a sensible way to express things, and it should be limited to well-known comparison matchers, like `expect('foo').to eq thing`?

backus:
@mvz I agree with the maximums I'm going to add an issue for that. In the meantime if you do `--auto-gen-config` it will show you what your maximum is and if you take a peek at `config/default.yml` in rubocop-rspec you can see what the default is for the gem itself

For `ExpectActual` unless you are writing specs for ruby itself (or maybe monkey patches) I'm not sure if that is actually a use case I'm worried about and I believe that cop is derived directly from an rspec guideline.

mvz:
> For `ExpectActual` unless you are writing specs for ruby itself (or maybe monkey patches) I'm not sure if that is actually a use case I'm worried about

That makes a lot of sense, yes :smile:.

I think we should rewrite the offending examples to store the source into a variable, since it will improve readability (newlines!) and stop suggesting that reeking is a property of strings.

mvz:
[...]

PR #1042, opened 2016-08-26
Introduce rubocop-rspec

mvz:
Implements #1036 based on #1037.

Because #1035 will introduce a lot of changes, this PR is based on that one. For now I have only touched the files that have been changed in #1035. This already gives a good feel for what settings and fixes to our specs are needed.

Some highlights:
* I had to increase the maximum for RSpec/ExampleLength significantly because of our multi-line source literals.
* To fix RSpec/MultipleExpectations, I used RSpec's [compound expectations](http://www.relishapp.com/rspec/rspec-expectations/v/3-5/docs/compound-expectations), in one case with a [negated matcher](http://www.relishapp.com/rspec/rspec-expectations/v/3-5/docs/define-negated-matcher).

I'll rebase this once #1035 is done.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ group :development do
   gem 'rake',          '~> 11.1'
   gem 'rspec',         '~> 3.0'
   gem 'rubocop',       '~> 0.42.0'
+  gem 'rubocop-rspec', '~> 1.7'
   gem 'simplecov',     '~> 0.12.0'
   gem 'yard',          '~> 0.9.5'
   gem 'activesupport', '~> 4.2'
--- a/Gemfile
+++ b/Gemfile
@@ -10,11 +10,14 @@ group :development do
   gem 'mutant-rspec',  '~> 0.8.8'
   gem 'rake',          '~> 11.1'
   gem 'rspec',         '~> 3.0'
-  gem 'rubocop',       '~> 0.42.0'
-  gem 'rubocop-rspec', '~> 1.7'
   gem 'simplecov',     '~> 0.12.0'
   gem 'yard',          '~> 0.9.5'
   gem 'activesupport', '~> 4.2'
+
+  if RUBY_VERSION >= '2.3'
+    gem 'rubocop',       '~> 0.42.0'
+    gem 'rubocop-rspec', '~> 1.7'
+  end
 
   platforms :mri do
     gem 'redcarpet', '~> 3.3.1'
--- a/lib/reek/spec/should_reek_of.rb
+++ b/lib/reek/spec/should_reek_of.rb
@@ -9,6 +9,8 @@ module Reek
     # code smell.
     #
     class ShouldReekOf
+      include RSpec::Matchers::Composable
+
       # Variant of Examiner that doesn't swallow exceptions
       class UnsafeExaminer < Examiner
         def run
--- a/spec/reek/ast/object_refs_spec.rb
+++ b/spec/reek/ast/object_refs_spec.rb
@@ -2,57 +2,16 @@ require_relative '../../spec_helper'
 require_lib 'reek/ast/object_refs'
 
 RSpec.describe Reek::AST::ObjectRefs do
-  let(:refs) { Reek::AST::ObjectRefs.new }
+  let(:refs) { described_class.new }
 
   context 'when empty' do
-    it 'should report no refs to self' do
+    it 'reports no refs to self' do
       expect(refs.references_to(:self)).to be_empty
     end
   end
 
-  context 'with references to a, b, and a' do
-    context 'with no refs to self' do
-      before(:each) do
-        refs.record_reference(name: :a)
-        refs.record_reference(name: :b)
-        refs.record_reference(name: :a)
-      end
-
-      it 'should report no refs to self' do
-        expect(refs.references_to(:self)).to be_empty
-      end
-
-      it 'should report :a as the max' do
-        expect(refs.most_popular).to include(:a)
-      end
-
-      it 'should not report self as the max' do
-        expect(refs.self_is_max?).to eq(false)
-      end
-
-      context 'with one reference to self' do
-        before(:each) do
-          refs.record_reference(name: :self)
-        end
-
-        it 'should report 1 ref to self' do
-          expect(refs.references_to(:self).size).to eq(1)
-        end
-
-        it 'should not report self among the max' do
-          expect(refs.most_popular).to include(:a)
-          expect(refs.most_popular).not_to include(:self)
-        end
-
-        it 'should not report self as the max' do
-          expect(refs.self_is_max?).to eq(false)
-        end
-      end
-    end
-  end
-
   context 'with many refs to self' do
-    before(:each) do
+    before do
       refs.record_reference(name: :self)
       refs.record_reference(name: :self)
       refs.record_reference(name: :a)
@@ -62,21 +21,21 @@ RSpec.describe Reek::AST::ObjectRefs do
       refs.record_reference(name: :self)
     end
 
-    it 'should report all refs to self' do
+    it 'reports all refs to self' do
       expect(refs.references_to(:self).size).to eq(4)
     end
 
-    it 'should report self among the max' do
+    it 'reports self among the max' do
       expect(refs.most_popular).to include(:self)
     end
 
-    it 'should report self as the max' do
+    it 'reports self as the max' do
       expect(refs.self_is_max?).to eq(true)
     end
   end
 
   context 'when self is not the only max' do
-    before(:each) do
+    before do
       refs.record_reference(name: :a)
       refs.record_reference(name: :self)
       refs.record_reference(name: :self)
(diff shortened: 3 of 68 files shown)


backus:
This is awesome :)

mvz:
@backus it seems `described_class` doesn't play well with shared examples: The autocorrects in  	2ac16f6 break the specs in `smell_warning_spec.rb` (Fixed in 67acd13 by throwing out the shared examples, because it was only used in one place).

backus:
@mvz which line in particular? The diff is pretty big.

mvz:
@backus this one: https://github.com/troessner/reek/commit/2ac16f682ce4d073b0280b2172d277b6cd658b8e#diff-b7c4a27fee5fdb10ab663833e4b82ac7L118.

However, I found out what the problem was: There's a `let(:class) { 'FeatureEnvy' }` just a few lines above it that is completely unnecessary and confuses `described_class`. RSpec probably does a `self.class.described_class` somewhere.

mvz:
Rebased and updated. I've left several exclusions in `.rubocop.yml` that are either quite complex and need some refactoring to fix, or that I think require an update of rubocop-rspec. I'll create tickets for those.

mvz:
Hm .. we also need to think about the ruby 2.2 dependency that rubocop-rspec apparently has.

backus:
> Hm .. we also need to think about the ruby 2.2 dependency that rubocop-rspec apparently has.

Should it matter if it is a development dependency?

backus:
Also wow I haven't checked in on this PR in a while. Awesome work!

mvz:
> Should it matter if it is a development dependency?

We just need to do some juggling to make the build green on Travis.

troessner:
>> You could update your build to only run rubocop on the most recent ruby version. I think it is redundant to run it three times

Fully agree. Let's do this in the course of this PR as well ;)

mvz:
.. and it's green! Suggestions for less hacky ruby version handling are more than welcome.

spec/reek/ast/object_refs_spec.rb
```diff
@@ -2,57 +2,16 @@
 require_lib 'reek/ast/object_refs'
 
 RSpec.describe Reek::AST::ObjectRefs do
-  let(:refs) { Reek::AST::ObjectRefs.new }
+  let(:refs) { described_class.new }
```
troessner: How about just using `subject` here and below?

spec/reek/ast/object_refs_spec.rb
```diff
@@ -2,57 +2,16 @@
...
-          expect(refs.most_popular).to include(:a)
-          expect(refs.most_popular).not_to include(:self)
-        end
-
-        it 'should not report self as the max' do
-          expect(refs.self_is_max?).to eq(false)
-        end
-      end
-    end
-  end
```
troessner: All the specs above were unnecessary?

spec/reek/ast/sexp_extensions_spec.rb
```diff
@@ -402,73 +402,70 @@
 
 RSpec.describe Reek::AST::SexpExtensions::ModuleNode do
   context 'with a simple name' do
-    subject do
```
troessner: So using `subject` is considered bad, right?

spec/reek/cli/options_spec.rb
```diff
@@ -2,23 +2,25 @@
 require_lib 'reek/cli/options'
 
 RSpec.describe Reek::CLI::Options do
+  let(:options) { described_class.new }
```
troessner: I disagree 100% that this is better than using `subject`. 
@waldyr what's your take on this?

spec/reek/context/code_context_spec.rb
```diff
@@ -90,18 +90,16 @@
         end
       end
 
-      context 'with no block' do
-        it 'returns an empty array of ifs' do
-          expect(ctx.each_node(:if, [])).to be_empty
-        end
+      it 'returns an empty array of ifs when no block is passed' do
```
troessner: Nice. Overall we had / have way to many 1-example contexts, something we kill throughout our spec base.

spec/reek/report/code_climate_formatter_spec.rb
```diff
@@ -1,45 +1,47 @@
...
+    end
+
+    it 'sets a content based on the smell detector' do
+      expect(json['content']['body']).
+        to eq "A _Utility Function_ is any instance method that has no dependency on the state of the instance.\n"
+    end
+
+    it 'sets remediation points based on the smell detector' do
+      expect(json['remediation_points']).to eq 250_000
+    end
```
troessner: 👍

spec/reek/smells/attribute_spec.rb
```diff
@@ -23,12 +23,9 @@ class Alfa
...
 
-    expect(src).to reek_of(:Attribute,
-                           lines:   [2],
-                           context: 'Alfa#bravo')
-    expect(src).to reek_of(:Attribute,
-                           lines:   [3],
-                           context: 'Alfa#charlie')
+    expect(src).
+      to reek_of(:Attribute, lines: [2], context: 'Alfa#bravo').
+      and reek_of(:Attribute, lines: [3], context: 'Alfa#charlie')
```
troessner: TIL "and"

spec/reek/smells/feature_envy_spec.rb
```diff
@@ -32,62 +32,69 @@ def golf(hotel)
...
-                           lines: [7, 7],
-                           name:  'hotel')
+    expect(src).
+      to reek_of(:FeatureEnvy, lines: [3, 3], name:  'charlie').
+      and reek_of(:FeatureEnvy, lines: [7, 7], name:  'hotel')
   end
 
   it 'does not report use of self' do
-    expect('def alfa; self.to_s + self.to_i; end').not_to reek_of(:FeatureEnvy)
+    src = 'def alfa; self.to_s + self.to_i; end'
```
troessner: Yeah, good call here and below. I thought of doing this as well while working on #1056 but then decided against it to not make the PR bigger than it already was.

spec/reek/source/source_locator_spec.rb
```diff
@@ -9,19 +9,23 @@
...
       it 'does not scan hidden directories' do
         sources = described_class.new([path]).sources
 
         expect(sources).not_to include(*paths_that_are_expected_to_be_ignored)
-        expect(sources).to eq expected_paths
+      end
+
+      it 'scans directories that are not hidden' do
+        sources = described_class.new([path]).sources
+
```
troessner: We never did newlines in between 1-line setups and 1-line expectations. Would you like to change this policy?

spec/spec_helper.rb
```diff
@@ -93,6 +94,8 @@ def sexp(type, *children)
   end
 end
 
+RSpec::Matchers.define_negated_matcher :not_reek_of, :reek_of
```
troessner: Wow, pretty cool!

lib/reek/spec/should_reek_of.rb
```diff
@@ -9,6 +9,8 @@ module Spec
     # code smell.
     #
     class ShouldReekOf
+      include RSpec::Matchers::Composable
```
mvz: It is needed to make `#and` work.

spec/reek/ast/object_refs_spec.rb
mvz: Yes, they were essentially a duplicate of the specs below, but with more nesting :smile:.

spec/reek/ast/sexp_extensions_spec.rb
mvz: `rubocop-rspec` considers explicit use of `subject` in expectations bad, e.g.:

```
  expect(subject).to match_something
```

spec/reek/source/source_locator_spec.rb
mvz: No, the newline is a leftover from the original single spec having it.

spec/reek/ast/object_refs_spec.rb
mvz: I find the fact that `subject` is actually `described_class.new` automatically a little confusing.

mvz:
So the big thing is this 'explicit subject' stuff. Basically, rubocop-rspec is fine with stuff like

```
  subject { foo }

  it { is_expect_to match_bar }
```

but not with:

```
  subject { foo }
  it 'bars' do
    expect(subject).to match_bar
  end
```

There are four solutions for the latter case:
* Make the subject implicit. I'm not a fan.
* Define the object under test with let:

  ```
  let(:the_foo) { foo }
  ```

* Give the subject a name:
  ```
  subject(:the_foo) { foo }
  ```

* Disable this cop :smile:

Gemfile
```diff
@@ -10,11 +10,15 @@ group :development do
   gem 'mutant-rspec',  '~> 0.8.8'
   gem 'rake',          '~> 11.1'
   gem 'rspec',         '~> 3.0'
-  gem 'rubocop',       '~> 0.42.0'
   gem 'simplecov',     '~> 0.12.0'
   gem 'yard',          '~> 0.9.5'
   gem 'activesupport', '~> 4.2'
 
+  if RUBY_VERSION >= '2.2'
```
troessner: And shouldn't we update our travis configuration as well? I mean, the way it is set up below it will still output something on STDOUT which I find confusing.

troessner:
Ok, then let's go with 

>> Define the object under test with let:

Not the biggest fan but I can live with it.
Now only question left (regarding >= 2.2) and then lgtm ;)

Gemfile
mvz: Do you mean where it says 'Install rubocop to run its rake tasks'? That is output by the dummy rake task I generate to avoid having to fiddle with the `:ci` task. I'm not too happy with how that turned out, suggestions are welcome. One option is to add the `:rubocop` task to the `:ci` task when it is defined, but that doesn't make it easier to see what `rake ci` will do.

mvz: > Why not >= 2.3?

I wanted to just eliminate the versions where rubocop-rspec wouldn't install, but I'm happy to limit this more.

troessner: Ok, then let's make this >= 2.3 ;)

mvz:
Updated the 2.2 to 2.3.

backus:
:tada: Hurray!

backus:
> Ok, then let's go with

> > Define the object under test with let:

@troessner I'm curious, why you prefer this over naming the subject? It seems like having `subject` at the top of the specs makes it very clear what the most important object of the test is.

troessner:
My rationale here is that if we agree to not use "subject" the way rspec envisions it we should rather not use it at all. Since i am not the biggest "subject" fan myself (i also find it occassionally confusing) not using it would ve fine for me.
@mvz @waldyr whats your take on this?

backus:
Actually naming the subject is more or less the way the rspec creator envisioned: http://blog.davidchelimsky.net/blog/2012/05/13/spec-smell-explicit-use-of-subject/

mvz:
> @mvz @waldyr whats your take on this?

I hardly ever use `subject`. I'm usually testing some method, so I would still end up with `expect(thing.foo).to eq bar`, so it makes not much sense to me to designate `thing` as the subject.

I find all the shortcuts like `it { is_expected_to be_bar(foo) }` to completely obscure the meaning of what I'm trying to test, so I don't use them. This means using `let` instead of a named subject works just as well.

backus:
btw regardless of which you chose I think it is correct enough. I personally like seeing a `subject` at the top even if it is functionally equivalent to `let` with the same name. Different strokes though

Issue #1044, opened 2016-08-27
ManualDispatch only reports once


troessner:
This 

```Ruby
  class Dummy
    def m1(a)
      return true if a.respond_to?(:to_a)
      false if a.respond_to?(:to_s)
    end
  end
````

will report ManualDispatch once for the first respond_to? when it should report it twice.

@backus would you like to handle this?

**Comment by backus** [comment]
Sure

**Comment by mvz** [comment]
The second one is probably removed since it only differs by line number.

**Comment by backus** [comment]
@mvz correct:

```ruby
$ be rspec
Run options: include {:focus=>true}

Frame number: 0/67

From: /Users/johnbackus/Projects/reek/lib/reek/smells/smell_detector.rb @ line 45 Reek::Smells::SmellDetector#run_for:

    40: def run_for(context)
    41:   return [] unless enabled_for?(context)
    42:   return [] if exception?(context)
    43:
    44:   require 'pry'; require 'pry-byebug'; binding.pry
 => 45:   sniff(context)
    46: end

1 (#<Reek::Smells::ManualDispatch>) sniff(context)
=> [#<Reek::Smells::SmellWarning:0x007ff881ee37d0
  @context="Dummy#call",
  @lines=[3],
  @message="manually dispatches method call",
  @parameters={},
  @smell_detector=
   #<Reek::Smells::ManualDispatch:0x007ff89abb27a0
    @config=
     #<Reek::Smells::SmellConfiguration:0x007ff89abb2408
      @options={"enabled"=>true, "exclude"=>[]}>>,
  @source="string">,
 #<Reek::Smells::SmellWarning:0x007ff881ee31b8
  @context="Dummy#call",
  @lines=[4],
  @message="manually dispatches method call",
  @parameters={},
  @smell_detector=
   #<Reek::Smells::ManualDispatch:0x007ff89abb27a0
    @config=
     #<Reek::Smells::SmellConfiguration:0x007ff89abb2408
      @options={"enabled"=>true, "exclude"=>[]}>>,
  @source="string">]
```

is there something different I should have done for my implementation?

mvz:
I recently fixed basically the same issue in NilCheck: 7d59f2adc775f5f26aaabe45b4716594f305b2f2.

Another option is to make the smells sufficiently different by including more information.

A third option would be to create some merging mechanism that would combine the sets of lines for 'identical' smells.

[...]

PR #1048, opened 2016-08-27
Add `#lines` to smell warning identity

backus:
This is related to #1044 but I don't want to close that issue because I think it also revealed that the smell's message should be improved.

I've broken this PR into three commits:

1. Adding `#lines` to the `#sort_key` method for `SmellWarning`. If two smells share the same type, context, and message but are on different lines then I think reek should treat these as being different smells.

2. `SmallWarning` seems like a value  object to me so I renamed `#sort_keys` to `#identifying_values` because that seems to better convey the very important role that method is playing.

3. Freezing `SmellWarning` on initialization. This doesn't break any specs or public API and, if you agree that `SmellWarning` is a Value then it should forbid mutating its state. 

These are certainly opinionated changes so I wont be offended if you disagree with one or more of the commits.


backus:
@mvz or @troessner this is ready for review

mvz:
I agree with all three changes. The only remaining question for me is whether we should merge the ManualDispatch smells.

troessner:
Merged, awesome!

👍 for value objects!

Wdyt about integrating [Adamantium](https://github.com/dkubb/adamantium) into Reek?

backus:
@troessner I'd support that I think. I would want to see how many things are being deep frozen before saying for sure whether it should be a dependency but I think it is usually a safe call

PR #1053, opened 2016-08-30
Update InstanceVariableAssumption docs.

### LINKED ISSUE #1051: InstanceVariableAssumption problems with inheritance
Given

```Ruby
class Parent
  def initialize(omg)
    @omg = omg
  end
end

class Child < Parent
  def initialize(omg)
    super
  end

  def foo
    @omg
  end
end
```

Reek would report:

>>
  [7]:InstanceVariableAssumption: Child assumes too much for instance variable @omg [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]

I don't think we should report here. I can't see anything dirty about it.
For me this smell detector was about doing things like this:

```Ruby
class Foo
  def bar
    @hola = 42
  end
  
  def baz
    @hola + 99
  end
end
```

but **not** about the example above.

What do you guys think?

troessner:
Fixes #1051

### DIFF
--- a/docs/Instance-Variable-Assumption.md
+++ b/docs/Instance-Variable-Assumption.md
@@ -6,7 +6,7 @@ Classes should not assume that instance variables are set or present outside of 
 
 Good:
 
-```ruby
+```Ruby
 class Foo
   def initialize
     @bar = :foo
@@ -18,23 +18,9 @@ class Foo
 end
 ```
 
-Bad:
+Good as well:
 
-```ruby
-class Foo
-  def go_foo!
-    @bar = :foo
-  end
-
-  def foo?
-    @bar == :foo
-  end
-end
-```
-
-Good:
-
-```ruby
+```Ruby
 class Foo
   def foo?
     bar == :foo
@@ -42,6 +28,20 @@ class Foo
 
   def bar
     @bar ||= :foo
+  end
+end
+```
+
+Bad:
+
+```Ruby
+class Foo
+  def go_foo!
+    @bar = :foo
+  end
+
+  def foo?
+    @bar == :foo
   end
 end
 ```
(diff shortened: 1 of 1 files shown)


docs/Instance-Variable-Assumption.md
```diff
@@ -18,30 +18,30 @@ class Foo
...
+  def go_foo!
+    @bar = :foo
   end
 
-  def bar
-    @bar ||= :foo
+  def foo?
+    @bar == :foo
   end
 end
```
troessner: The diff above is hard to read. I basically just switched the two examples around, that's all.

docs/Instance-Variable-Assumption.md
```diff
@@ -64,6 +64,50 @@ would report:
...
+
+```Ruby
+class Parent
+  def initialize(omg)
+    @omg = omg
+  end
+end
+
+class Child < Parent
+  def initialize(omg)
```
mvz: This initializer can go I think.

docs/Instance-Variable-Assumption.md
```diff
@@ -64,6 +64,50 @@ would report:
...
+class Parent
+  attr_reader :omg
+
+  def initialize(omg)
+    @omg = omg
+  end
+end
+
+class Child < Parent
+  def initialize(omg)
```
mvz: Same here.

docs/Instance-Variable-Assumption.md
```diff
@@ -64,6 +64,50 @@ would report:
...
+    super
+  end
+
+  def foo
+    omg
+  end
+end
+```
+
+Directly accessing instance variables is considered a smell because it [breaks encapsulation](http://designisrefactoring.com/2015/03/29/organizing-data-self-encapsulation/) and makes it harder to reason about code.
```
mvz: That article goes a lot further than our smell detectors. Do we have another text to link to? Or maybe we can explain it ourselves?

troessner: >> That article goes a lot further than our smell detectors.

I have to disagree on this. It explains why to prefer setters and getter over direct access which is what we should explain here.

>> Or maybe we can explain it ourselves?

What do you feel is missing in the article?

troessner:
@mvz just updated the article and addressed your comments.

PR #1055, opened 2016-08-31
Update Simulated-Polymorphism.md

chadmetcalf:
Readers can be all different types.


troessner:
Well spotted, merged, thanks!

troessner:
But shouldn't it then be "the readers" as well? I'm not a native speaker so I kind of trust your judgement here ;)

chadmetcalf:
Just to make sure that English is as confusing as possible, "they" is a singular gender-neutral pronoun as well as the pronoun for a group of folks. So either way works!

PR #1056, opened 2016-09-01
Refactor and simplify all our smell detector specs.

troessner:
Follow up to #1035 after this one got bigger and bigger now with a better branch name and a better commit message.

This is still - again - wip. Right now I'm literally going through all the specs to make them simpler and more concise (my current position is FeatureEnvy). This will take me a couple of more days.

What I did:

* Killed all the shared examples
* Introduced our Spec matcher everywhere. This and the bullet point above also have the nice side effect that now all requires at the top of a smell detector spec file look
like this:

```Ruby
require_lib 'reek/smells/irresponsible_module'
````

And that's it. No more references to `Examiner`, `CodeContext` and god knows what else.

* Removed silly specs. Some areas of Reek are shockingly under-speced whilst others are over-speced. If we have a threshold of 4 x it is just silly to have a spec that 3 x
does not smell. And 2 x does not smell. And 1 x does not smell. That's just bloated code.

* Removed redundant specs. Bigger smell detector specs were full of duplication.

* Removed foo / bar because I honestly can't see that anymore and replaced it with "alpha/bravo/charlie" as discussed.

* Made our syntax consistent. Introduced heredocs everywhere and removed multi line strings. Introduced newlines. And so on.

* Simplified a lot, a lot of our examples. For instance:

```Ruby
src = 'def simple(arg) args = {}; if arg then args.merge(\'a\' => \'A\') end end'
# to
src = 'def alpha(bravo); if bravo then charlie end; end'
# or
src = 'def <=>(other) Options[:sort_order].compare(self, other) end'
# to
src = 'def alfa(bravo); alfa.bravo(self); end'
```

The meaning stays the same but the example is less confusing and better readable.

* Introduced the same general examples for every spec at the beginning. Every smell detector spec now starts out like this:

```Ruby
RSpec.describe Reek::Smells::IrresponsibleModule do
  it 'reports the right values' do
    src = <<-EOS
      class Dummy
      end
    EOS

                           lines:   [1],
                           context: 'Dummy',
                           source:  'string')
  end

  it 'does count all occurences' do
    src = <<-EOS
      class Outer
        # Method is necessary because we don't count empty classes.
        def something; end
        class Inner
        end
      end

    EOS

                               lines:   [1],
                               context: 'Outer')
                           lines:   [4],
                           context: 'Outer::Inner')
  end
end
```

The 2nd spec 'does count all occurences' is only added when it actually makes sense, so there are some detectors that only start with the first spec above.

* Removed contexts that didn't make any sense. E.g. I removed non-nested contexts with only example and just updated the example description. Same goes for specs that were
covered by one giant context.


mvz:
Awesome work, @troessner!

CONTRIBUTING.md
```diff
@@ -52,12 +54,82 @@ start contributing.
...
 When it comes to Rspec we're trying to follow [betterspecs](http://betterspecs.org/).
-Additonally you can find an excellent cheat sheet on how to write idiomatic Rspec [here](http://www.rubypigeon.com/posts/rspec-core-cheat-sheet).
+We're not using Rspec's [shared examples](https://www.relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples) because we find
+them rather harming than helpful.
+You can find an excellent cheat sheet on how to write idiomatic Rspec [here](http://www.rubypigeon.com/posts/rspec-core-cheat-sheet).
+
+#### Spec naming conventions
+
+We do not use the popular "foo" / "bar" naming when it comes to the question "how to come up with good example names?".
+Instead, we use the [military alphabet](https://en.wikipedia.org/wiki/NATO_phonetic_alphabet) in descending order
```
mvz: `s/descending/ascending/`

spec/reek/smells/unused_private_method_spec.rb
```diff
@@ -1,100 +1,147 @@
...
   context 'unused private methods' do
     it 'reports instance methods' do
       source = <<-EOF
-        class Car
+        class Alfa
           private
-          def start; end
-          def drive; end
+          def charlie; end
+          def charlie; end
```
mvz: Should be bravo, then charlie.

CONTRIBUTING.md
troessner: Errr, yeah, fixed!

spec/reek/smells/unused_private_method_spec.rb
troessner: Good catch, fixed!

troessner:
Status update:

* UtilityFunction + FeatureEnvy
* NestedIterators - RepeatedConditional

are still refactor worthy.
And then I need to go through each spec again once more.
And then it's reviewable ;)

waldyr:
I closed this by accident. I was going to comment something and then I decided not to comment. I ended up closing the issue. Really sorry.

troessner:
This bad boy is ready for review! The diff is probably wayyyyy to confusing, might make more sense to look at the specs directly.

CONTRIBUTING.md
```diff
@@ -52,12 +54,82 @@ start contributing.
...
+    charlie.delta
+  end
+end
+```
+
+#### Smell detector conventions
+
+All smell detector specs start out with 2 generic examples like this (the second one only if it makes sense):
+
+TODO: Example below does not get rendered properly. Wtf.
```
mvz: Seems to be fine now, so this line can go :tada:!

spec/reek/smells/irresponsible_module_spec.rb
```diff
@@ -1,208 +1,191 @@
...
-    expect(src).to reek_of(:IrresponsibleModule, context: 'Qux::Foo')
-  end
-
-  it 'reports top level classes defined through assignment' do
-    src = <<-EOS
-      Foo = Class.new Bar
-    EOS
-    expect(src).to reek_of(:IrresponsibleModule, context: 'Foo')
-  end
+    it "does not report a namespace #{scope} that has a nested #{scope} through assignment" do
```
mvz: The nested item is always a class, so you shouldn't interpolate scope there.

spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -100,152 +141,104 @@ def super_call_with_block
...
-
-    it 'reports the name of the method and line of the deepest iterator' do
-      source = <<-EOS
-        def foo
-          bar do |bar|
-            baz {|baz| }
-          end
-        end
-      EOS
-      expect(source).to reek_of(:NestedIterators, lines: [3])
```
mvz: I think this test is still important, but this assertion can also be merged into the one that is now in line 128.

spec/reek/smells/nested_iterators_spec.rb
```diff
@@ -100,152 +141,104 @@ def super_call_with_block
...
-
-    it 'reports all lines on which nested iterators occur' do
-      source = <<-EOS
-        def bad
-          @fred.each {|item| item.each {|part| @joe.send} }
-          @jim.each {|ting| ting.each {|piece| @hal.send} }
-        end
-      EOS
-
-      expect(source).to reek_of(:NestedIterators, lines: [2, 3])
```
mvz: I think this case and the next are still important and not covered by the remaining tests. They are about the interaction of separate cases of NestedIterators within a single method.

spec/reek/smells/smell_repository_spec.rb
```diff
@@ -15,7 +15,7 @@
       expect(smell_types).to_not include(Reek::Smells::SmellDetector)
     end
 
-    it 'should return the smell types in alphabetic order' do
+    it 'should return the smell types in alfabetic order' do
```
mvz: -> alphabetic

mvz:
Phew, that **is** a big PR. I have put a few comments, but for the rest, I think we should merge this and then use mutant to sort out any bits we may have under-specified.

spec/reek/smells/nested_iterators_spec.rb
troessner: Updated the example ending in line 128 to be

>>     expect(src).to reek_of(:NestedIterators, depth: 3, lines: [4])

spec/reek/smells/irresponsible_module_spec.rb
troessner: Good catch, thanks.

CONTRIBUTING.md
troessner: Updated.

spec/reek/smells/nested_iterators_spec.rb
troessner: What do you mean? Isn't this case here covered by the "does count all occurrences" example in line 22 [here](https://github.com/troessner/reek/pull/1056/files#diff-b34733d371fc76dfe6eabf2134a73744R22)?

troessner:
@mvz addressed all of your feedback as far as possible. How about merging this one in its current state and then address whatever you feel missing in the NestedIterators specs afterwards?

spec/reek/smells/nested_iterators_spec.rb
mvz: No, that one has two cases with different nesting, and hence two smells, whereas this one has two cases with the same nesting, hence one smell with two lines.

troessner: Good point, let me re-add this as separate commit so you can review it.

troessner: @mvz added a spec for this as separate commit [here](https://github.com/troessner/reek/pull/1056/commits/44263ed14ff99dfc72f3fee05f861b0a2e37e0e0), please review ;)

mvz: Looks good, :shipit:.

mvz:
Ready to merge, @troessner!

troessner:
Squashed, please merge @mvz ;)

mvz:
:tada:

Issue #961, opened 2016-06-28
Add reporting of unneeded inline suppressions


focusshifter:
[Rubocop's Lint/UnneededDisable](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/UnneededDisable) is performing a very useful task - it runs the disabled code smell check nevertheless and reports if it is no longer needed.

It would be nice of reek to do the same - report if some inline suppression is no longer needed and could be removed.

troessner:
>> It would be nice of reek to do the same - report if some inline suppressions are no longer needed and could be removed.

Yes, yes and double yes :)
Couldn't agree more, this is definitely a feature `Reek` should have.

I'm blocked for the next weeks but if nobody looks into this feature in the meantime (looking at you @mvz  or @waldyr) I'll definitely get this implemented.

waldyr:
@troessner @mvz I look forward to implement this feature.

How can it be achieved and how would you do it?

mvz:
> How can it be achieved and how would you do it?

I think the idea would be to run the detectors both with and without the suppressions. The question is where do you put this duplication?

If you put it at the top, Reek would do a lot of duplicate work, since most detectors will have no suppressions for most of the code.

Therefore, it makes more sense to do the duplication at the bottom, i.e., when running one detector on one context: For a given detector and context,
* First run the detector on the context as usual.
* Then, if there is no smell,
* And there is context-dependent configuration,
* Run the detector again.
* If there is still no smell, flag the context-dependent configuration as unneeded.

I would also take a look at how RuboCop handles this.

troessner:
Sounds like a plan @mvz !

waldyr:
Thanks @mvz :)

waldyr:
They have a cop to do it [UnneededDisable Cop](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/lint/unneeded_disable.rb).

Based on how they do it, Is it possible with today's Reek?
- Read the reek's magic comments
- Call the Detector being modified without the comments' configuration
- If there is no smell, flag it as unneeded.

mvz:
Well, like RuboCop this would be a slightly different detector. It needs to know about other detectors for example.

I think RuboCop's method works best if we *only* look at disabling, since then you only have to run the detector enabled. If we also want to look at other configuration (`max_something`, etc.), then we have to run the detector both with and without the inline config, which means extra work since we already ran the detector with the config when we were detecting its smell.

[...]

PR #1060, opened 2016-09-08
Unneeded suppression detector

waldyr:
#961

### DIFF
--- a/lib/reek/smells/smell_detector.rb
+++ b/lib/reek/smells/smell_detector.rb
@@ -38,10 +38,13 @@ module Reek
       end
 
       def run_for(context)
-        return [] unless enabled_for?(context)
         return [] if exception?(context)
 
-        sniff(context)
+        if enabled_for?(context)
+          sniff(context)
+        else
+          check_suppression(context)
+        end
       end
 
       def exception?(context)
@@ -58,6 +61,12 @@ module Reek
 
       def enabled_for?(context)
         config.enabled? && config_for(context)[SmellConfiguration::ENABLED_KEY] != false
+      end
+
+      def check_suppression(context)
+        return [] unless sniff(context).empty?
+
+        raise Reek::Errors::UnneededSupression
       end
 
       def value(key, ctx)
--- a/lib/reek/errors/unneeded_suppression.rb
+++ b/lib/reek/errors/unneeded_suppression.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+module Reek
+  module Errors
+    # Gets raised when trying to disable a smell that does not reek.
+    # This might happen for multiple reasons. The users might have fixed
+    # the smell or speculating in advance and forgetting to remove the.
+    class UnneededSupression < RuntimeError
+      UNNEEDED_SUPPRESSION_MESSAGE = <<-EOS.freeze
+        Error: UnneededSuppression
+      EOS
+
+      def initialize
+        message = format(UNNEEDED_SUPPRESSION_MESSAGE)
+        super message
+      end
+    end
+  end
+end
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -2,6 +2,7 @@
 require_relative 'context_builder'
 require_relative 'source/source_code'
 require_relative 'errors/bad_detector_in_comment_error'
+require_relative 'errors/unneeded_suppression'
 require_relative 'errors/bad_detector_configuration_in_comment_error'
 require_relative 'smells/detector_repository'
 
@@ -121,6 +122,7 @@ module Reek
       begin
         examine_tree
       rescue Errors::BadDetectorInCommentError,
+             Errors::UnneededSupression,
              Errors::BadDetectorConfigurationInCommentError => exception
         warn exception
         []
--- a/spec/reek/examiner_spec.rb
+++ b/spec/reek/examiner_spec.rb
@@ -145,6 +145,20 @@ RSpec.describe Reek::Examiner do
     end
   end
 
+  describe 'unneeded comment config' do
+    it 'prints out a proper message' do
+      source = <<-EOS
+        # :reek:TooManyStatements
+        def alfa; end
+      EOS
+      expected_output = 'Error: UnneededSuppression'
+
+      expect do
+        described_class.new(source).smells
+      end.to output(/#{expected_output}/).to_stderr
+    end
+  end
+
   describe 'bad comment config' do
     context 'unknown smell detector' do
       let(:source) do


waldyr:
@mvz @troessner 

How should I proceed from here? Should I create another detector which does not inherit from SmellDetector and all detectors know and depend?

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -46,6 +46,18 @@ def ivars
       expect(src).not_to reek_of(described_class)
     end
 
+    it 'reeks of unneeded supression' do
+      src = <<-EOS
+        # :reek:TooManyInstanceVariables: { enabled: false }
+        class Empty
```
troessner: Let's make this alpha, bravo, charlie and so on as outlined in #1056

waldyr: Sure thing

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -46,6 +46,18 @@ def ivars
       expect(src).not_to reek_of(described_class)
     end
 
+    it 'reeks of unneeded supression' do
+      src = <<-EOS
+        # :reek:TooManyInstanceVariables: { enabled: false }
```
mvz: Just `# :reek:TooManyInstanceVariables:` should work. Note the trailing colon.

waldyr:
@troessner @mvz Ready for review. But I have some questions?
- Should I create a class to handle the if-elsif-else or it's an unneeded level of indirection?
- How should I test this behaviour?
  - A specific spec (smell_detector_spec)
  - In each smell detector

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -58,7 +58,20 @@ def bravo
       end
     EOS
 
-    expect(src).not_to reek_of(:TooManyInstanceVariables).with_config(config)
+    expect(src).not_to reek_of(:TooManyInstanceVariables)
+  end
+
+  it 'reeks of unneeded supression' do
+    src = <<-EOS
```
troessner: I feel that this test is not really helpful since it would succeed regardless of your code changes above.
How about rather  testing the sniff method?
Also we should test the full stack here including the output so lets add a cucmber feature for this.

lib/reek/smells/smell_detector.rb
```diff
@@ -38,10 +38,19 @@ def contexts
       end
 
       def run_for(context)
-        return [] unless enabled_for?(context)
         return [] if exception?(context)
 
-        sniff(context)
+        smells = sniff(context)
+
+        if enabled_for?(context)
```
troessner: You're checking for a boolean but have 3 branches which is confusing.
I'd prefer

```Ruby
if
else
  if
  else
  end
end
```

mvz: Then I would move the second if .. else into a separate method, and even move the call to `sniff` inside the conditional:

```ruby
if enabled_for? context
  sniff(context)
else
  check_for_unneeded_suppression(context)
end
```

or even (going back to the original code layout):

```ruby
return check_for_unneeded_suppression(context) unless enabled_for? context
sniff(context)
```

spec/reek/smells/too_many_instance_variables_spec.rb
```diff
@@ -58,7 +58,20 @@ def bravo
       end
     EOS
 
-    expect(src).not_to reek_of(:TooManyInstanceVariables).with_config(config)
+    expect(src).not_to reek_of(:TooManyInstanceVariables)
```
mvz: This change shouldn't be here.

lib/reek/smells/smell_detector.rb
waldyr: I like most back to the original code layout with an additional method

spec/reek/smells/too_many_instance_variables_spec.rb
mvz: Are you sure it succeeds without the changes?

I think the test would be much improved if it added an expectation on the smell message.

To make this more a test about SmellDetector, it should be moved to the `smell_detector_spec.rb` file.

The `sniff` method is not what is tested here.

mvz:
A big design question is what smell type the smell warnings for 'Unneeded Suppression' should use.

waldyr:
The current code is sending the same smell_type with a different message.

lib/reek/smells/smell_detector.rb
troessner: I'm a big fan of early returns instead of if/else's but I think in this case @mvz first suggestion:

```Ruby
if enabled_for? context
  sniff(context)
else
  check_for_unneeded_suppression(context)
end
```

is way more readable.

spec/reek/smells/too_many_instance_variables_spec.rb
waldyr: I messed it up when rebased.

waldyr:
> So, I think we should have an added UnneededSuppression class.

@mvz should this new class inherit from SmellDetector?

mvz:
> @mvz should this new class inherit from SmellDetector?

Probably not needed since it only needs to respond to the interface SmellWarning requires of it (which is tiny).

waldyr:
@mvz @troessner Ready for another round of reviews

mvz:
> Even thought UnneededSuppresion has its own smell_type the smell type of the specs is coming from the smell I've used suppressed.

This happens because `reek_of` only runs the named detector, so in order to have it find the UnneededSuppression, you need to run it with the name of the suppressed detector. As a workaround, you can use `reek_only_of`, which runs all detectors. Also, we need to update `reek_of` to check the resulting smell type, since it is now no longer automatically the type of the detector.

waldyr:
Nice! Gonna update the specs and docs.

waldyr:
@mvz @troessner ready for another round of reviews!

troessner:
I'll review the rest tonight...

troessner:
Done reviewing, sorry for the many comments :)

mvz:
@troessner your comments are not showing up. I think you need to press the new 'I'm done reviewing' button (not sure what the label is).

troessner:
>> @troessner your comments are not showing up. I think you need to press the new 'I'm done reviewing' button (not sure what the label is).

thanks for the tip! Damn it. The reason was that somewhere, somewhen I had started to write a comment but then didnt go through with it. As a consequence, all subsequent comment were listed as pending. I also couldnt then delete this unfinished comment cause otherwise all pending comments would have been deleted as well. Hence my comment

>> ......test comment because I really think github fucked up this feature.

above.

Anyway, now you should see all comments.

lib/reek/smells/smell_detector.rb
```diff
@@ -38,10 +39,13 @@ def contexts
...
 
       def run_for(context)
-        return [] unless enabled_for?(context)
         return [] if exception?(context)
 
-        sniff(context)
+        if enabled_for?(context)
+          sniff(context)
+        else
+          ::Reek::Smells::UnneededSuppression.new(context, self).sniff
```
mvz: `context` is a parameter to `sniff`, so that won't work as-is. Also, we would be creating a lot of unneeded UnneededSuppression objects.

mvz: The end goal may be to create something like:
```
UnneededSuppression.new(self).sniff(context)
```

lib/reek/smells/unneeded_suppression.rb
```diff
@@ -0,0 +1,39 @@
...
+    # An unneeded suppression is a Reek's comment disabling a unexistent smell.
+    #
+    # See {file:docs/Unneeded-Suppression.md} for details.
+    class UnneededSuppression
+      def initialize(context, smell_detector)
+        @context = context
+        @smell_detector = smell_detector
+      end
+
+      def sniff
```
mvz: I would move the check for the condition 'disabled but not smelly' back up into `SmellDetector#sniff`, end create a smell warning right there, passing a light-weight smell detector like object to provide the smell warning with what it needs.

lib/reek/smells/unneeded_suppression.rb
```diff
@@ -0,0 +1,39 @@
...
+        if @smell_detector.sniff(@context).empty?
+          exp = @context.exp
+          ctx_source = exp.loc.expression.source_buffer.name
+          [SmellWarning.new(self,
```
mvz: The reason seems to be that I suggested not deriving this class from SmellDetector because I had a different architecture in mind. Sorry for the confusion.

samples/smelly_with_inline_mask.rb
```diff
@@ -1,8 +1,7 @@
 # Smelly class
-# disables :reek:UncommunicativeVariableName
```
mvz: Why wasn't this working before?

lib/reek/smells/unneeded_suppression.rb
troessner: I see. But then maybe the overall architecture is not the right one.
I mean, we **need** stuff that is in SmellDetector (otherwise we're just duplicating code 1:1 like we do here) but at the same time we try really hard to **not** make this a smell detector.
That doesn't seem right.
How about not making this a detector-style thingie but instead follow the same approach you, @mvz, suggested in #1082 at the end?
So:

>> I have in mind to handle it at the Examiner level. So here we would first catch Reek::BadConfigurationError, and then StandardError.

wdyt?

lib/reek/smells/unneeded_suppression.rb
troessner: Again, we're trying really hard to shoehorn this into a detector-style thing. Maybe that's just not the right approach. Please see my comment below as well.

lib/reek/smells/smell_detector.rb
mvz: Actually, in view of my comment below, I think the end goal should be something like:
```ruby
if enabled_for?(context)
  sniff(context)
else
  sniff_unneeded_suppression(context)
end

# [...]

def sniff_unneeded_suppression(context)
  return [] if sniff(context)
  # return correct unneeded suppression smell warnings here
end
```

The 'return correct unneeded suppression smell warnings here' will need some work, since `smell_warning` is currently a private method.

lib/reek/smells/unneeded_suppression.rb
mvz: I'm not entirely sure what you have in mind. Do you mean moving this `enabled_for` check up into Examiner and putting all the logic there?

troessner:
After thinking about it a little more, shouldnt this one be combined with #1082 ?
Both unneeded supression AND malformed configuration comments should be handled by the **exact** same mechanism, shouldnt it?
Wdyt?

mvz:
> Both unneeded supression AND malformed configuration comments should be handled by the exact same mechanism, shouldnt it?

That depends on whether you consider unneeded suppression a smell or an error. We've been proceeding as if it were a smell, but maybe you're right and it should be an error.

waldyr:
Then I'll wait.

troessner:
Alright, seems like we have an agreement here, gentlemen! ;)

I'll go ahead with this plan:

>> Both unneeded supression AND malformed configuration comments should be handled by the exact same mechanism, shouldnt it?
If we agree on this I'd suggest to first finish #1082 and then come back to this one and rebase against master.

and then let's come back to this one again :)

troessner:
@waldyr #1082 has been merged, you can now rebase your branch on master and continue from there on :)

waldyr:
Going to tackle this on the weekend. Too much on my plate right now, sorry :(

troessner:
No problem, take all the time you need :)

waldyr:
@troessner I have a few questions:
- Am I on the right track?
- How should be the message for this error?
- What should I pass to error's constructor?

troessner:
I'll look into your questions in detail the next couple of days, promised.

lib/reek/smells/smell_detector.rb
```diff
@@ -38,10 +38,13 @@ def contexts
...
 
       def run_for(context)
-        return [] unless enabled_for?(context)
         return [] if exception?(context)
 
-        sniff(context)
+        if enabled_for?(context)
+          sniff(context)
+        else
+          check_suppression(context)
```
troessner: Wouldn't that also raise if the smell was disabled via configuration using `exclude`? That would be great, but it also means that we need to output different error message and that we need to make a difference here.
Can you maybe create some cucumber scenarios so we can hash out the details by looking at the concrete scenarios?

lib/reek/smells/smell_detector.rb
```diff
@@ -60,6 +63,12 @@ def enabled_for?(context)
         config.enabled? && config_for(context)[SmellConfiguration::ENABLED_KEY] != false
       end
 
+      def check_suppression(context)
+        return [] unless sniff(context).empty?
```
troessner: I think an "unless" with another negative like "empty?" is hard to read.
How about turning this into a positive like

```Ruby
return [] if sniff(context) > 0
```

Or even better: You are doing the `sniff(context)` in **any** case.
How about doing doing `sniff(context)` once in `run_for` above and then use it here?

troessner:
Ok, couldn't sleep so here we go ;)

>> Am I on the right track?

Yes, I think that's the right way to go. Meaning, not using something that looks like a SmellDetector :)

Please see my fresh comments above.

>> How should be the message for this error?

How about something like [here](https://github.com/troessner/reek/blob/master/lib/reek/errors/bad_detector_configuration_in_comment_error.rb#L7) but with an updated message? 

Another question: This PR just takes enabling / disabling into account, what about obsolete detail configurations like `max_methods` for `TooManyMethods` and so on? I'm fine if we handle this in a separate PR to not make this one too big but we should be aware of where we want to go with this feature.

Overall I think we might be approaching this whole thing the wrong way - we should have talked more and maybe for big features like this a chat like gitter would have been better.

Let's fix this ;)
How about creating high-level cucumber scenarios so we can then discuss the details?

I'm thinking about:

```Cucumber
Scenario: Warns about unneeded suppression of detector via code comments
Scenario: Warns about unneeded suppression of detector via configuration file # or not depending on our discussion!
Scenario: Warns about unneeded suppression via obsolete detector configuration via code comments
```

And then we also need mixed scenarios where we have "clean" files (in the sense that there is no unneeded suppression) and dirty files and so on.
I feel we underestimated the complexity of this topic and now we need to pay for it by stepping back and really think about what we want to achieve here.

lib/reek/smells/smell_detector.rb
mvz: The positive formulation is `if sniff(context).any?`.

I think @waldyr had a common `sniff(context)` first, but that became confusing.

troessner: >> The positive formulation is if sniff(context).any?.

Yes. Or that 😆

waldyr:
Closing in favor of #1153

PR #1065, opened 2016-09-09
Fix UncommunicativeVariableName does not take regex configuration into account.

### LINKED ISSUE #1063: UncommunicativeVariableName does not take regex configuration into account
Found this bug while investigating #1058 

Given this config.reek:

```yaml
---
UncommunicativeVariableName:
  accept:
  - /\d+$/
```

this:

```Ruby
def foo; x1 = 2; end
```

shouldn't smell of UncommunicativeVariableName.
Unfortunately it does:

```
 reek -c config.reek foo.rb
foo.rb -- 1 warning:
  [1]:UncommunicativeVariableName: foo has the variable name 'x1'
```

The reason for this behaviour is [here](https://github.com/troessner/reek/blob/master/lib/reek/smells/uncommunicative_variable_name.rb#L66):

```Ruby
      def bad_name?(name, _ctx)
        var = name.to_s.gsub(/^[@\*\&]*/, '')
        return false if accept_names.include?(var) # <- here is the problem!
        reject_names.find { |patt| patt =~ var }
      end
```

We only check for literals here, not for regexes. 
The other Uncommunicative* smell detectors are **not** affected by the issue since they rightfully check for literals and regexes.

troessner:
Fixes #1063

### DIFF
--- a/lib/reek/smells/uncommunicative_variable_name.rb
+++ b/lib/reek/smells/uncommunicative_variable_name.rb
@@ -13,9 +13,11 @@ module Reek
     # and they hurt the flow of reading, because the reader must slow
     # down to interpret the names.
     #
-    # Currently +UncommunicativeName+ checks for
-    # * 1-character names
-    # * names ending with a number
+    # Currently +UncommunicativeName+ checks for:
+    #
+    # * single-character names
+    # * any name ending with a number
+    # * camelCaseVariableNames
     #
     # See {file:docs/Uncommunicative-Variable-Name.md} for details.
     #
@@ -24,13 +26,17 @@ module Reek
       # The name of the config field that lists the regexps of
       # smelly names to be reported.
       REJECT_KEY = 'reject'.freeze
-      DEFAULT_REJECT_SET = [/^.$/, /[0-9]$/, /[A-Z]/].freeze
+      DEFAULT_REJECT_SET = [
+        /^.$/, # single-character names
+        /[0-9]$/,  # any name ending with a number
+        /[A-Z]/    # camelCaseVariableNames
+      ].freeze
 
       # The name of the config field that lists the specific names that are
       # to be treated as exceptions; these names will not be reported as
       # uncommunicative.
       ACCEPT_KEY = 'accept'.freeze
-      DEFAULT_ACCEPT_SET = ['_'].freeze
+      DEFAULT_ACCEPT_SET = [/^_$/].freeze
 
       def self.default_config
         super.merge(
@@ -51,7 +57,7 @@ module Reek
         self.reject_names = value(REJECT_KEY, ctx)
         self.accept_names = value(ACCEPT_KEY, ctx)
         variable_names(ctx.exp).select do |name, _lines|
-          bad_name?(name, ctx)
+          uncommunicative_variable_name?(name)
         end.map do |name, lines|
           smell_warning(
             context: ctx,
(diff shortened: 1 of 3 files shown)


lib/reek/smells/uncommunicative_variable_name.rb
```diff
@@ -24,7 +24,7 @@ class UncommunicativeVariableName < SmellDetector
       # The name of the config field that lists the regexps of
       # smelly names to be reported.
       REJECT_KEY = 'reject'.freeze
-      DEFAULT_REJECT_SET = [/^.$/, /[0-9]$/, /[A-Z]/].freeze
+      DEFAULT_REJECT_SET = [/^[a-z]$/, /[0-9]$/].freeze
```
troessner: This /[A-Z]/ would have never ended up here (since it would have been parsed differently) and this /^.$/ was a bug already in my mind. Should have been /[a-z]/ from the beginning.

lib/reek/smells/uncommunicative_variable_name.rb
```diff
@@ -61,10 +62,15 @@ def sniff(ctx)
...
-        var = name.to_s.gsub(/^[@\*\&]*/, '')
-        return false if accept_names.include?(var)
-        reject_names.find { |patt| patt =~ var }
+      private
+
+      def uncommunicative_variable_name?(name)
+        !acceptable_name?(name)
+      end
+
+      def acceptable_name?(name)
```
troessner: I know this seems unnecessary. The method body from here could go into `uncommunicative_variable_name` directly. But I find positive predicate method names easier to read and it's easier to build up a mental image of them.

lib/reek/smells/uncommunicative_variable_name.rb
```diff
@@ -124,8 +130,6 @@ def record_variable_name(exp, symbol, accumulator)
         accumulator[var].push(exp.line)
       end
 
-      private
```
troessner: All methods of this detector were public before :-/

lib/reek/smells/uncommunicative_variable_name.rb
mvz: I think `/[A-Z]/` is meant to catch stuff like `aVariable`.

lib/reek/smells/uncommunicative_variable_name.rb
```diff
@@ -61,10 +62,15 @@ def sniff(ctx)
...
-        return false if accept_names.include?(var)
-        reject_names.find { |patt| patt =~ var }
+      private
+
+      def uncommunicative_variable_name?(name)
+        !acceptable_name?(name)
+      end
+
+      def acceptable_name?(name)
+        Array(accept_names).any? { |accept_pattern| name.match accept_pattern } ||
```
troessner: This is now **almost** the same as we do in all other Uncommunicative* detectors with the exception of the "Array()" call (which we need if the user gives us literal values). 
I intend to remove the "Array()" call again when I tackle #1066

troessner:
Ready for review!

lib/reek/smells/uncommunicative_variable_name.rb
troessner: A yes! I'll add a spec (nothing broke on removal) and reintroduce it.

troessner:
Addressed all comments and added a spec for camelcasing! LGTM? :)

lib/reek/smells/uncommunicative_variable_name.rb
```diff
@@ -24,13 +26,17 @@ class UncommunicativeVariableName < SmellDetector
...
+        /[0-9]$/,  # any name ending with a number
+        /[A-Z]/    # camelCaseVariableNames
+      ].freeze
 
       # The name of the config field that lists the specific names that are
       # to be treated as exceptions; these names will not be reported as
       # uncommunicative.
       ACCEPT_KEY = 'accept'.freeze
-      DEFAULT_ACCEPT_SET = ['_'].freeze
+      DEFAULT_ACCEPT_SET = [/^_/].freeze
```
mvz: This makes us accept `_aBadName42`. I think the original intent was to accept `_` as the only valid single-character name.

lib/reek/smells/uncommunicative_variable_name.rb
```diff
@@ -24,13 +26,17 @@ class UncommunicativeVariableName < SmellDetector
       # The name of the config field that lists the regexps of
       # smelly names to be reported.
       REJECT_KEY = 'reject'.freeze
-      DEFAULT_REJECT_SET = [/^.$/, /[0-9]$/, /[A-Z]/].freeze
+      DEFAULT_REJECT_SET = [
+        /^[a-z]$/, # single-character names
```
mvz: Since Ruby allows loads of unicode characters in variable names, I think this is too narrow. This detector will now happily accept the variable name `é`.

troessner:
@mvz addressed all your comments! LGTM? :)

PR #1068, opened 2016-09-10
Make our CodeComment regex more lenient.

### LINKED ISSUE #1054: Error for certain malformed smell suppressions
A while back after reading the [Smell Suppression](https://github.com/troessner/reek/blob/master/docs/Smell-Suppression.md) doc I thought I would be clever and apply my new knowledge about smell suppression. I added the following comment:

```
# :reek:TooManyStatements { max_statements: 8 }
```

Fast forward to today: I'm grepping through our code and removing stale `:reek:*` comments. I get to this comment and check if I can lower the number. I change it to `# :reek:TooManyStatements { max_statements: 7 }` and reek still passes, `max_statements: 6`, `5`, `4`, etc.. until I hit zero.

After a moment of confusion I pull the docs up and realize the mistake. The correct original comment would have been

```
# :reek:TooManyStatements: { max_statements: 8 }
```

subtle! Here is the diff if you can't see it:

```diff
--- incorrect
+++ corrected
@@ -1 +1 @@
-# :reek:TooManyStatements { max_statements: 8 }
+# :reek:TooManyStatements: { max_statements: 8 }
```

Could reek help me in this situation? I imagine I'm not the only person that has done this.

troessner:
Fixes #1054 
This PR makes the ":" at the end of ":reek:UncommunicativeMethodName:" optional with 1 max

I'd consider the ":" at the end unnecessary and confusing syntax that a lot of users will get wrong.
Not to mention that a lot, a lot of people will forget about this when you go from the optionless form to the form with options like [we did ourselves](https://github.com/troessner/reek/blob/master/lib/reek/ast/node.rb#L89).

### DIFF
--- a/lib/reek/code_comment.rb
+++ b/lib/reek/code_comment.rb
@@ -7,10 +7,19 @@ module Reek
   # module, class and method definitions.
   #
   class CodeComment
-    CONFIGURATION_REGEX            = /:reek:(\w+)(:\s*\{.*?\})?/
+    CONFIGURATION_REGEX = /
+                          :reek: # prefix
+                          (\w+)  # smell detector e.g.: UncommunicativeVariableName
+                          (
+                            :? # legacy separator
+                            \s*
+                            (\{.*?\}) # optional details in hash style e.g.: { max_methods: 30 }
+                          )?
+                         /x
     SANITIZE_REGEX                 = /(#|\n|\s)+/ # Matches '#', newlines and > 1 whitespaces.
-    DISABLE_DETECTOR_CONFIGURATION = ': { enabled: false }'.freeze
+    DISABLE_DETECTOR_CONFIGURATION = '{ enabled: false }'.freeze
     MINIMUM_CONTENT_LENGTH         = 2
+    LEGACY_SEPARATOR               = ':'.freeze
 
     attr_reader :config
 
@@ -23,8 +32,8 @@ module Reek
       @original_comment  = comment
       @config            = Hash.new { |hash, key| hash[key] = {} }
 
-      @original_comment.scan(CONFIGURATION_REGEX) do |smell_type, options|
-        @config.merge! YAML.load(smell_type + (options || DISABLE_DETECTOR_CONFIGURATION))
+      @original_comment.scan(CONFIGURATION_REGEX) do |detector, _option_string, options|
+        @config.merge! detector => YAML.load(options || DISABLE_DETECTOR_CONFIGURATION)
       end
     end
 
--- a/README.md
+++ b/README.md
@@ -356,7 +356,7 @@ end
 You can even pass in smell specific configuration settings:
 
 ```Ruby
-# :reek:NestedIterators: { max_allowed_nesting: 2 }
+# :reek:NestedIterators { max_allowed_nesting: 2 }
 def smelly_method foo
   foo.each {|bar| bar.each {|baz| baz.qux}}
 end
--- a/docs/Smell-Suppression.md
+++ b/docs/Smell-Suppression.md
@@ -51,10 +51,10 @@ def many_parameters_it_has foo, bar, baz, qux
 end
 ```
 
-It is also possible to specify options for a particular smell detector, like so:
+It is also possible to specify options for a particular smell detector in hash-style:
 
 ```ruby
-# :reek:LongParameterList: { max_params: 4 }
+# :reek:LongParameterList { max_params: 4 }
 def many_parameters_it_has foo, bar, baz, qux
   # ...
 end
@@ -66,17 +66,17 @@ also use via comment.
 E.g.:
 
 ```ruby
-# :reek:TooManyStatements: { max_statements: 6 }
+# :reek:TooManyStatements { max_statements: 6 }
 def too_many
   # ...
 end
 
-# :reek:NestedIterators: { max_allowed_nesting: 2 }
+# :reek:NestedIterators { max_allowed_nesting: 2 }
 def quax
   foo.each {|bar| bar.each {|baz| baz.qux(qux)}}
 end
 
-# :reek:DuplicateMethodCall: { max_calls: 3 }
+# :reek:DuplicateMethodCall { max_calls: 3 }
 def quax
   foo.to_i + foo.to_i + foo.to_i
 end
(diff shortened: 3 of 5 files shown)


troessner:
Ready for eview!

troessner:
Ok, back to the drawing board. Removing the ":" causes "Yaml.load" from the [CodeComment initializer](https://github.com/troessner/reek/blob/master/lib/reek/code_comment.rb#L27) to do something very stupid. 
Behold:

```Ruby
smell_type =  "TooManyMethods"
options = ": { max_methods: 30 }"
YAML.load(smell_type + options)
# => {"TooManyMethods"=>{"max_methods"=>30}}
# Looking good!
# Not so good without the ":"
new_options = " { max_methods: 30 }"
YAML.load(smell_type + new_options)
# => {"TooManyMethods { max_methods"=>"30 }"}
# For sure not what we wanted!
```

troessner:
Review me baby one moar time! Now everything is green! Review pls ;)

lib/reek/code_comment.rb
```diff
@@ -24,6 +46,13 @@ def initialize(comment)
       @config            = Hash.new { |hash, key| hash[key] = {} }
 
       @original_comment.scan(CONFIGURATION_REGEX) do |smell_type, options|
+        # We need the legacy separator in front of the options string for
+        # the "Yaml.load" below to work properly. With the ":" in front of the options
+        # "Yaml.load" will return something like this:
+        #   {"TooManyMethods"=>{"max_methods"=>30}}
+        # Without it something bogus like this:
+        #  {"TooManyMethods { max_methods"=>"30 }"}
+        options.prepend(LEGACY_SEPARATOR) if options && !options.start_with?(LEGACY_SEPARATOR)
```
mvz: How about not putting the `:` inside the regex group and always prepending `:` here?

lib/reek/code_comment.rb
```diff
@@ -7,10 +7,32 @@ module Reek
   # module, class and method definitions.
   #
   class CodeComment
-    CONFIGURATION_REGEX            = /:reek:(\w+)(:\s*\{.*?\})?/
+    CONFIGURATION_REGEX = /
+                          # 1st group is the smell name e.g.:
```
mvz: I'm having a hard time distinguishing between code and comments here. I think this will be clearer if you put the large explanation above the constant, and the rest in trailing comments, like so:

```ruby
# Some text about what we're allowing with examples goes here.
CONFIGURATION_REGEX = /
  :reek:
  (\w+)         # smell name, e.g., UncommunicativeVariableName
  :?            # legacy separator
  \s*
  ( \{.*?\} )?  # optional details in hash style
  /x
```

I also have no idea what the first `?` is doing in the second group.

lib/reek/code_comment.rb
mvz: ... and we could even do something like this (untested), and avoid the `LEGACY_SEPARATOR` complications:

```ruby
@config.merge! smell_type => YAML.load(options || DISABLE_DETECTOR_CONFIGURATION)
```

mvz:
Review done!

lib/reek/code_comment.rb
troessner: >> How about not putting the : inside the regex group and always prepending : here?

Good idea!

>> ... and we could even do something like this (untested), and avoid the LEGACY_SEPARATOR complications:

That doesnt work - Yaml.load requires a string like it is right now. Using it just on options will crash Psych. Lets rather keep i like that.

mvz: > Using it just on options will crash Psych

Wow. Yeah, let's leave that part as it is then :laughing:.

troessner: Hmmm, I retract my statement. Let me double check.

lib/reek/code_comment.rb
troessner: >> I'm having a hard time distinguishing between code and comments here. I think this will be clearer if you put the large explanation above the constant, and the rest in trailing comments, like so:

Sure, I can change this.

>> I also have no idea what the first ? is doing in the second group.

You mean the 

>> ( \{.*?\} )?  # optional details in hash style

part?

That's for greedy matching, so in case you get stuff like

```
{ holla: waldfee } omg: bar }
````

just using "(.*)" would match the whole expression when we actually want the closest matching bracket - that's what "(.*?)" does :)

**Comment by troessner** [comment]
@mvz addressed all your comments, good suggestions from your side, looks way more elegant now.
Please re-review ;)

File: lib/reek/code_comment.rb [hunk 3]
```diff
@@ -7,10 +7,21 @@ module Reek
...
   class CodeComment
-    CONFIGURATION_REGEX            = /:reek:(\w+)(:\s*\{.*?\})?/
+    CONFIGURATION_REGEX = /
+                          :reek: # prefix
+                          (\w+)  # smell detector e.g.: UncommunicativeVariableName
+                          (
+                            (
+                              :? # legacy separator
+                              \s*
+                            )
```
mvz: Why make this a group?

lib/reek/code_comment.rb
```diff
@@ -23,8 +34,8 @@ def initialize(comment)
       @original_comment  = comment
       @config            = Hash.new { |hash, key| hash[key] = {} }
 
-      @original_comment.scan(CONFIGURATION_REGEX) do |smell_type, options|
-        @config.merge! YAML.load(smell_type + (options || DISABLE_DETECTOR_CONFIGURATION))
+      @original_comment.scan(CONFIGURATION_REGEX) do |detector, _option_string, _padding, options|
+        @config.merge! detector => YAML.load(options || DISABLE_DETECTOR_CONFIGURATION)
```
mvz: :+1:

lib/reek/code_comment.rb
```diff
@@ -7,10 +7,21 @@ module Reek
   # module, class and method definitions.
   #
   class CodeComment
-    CONFIGURATION_REGEX            = /:reek:(\w+)(:\s*\{.*?\})?/
+    CONFIGURATION_REGEX = /
+                          :reek: # prefix
+                          (\w+)  # smell detector e.g.: UncommunicativeVariableName
```
mvz: :+1:

mvz:
One little nit left, then :shipit:.

troessner:
Addressed your comments @mvz !

PR #1071, opened 2016-09-11
Quote names in smell detector messages.

### LINKED ISSUE #1046: Quote names in SmellDetector messages
Good:

UnusedParameters:

"has unused parameter 'xyz'"

Bad:

FeatureEnvy:

"refers to xyz more than self"

All names should be single-quoted like in UnusedParameters so that it is clear to the user that now we're talking about something in his source code.

troessner:
Fixes #1046


troessner:
Ready for review!

lib/reek/smells/data_clump.rb
```diff
@@ -59,7 +59,7 @@ def sniff(ctx)
           smell_warning(
             context: ctx,
             lines: methods.map(&:line),
-            message: "takes parameters #{DataClump.print_clump(clump)} " \
+            message: "takes parameters '#{DataClump.print_clump(clump)}' " \
```
mvz: This one may be odd. We may need to quote the individual clump members instead.

troessner: Good point, updated!

mvz:
There's a build error that's due to a typo I think, otherwise LGTM.

troessner:
>> There's a build error that's due to a typo I think, otherwise LGTM.

Fixed!
Taking the liberty of merging this myself so we can release all of the latest pull requests.

PR #1076, opened 2016-09-12
Validate the configuration in total before running.

### LINKED ISSUE #1062: Validate the configuration in total before executing any smell detectors.
Coming from #1058 - the way we catch exceptions in our smell detectors actually masks errors users might make in Reeks configuration.

We should definitely fix this by validating the configuration in total before executing any smell detectors.
I'd consider this as rather urgent since it's very easy to fuck up and then be left with no clue what actually went wrong.

troessner:
Fixes #1062 - heavily wip.


troessner:
I'll look into the comment configuration validation first, closing this one for now.

Issue #1069, opened 2016-09-10
Problems with the way we handle CodeComments


troessner:
I see 2 huge problems at the moment:

1.) Something like

>>
      # :reek:DuplicateMethodCall: { wtf is this!? }

is a **valid** expression looking at our [comment regex](https://github.com/troessner/reek/blob/master/lib/reek/code_comment.rb#L10)

2.) The comment above will lead to a Runtime error like this:

```
[3] pry(#<Reek::CodeComment>)> c
      !!!
      Source lib/reek/ast/node.rb can not be processed by Reek.
      This is most likely a Reek bug.
      It would be great if you could report this back to the Reek team
      by opening up a corresponding issue at https://github.com/troessner/reek/issues
      Make sure to include the source in question, the Reek version
      and the original exception below.

      Original exception:
      #<Psych::SyntaxError: (<unknown>): did not find expected ',' or '}' while parsing a flow mapping at line 1 column 22>
      !!!
```

Obviously this is a catastrophe from a users perspective. Nobody will get this.
This behaviour is a consequence of #996 which we completely missed when discussing the pull request  👎 

I suggest the following:

* make the comment regex more robust
* do NOT catch errors that emerge from configuration handling in our smell detectors but let them bubble up analogous to #1039 
* post-process those kind of exceptions in a way that makes them more meaningful for the users.

WDYT?

mvz:
Yes, that sounds good. We should in general make a difference between specific Reek-internal errors and the no-idea-what-just-happened-I-guess-we-broke-stuff kind. I have some ideas about this so I'll make a pull request.

troessner:
A first sketch of what I have in mind:

* We pass in the context to [CodeComments](https://github.com/troessner/reek/blob/master/lib/reek/code_comment.rb) initializer
* In this initializer part:

```Ruby
@original_comment.scan(CONFIGURATION_REGEX) do |smell_type, options|
  @config.merge! YAML.load(smell_type + (options || DISABLE_DETECTOR_CONFIGURATION))
end
```

we validate if we can create a valid SmellDetector out of the given options
* if we can't we raise a Reek::ConfigException that takes the calling context into account to present the user with a helpful error message
* We patch [Examiner#run](https://github.com/troessner/reek/blob/master/lib/reek/examiner.rb#L103) so that it doesn't catch Reek::ConfigExceptions
-> Done!

How does that sound to you guys?
I'd like to sketch out the overall plan first before addressing the biggest problem: How to do this:

[...]

PR #1080, opened 2016-09-13
Properly feature test configuration via source comments.

troessner:
I know we want to cut down on cucumber features but for proper integration testing that tests the full stack a la:

"creating source file" -> "add configuration" -> "run Reek" -> "validate output"

cucumber is and probably will be our weapon of choice for the time to come.

I believe that configuration via source comments was not properly tested at all except for a couple of spot checks in our detector specs here and there, but that was is it.
This is an attempt to fix this.
A fast review and merge would be appreciated because this is the base for my current work on #1069


troessner:
Anyone up for merging this bad boy? Should be none controversial :)

PR #1082, opened 2016-09-13
Fail properly on bad configuration comments.

### LINKED ISSUE #1069: Problems with the way we handle CodeComments

troessner:
Fixes #1069 - heavily wip.

### DIFF
--- a/lib/reek/code_comment.rb
+++ b/lib/reek/code_comment.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 require 'yaml'
+require_relative 'smells/smell_detector'
+require_relative 'errors'
 
 module Reek
   #
@@ -25,14 +27,20 @@ module Reek
 
     #
     # @param comment [String] - the original comment as found in the source code
-    #   E.g.:
-    #   "\n        # :reek:Duplication: { enabled: false }\n      "
+    # @param line [Integer] - start of the expression the comment belongs to
+    # @param source [String] - Path to source file or "string"
     #
-    def initialize(comment)
+    def initialize(comment:, line: nil, source: nil)
       @original_comment  = comment
       @config            = Hash.new { |hash, key| hash[key] = {} }
 
       @original_comment.scan(CONFIGURATION_REGEX) do |detector, _option_string, options|
+        unless Smells::SmellDetector.valid_detector?(detector)
+          raise BadDetectorInCommentError, detector: detector,
+                                           source: source,
+                                           line: line,
+                                           original_comment: @original_comment
+        end
         @config.merge! detector => YAML.load(options || DISABLE_DETECTOR_CONFIGURATION)
       end
     end
--- a/lib/reek/context/module_context.rb
+++ b/lib/reek/context/module_context.rb
@@ -64,7 +64,7 @@ module Reek
       end
 
       def descriptively_commented?
-        CodeComment.new(exp.leading_comment).descriptive?
+        CodeComment.new(comment: exp.leading_comment).descriptive?
       end
 
       # A namespace module is a module (or class) that is only there for namespacing
--- a/lib/reek/smells/smell_detector.rb
+++ b/lib/reek/smells/smell_detector.rb
@@ -72,9 +72,8 @@ module Reek
       def smell_warning(options = {})
         context = options.fetch(:context)
         exp = context.exp
-        ctx_source = exp.loc.expression.source_buffer.name
         SmellWarning.new(self,
-                         source: ctx_source,
+                         source: exp.source,
                          context: context.full_name,
                          lines: options.fetch(:lines),
                          message: options.fetch(:message),
@@ -102,8 +101,26 @@ module Reek
           descendants << subclass
         end
 
+        #
+        # Returns all descendants of SmellDetector
+        #
+        # @return [Array<Constant>], e.g.:
+        #   [Reek::Smells::Attribute,
+        #    Reek::Smells::BooleanParameter,
+        #    Reek::Smells::ClassVariable,
+        #    ...]
+        #
         def descendants
           @descendants ||= []
+        end
+
+        #
+        # @param detector [String] the detector in question, e.g. 'DuplicateMethodCall'
+        # @return [Boolean]
+        #
+        def valid_detector?(detector)
+          descendants.map { |descendant| descendant.to_s.split('::').last }.
+            include?(detector)
         end
       end
     end
--- a/spec/reek/code_comment_spec.rb
+++ b/spec/reek/code_comment_spec.rb
@@ -1,12 +1,15 @@
 require_relative '../spec_helper'
 require_lib 'reek/code_comment'
+require_lib 'reek/errors'
 
 RSpec.describe Reek::CodeComment do
   context 'with an empty comment' do
-    let(:comment) { described_class.new('') }
+    let(:comment) { FactoryGirl.build(:code_comment, comment: '') }
+
     it 'is not descriptive' do
       expect(comment).not_to be_descriptive
     end
+
     it 'has an empty config' do
       expect(comment.config).to be_empty
     end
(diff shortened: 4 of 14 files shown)


troessner:
Hmmm, I'm a little stuck with this one. The problem is that when I put the validation into CodeComments initializer then the error message gets reported multiple times - once for every context.
E.g. this innocent example:

```Ruby
# :reek:UnknownSmellDetector
def x
end
```

would cause the same output on stderr twice because the code sample would be parsed into a RootContext that contains a MethodContext.

Off the top of my head I can think of 3 possible solutions:

1.) We kind of check what context it makes sense to output something. RootContext is probably something that should never write anything on stderr. But that feels very hacky and again our overall design.
2.) Do the validation in `TreeDresser.new.dress`. That would work well I believe. But then we have the problem that the comment creation and validation are done in different places. Pretty uncool. 
3.) We kind of buffer everything we want to write to stderr somewhere, somehow and then only write the unique elements of this buffer to stderr at the very end.

I'm kind of leaning towards [3] but I'd love to hear your ideas about this @mvz && @waldyr  ;)

troessner:
Come to think about this, such a buffering mechanism might come in handy for all kinds of things. We could basically buffer all-the-things to stderr including our error handling in the Examiner.
And then, at the very end, we could put out one concise summary.

mvz:
I'm leaning towards just aborting the examination of that file at the first error and printing a more specific message in our exception handler.

lib/reek/code_comment.rb
```diff
@@ -20,19 +21,42 @@ class CodeComment
...
+    # @param source [String] - Path to source file or "string"
     #
-    def initialize(comment)
+    def initialize(comment:, line:, source:)
       @original_comment  = comment
       @config            = Hash.new { |hash, key| hash[key] = {} }
 
       @original_comment.scan(CONFIGURATION_REGEX) do |detector, _option_string, options|
+        unless Smells::SmellDetector.valid_detector?(detector)
+          $stderr.puts format(UNKNOWN_SMELL_DETECTOR_MESSAGE,
```
mvz: Ah I see what your approach is. Rather than having output to `$stderr` all over the place, let's just raise a custom exception here that we catch further up the call stack.

lib/reek/code_comment.rb
```diff
@@ -20,19 +21,42 @@ class CodeComment
...
+    # @param line [Integer] - start of the expression the comment belongs to
+    # @param source [String] - Path to source file or "string"
     #
-    def initialize(comment)
+    def initialize(comment:, line:, source:)
       @original_comment  = comment
       @config            = Hash.new { |hash, key| hash[key] = {} }
 
       @original_comment.scan(CONFIGURATION_REGEX) do |detector, _option_string, options|
+        unless Smells::SmellDetector.valid_detector?(detector)
```
mvz: Can we somehow avoid this dependency on a completely different part of the code? Perhaps we can move this check up the call stack a bit?

troessner:
>> I'm leaning towards just aborting the examination of that file at the first error 

I like the sound of that, but aborting the examination of that file would require us to jump up **really** high on Examiner level again. And then catch this [here](https://github.com/troessner/reek/blob/master/lib/reek/cli/command/report_command.rb#L24) at ReportCommand and then again, at [TodoCommand](https://github.com/troessner/reek/blob/master/lib/reek/cli/command/todo_list_command.rb#L31). 
And in `ShouldReek#matches?` and so on.
So that won't work.
Or did you have something else in mind?

mvz:
> Or did you have something else in mind?

I have in mind to handle it at the Examiner level. So [here](https://github.com/troessner/reek/blob/master/lib/reek/examiner.rb#L103) we would first catch `Reek::BadConfigurationError`, and then `StandardError`.

In `ShouldReek#matches?` it would just not be handled like all the other exceptions.

At a later stage, I'd like to extract all the exception handling in Examiner to a special handler object that we can pass to the Examiner at initialization, but that's a plan that still needs some simmering.

troessner:
>> I have in mind to handle it at the Examiner level. So here we would first catch Reek::BadConfigurationError, and then StandardError.
In ShouldReek#matches? it would just not be handled like all the other exceptions.

Good idea! I'm going to look into this tomorrow :)

lib/reek/smells/smell_detector.rb
```diff
@@ -105,6 +104,14 @@ def inherited(subclass)
         def descendants
           @descendants ||= []
         end
+
+        #
+        # @param detector [String] the detector in question, e.g. 'DuplicateMethodCall'
+        # @return [Boolean]
+        #
+        def valid_detector?(detector)
+          Reek::Smells.const_get(detector) rescue false
```
waldyr: Shouldn't we rescue only NameError?

backus: Could this just be `Reek::Smells.const_defined?(detector)`

lib/reek/code_comment.rb
```diff
@@ -20,19 +21,42 @@ class CodeComment
     DISABLE_DETECTOR_CONFIGURATION = '{ enabled: false }'.freeze
     MINIMUM_CONTENT_LENGTH         = 2
     LEGACY_SEPARATOR               = ':'.freeze
+    UNKNOWN_SMELL_DETECTOR_MESSAGE = <<-EOS.freeze
```
backus: If you wrap `EOS` in single quotes then it will disable string interpolation FYI (equivalent to a string with single quotes)

lib/reek/code_comment.rb
backus: I'm curious. Why you prefer `$stderr.puts` over `warn`?

lib/reek/code_comment.rb
troessner: I don't follow -  why do you think wrapping EOS here makes sense?

lib/reek/code_comment.rb
troessner: The only reason was for consistency reason since we use `stderr.puts` throughout Reek so far. But you're right, `warn` is more succinct and consistency in this case overrated, thanks for pointing that out!

lib/reek/code_comment.rb
troessner: >> Can we somehow avoid this dependency on a completely different part of the code? 

Hmmm, but it's actually modelling what we do, isn't it? We create a comment here and this comment contains the description for a smell detector. So we let SmellDetector validate it right here. To me this seems like the responsibilities are distributed properly...

>> Perhaps we can move this check up the call stack a bit?

But then we would separate creation from validation. To me they belong together. Do you have an example for what you have in mind?

lib/reek/smells/smell_detector.rb
troessner: Good point, let me look into this.

troessner:
Ladies and gentlemen, after incorporating all of your comments this is ready for re-review!

lib/reek/code_comment.rb
mvz: > But then we would separate creation from validation.

Yes, this would be an external validation. In general, I tend to avoid having internal validations if it creates too many dependencies.

However, it's probably best to go with this implementation for now, and then I can see if I can refactor it later and we can discuss this more then.

lib/reek/code_comment.rb
```diff
@@ -25,14 +27,20 @@ class CodeComment
...
       @original_comment  = comment
       @config            = Hash.new { |hash, key| hash[key] = {} }
 
       @original_comment.scan(CONFIGURATION_REGEX) do |detector, _option_string, options|
+        unless Smells::SmellDetector.valid_detector?(detector)
+          raise BadDetectorInCommentError, detector: detector,
+                                           source: source,
+                                           line: line,
+                                           original_comment: @original_comment
+        end
```
mvz: I think @waldyr's check is further up in `SmellDetector`, but it would use the same mechanism.

lib/reek/smells/smell_detector.rb
```diff
@@ -105,6 +104,14 @@ def inherited(subclass)
         def descendants
           @descendants ||= []
         end
+
+        #
+        # @param detector [String] the detector in question, e.g. 'DuplicateMethodCall'
+        # @return [Boolean]
+        #
+        def valid_detector?(detector)
+          Reek::Smells.const_defined?(detector)
```
mvz: This makes 'SmellWarning' a valid detector.

spec/reek/code_comment_spec.rb
```diff
@@ -1,103 +1,129 @@
...
-      ').config
-      expect(config).to include('Duplication', 'NestedIterators')
-      expect(config['Duplication']).to include('enabled')
-      expect(config['Duplication']['enabled']).to be_falsey
-      expect(config['NestedIterators']).to include('enabled')
-      expect(config['NestedIterators']['enabled']).to be_truthy
+      EOF
+      comment = FactoryGirl.build(:code_comment, comment: original_comment)
+
+      expect(comment.config).to include('DuplicateMethodCall', 'NestedIterators')
```
mvz: I'd prefer to avoid the duplicate `comment.config` here and in the other specs in this file by assigning to a local variable (as it was before).

lib/reek/context/module_context.rb
```diff
@@ -64,7 +64,9 @@ def node_instance_methods
       end
 
       def descriptively_commented?
-        CodeComment.new(exp.leading_comment).descriptive?
+        CodeComment.new(comment: exp.leading_comment,
+                        line: exp.line,
+                        source: exp.source).descriptive?
```
troessner: Ack.

lib/reek/smells/smell_detector.rb
troessner: Good catch! Good thing we have `SmellDetector#descendants` :)

spec/reek/code_comment_spec.rb
troessner: Fair enough, updated.

lib/reek/code_comment.rb
backus: I was just pointing out helpful syntax that I think many people aren't aware of. Basically you can write something like

```ruby
<<-ANY_HEREDOC
foo
bar
baz
ANY_HEREDOC
```

with single quotes wrapping the heredoc end of string token like so

```ruby
<<-'ANY_HEREDOC'
foo
bar
baz
ANY_HEREDOC
```

the benefit of this is that it signals that interpolation doesn't happen in the entire heredoc

troessner: Fair enough - thanks for the tip! I was indeed not aware of this ;)
I'll keep this in mind for the future. If we introduce this syntax into Reek we probably should update all heredocs to be consistent from the start, which would warrant its own PR.

backus: Fair enough

troessner:
Addressed all comments! LGTM? :)

troessner:
Pinging @mvz - sorry for being pushy but I'd like to continue with checking the config hash itself which will be build on top of this PR..

mvz:
@troessner no problem, this one slipped under the radar today. Looks good, merging!

troessner:
And I just released 4.4.2 with those changes!
Another small step to a Reek that is more resilient to bad input :)

PR #1084, opened 2016-09-15
Add missing links in documentations

pocke:


troessner:
Good catch, merged, thanks!

PR #1086, opened 2016-09-16
Make tests pass when run other than from a tty

mvz:
This may look a little weird, but on master the following fails:

```
echo 'foo' | bundle exec rake
```

This change fixes that.


troessner:
A "little" weird? :)

PR #1089, opened 2016-09-17
Improve how we handle internal error in detectors.

### LINKED ISSUE #1088: Improvements to the way we currently handle internal Reek errors in detectors
Related to #1069:

#996 will help a lot making Reek more resilient but as already outlined in #1069 we kind of forgot that the majority of all errors the user makes in either the configuration or the source code comments that would have caused Reek to crash before will now result in this:

```
      !!!
      Source lib/reek/ast/node.rb can not be processed by Reek.
      This is most likely a Reek bug.
      It would be great if you could report this back to the Reek team
      by opening up a corresponding issue at https://github.com/troessner/reek/issues
      Make sure to include the source in question, the Reek version
      and the original exception below.

      Original exception:
      #<Psych::SyntaxError: (<unknown>): did not find expected ',' or '}' while parsing a flow mapping at line 1 column 22>
      !!!
```

Which is highly confusing for the user and might lead to a lot of false-positives issues in the long run (like: "Reek told me to report this here, what now?").

Of course the best solution to this problem would be [validate all configuration](https://github.com/troessner/reek/issues/1062) before running (or in case of comments while Reek is running and then print sth on STDERR or similar) which is something we are actively [working on](https://github.com/troessner/reek/pull/1082). But this might take a while since the problem is not trivial.

So I was wondering what we could do in the meantime and I have 2 suggestions:

**1.) Improve the error message above**

Instead of

````
      This is most likely a Reek bug.
      It would be great if you could report this back to the Reek team
      by opening up a corresponding issue at https://github.com/troessner/reek/issues
      Make sure to include the source in question, the Reek version
      and the original exception below.
```

I'd suggest to print:

````
This is most likely either a bug in your Reek configuration (config file or 
source code comments) or a Reek bug.

Please double check your Reek configuration taking the original exception
below into account -  you might have mispelled a smell detector for instance.
(In the future Reek will handle configuration errors more gracefully, something
we are working on already).
      
If you feel that this is not a problem with your Reek configuration but with
Reek itself it would be great if you could report this back to the Reek
team by opening up a corresponding issue at https://github.com/troessner/reek/issues

Make sure to include the source in question, the Reek version
and the original exception below.
```

**2.) Fix the printing of the original exception**

That was another thing we missed in our review :)
As you can see above, we're only printing out the very last line. This should be a full stacktrace to make it easier on the user and on us.

WDYT?

troessner:
Fixes #1088

### DIFF
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -12,14 +12,30 @@ module Reek
     INCOMPREHENSIBLE_SOURCE_TEMPLATE = <<-EOS.freeze
       !!!
       Source %s can not be processed by Reek.
-      This is most likely a Reek bug.
-      It would be great if you could report this back to the Reek team
-      by opening up a corresponding issue at https://github.com/troessner/reek/issues
-      Make sure to include the source in question, the Reek version
+
+      This is most likely either a bug in your Reek configuration (config file or
+      source code comments) or a Reek bug.
+
+      Please double check your Reek configuration taking the original exception
+      below into account -  you might have misspelled a smell detector for instance.
+      (In the future Reek will handle configuration errors more gracefully, something
+      we are working on already).
+
+      If you feel that this is not a problem with your Reek configuration but with
+      Reek itself it would be great if you could report this back to the Reek
+      team by opening up a corresponding issue at https://github.com/troessner/reek/issues.
+
+      Please make sure to include the source in question, the Reek version
       and the original exception below.
 
-      Original exception:
+      Exception message:
+
       %s
+
+      Original exception:
+
+      %s
+
       !!!
     EOS
     #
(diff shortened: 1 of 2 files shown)


mvz:
See my comments on #1088.

troessner:
@mvz any objections to merging this one? I read your comments on #1088 but I don't think they really apply to this PR since this PR doesn't change anything in the bigger picture. It basically just improves the existing, potentially highly misleading error message.

mvz:
With #1082 done, do we still need the change in the error text?

troessner:
Yes, we do, because there are still a lot of possibilities to shoot yourself in the foot with Reek telling you to "please report a bug":

1.) #1095 
2.) #1062 -> that one is not really specific but right now we only validate our configuration file partially, so again, a lot of ways to get a confusing error message.

We can revert this PR after [1] and [2] have been implemented but that might be quite a while in the future (especially [2]) so we should merge this one now.

mvz:
Ok, that makes sense. If you can look at the build failures and/or rebase this on master, I'll merge.

troessner:
Fixed, build should be green in a second!

PR #1090, opened 2016-09-17
Make parameter handling in detectors consistent.

### LINKED ISSUE #1047: Parameter handling in SmellDetectors not consistent
Most of the time whatever is in `parameters` is a string but sometimes we use a symbol which is not only highly confusing but also leads to brittle tests like this one for UnusedPrivateMethod:

```Ruby
                           message: "has the unused private instance method `m`",
                           name:    :m) # NOT 'm'
```

Should be string everywhere.

troessner:
Fixes #1047


troessner:
Ready for review!

PR #1099, opened 2016-09-24
Validate detector config in source comments.

### LINKED ISSUE #1095: Validate the config hash in source comments
Follow-up to #1082 
Now with the initial set up out of the way we need to go the next step and validate the config hash in source comments.
We right now would catch

```Ruby
# :reek:UnknownDetector
```

but not:

```Ruby
 # :reek:DuplicateMethodCall: { wtf is this!? }
```

After that we just need to plug the holes when it comes to our configuration file (there are some checks but they are not exhaustive) and we're golden! :)

troessner:
Fixes #1095 - heavily wip.


troessner:
Outdated PR, closing this one :)

PR #1105, opened 2016-09-28
Warn about multiple configuration files.

troessner:


README.md
```diff
@@ -352,30 +352,6 @@ configurations you can also check out [the `default.reek` file in this repositor
...
-
-```Ruby
-# :reek:NestedIterators { max_allowed_nesting: 2 }
-def smelly_method foo
-  foo.each {|bar| bar.each {|baz| baz.qux}}
-end
-```
-
-This is an incredible powerful feature and further explained under [Smell Suppresion](docs/Smell-Suppression.md).
-
```
troessner: I just moved this whole paragraph one paragraph for better reading flow.

mvz:
If we need this in the README, shouldn't we let Reek emit a warning if it encounters two `.reek` files in the same directory?

troessner:
>> If we need this in the README, shouldn't we let Reek emit a warning if it encounters two .reek files in the same directory?

Great idea!
I'll take care of this in this PR and nudge you when it's ready, putting this one back to wip ;)

troessner:
Almost there! 
2 things left to do:

1.) Show the file names with the relative path to the root directory. This will make the failing cucumber feature [here](https://github.com/troessner/reek/pull/1105/files#diff-1687ba999df56d688fbd6789ceacd73dR21) pass
2.) Increase test coverage for ConfigurationFileFinder.find_by_dir -> this will make simplecov happy.

lib/reek/cli/application.rb
```diff
@@ -36,15 +37,15 @@ def execute
       def configure_options(argv)
         Options.new(argv).parse
       rescue OptionParser::InvalidOption => error
-        $stderr.puts "Error: #{error}"
-        exit Options::DEFAULT_ERROR_EXIT_CODE
+        warn "Error: #{error}"
```
troessner: Boyscout rule ;)

lib/reek/configuration/configuration_file_finder.rb
```diff
@@ -17,42 +17,102 @@ class ConfigFileException < StandardError; end
     # The order in which ConfigurationFileFinder tries to find such a
     # configuration file is exactly like above.
     module ConfigurationFileFinder
-      module_function
```
troessner: Replaced module_function with class << self below because module_function + private is just a fucking mess in ruby. Seriously. Hopefully this gets fixed in Ruby 3.
The whole diff is a little messy unfortunately, might be easier to look at the whole file.

troessner:
Ready for review!

lib/reek/configuration/configuration_file_finder.rb
mvz: Maybe we should settle on either `class << self` or `def self.something_something`. I've a slight preference for the latter, it also seems to be the method used most in Reek, and it would mean a nicer diff here :smile:. WDYT?

spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -32,8 +32,9 @@
     end
 
     it 'returns the file even if it’s just ‘.reek’' do
-      found = described_class.find(current: CONFIG_PATH)
-      expect(found).to eq(CONFIG_PATH.join('.reek'))
+      single_configuration_file_dir = CONFIG_PATH.join('single_configuration_file')
+      found = described_class.find(current: single_configuration_file_dir)
+      expect(found).to eq(single_configuration_file_dir.join('.reek'))
```
mvz: Maybe you should also add a specific spec for the case of two configuration files found? I think this would force you to not exit immediately but instead throw an error, but that would probably be a good thing.

lib/reek/configuration/configuration_file_finder.rb
```diff
@@ -17,42 +17,102 @@ class ConfigFileException < StandardError; end
...
-          raise ConfigFileException, "Invalid configuration file \"#{path}\" -- Not a hash"
+        #
+        # Writes a proper warning message to STDERR and then exits the program.
+        #
+        # @return [undefined]
+        #
+        def escalate_too_many_configuration_files(found, directory)
+          offensive_files = found.map { |file| "'#{file.basename}'" }.join(', ')
+          warn format(TOO_MANY_CONFIGURATION_FILES_MESSAGE, offensive_files, directory)
+          exit 1
```
mvz: I would prefer raising here and catching further up, similar to how we handle other config errors (except that for this one, handling means we exit, of course).

lib/reek/configuration/configuration_file_finder.rb
troessner: Let's go with `class << self` then. That way we'll also never fuck up the visibility (I **always** forget that `class C; private; def self.bar; end; end` will not do what I want).

Deal?

lib/reek/configuration/configuration_file_finder.rb
troessner: But the exceptions we raise so far are different since they always happen in the scope of one single [Examiner](https://github.com/troessner/reek/blob/master/lib/reek/examiner.rb#L123). And in this scope it makes sense to let exceptions bubble up to catch them in one central place.

This is not the case here, at least I don't see it.
Put it differently: Why do you think it makes sense to throw an exception here only to let it bubble up and where would you actually handle the exception?

spec/reek/configuration/configuration_file_finder_spec.rb
troessner: Added two specs, one for the warning and one for the system exit.

troessner:
Addressed all comments -> merge? :)

lib/reek/configuration/configuration_file_finder.rb
mvz: I generally dislike putting `exit` somewhere deep down in the bowels of the code. But let's leave it like this now.

PR #1106, opened 2016-09-29
Create a fresh todo file on every run.

### LINKED ISSUE #916: Tweak the behaviour of how we handle the --todo flag
Coming from #914:

>> I've noticed another strange behavior, though. It seems that when generating the todo file, reek actually respects an existing reek file. Is this intended?

It's obvious that we should tweak the behaviour of how we handle the --todo flag since it's confusing and / or not properly explained.

### I believe we have 2 use cases:

1.) There is already an existing configuration file. 

In this case we could ask the user:

>>
We found the existing configuration file xyz.reek and will take this into account.
Should we append the output of --todo to this file? 
In case of "no" we'll just write the results to STDOUT 
[Yes/no/abort]

In case the user answers "yes", we'll just append the results to the existing configuration file using the existing configuration file with a marker above it like this

```Yaml
    ### Autogenerated by `reek --todo -c xyz.reek` ###
```

In case of "no", we'll just write it to STDOUT and in case of "abort" we'll....just abort.  

2.) There is not existing configuration file

In this case, the behaviour stays as it is right now.

WDYT?

troessner:
Fixes #916 - ready for review!

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,13 +404,23 @@ via:
 reek -c whatever/.todo.reek lib/
 ```
 
-Note that if you re-run
+It's important to understand that the number one use case of the `--todo` flag
+is to be run once at the beginning of the introduction of Reek to ease the transition.
+If you find yourself re-running Reek with the `--todo` flag in order to silence new warnings
+you're defeating the purpose of both the `--todo` flag and of Reek itself.
+
+As a consequence, running Reek with the `--todo` flag again will overwrite '.todo.reek' without
+asking (should not be a problem since this file is supposed to be versioned) and
+without taking **any** other configuration file you might have into account.
+
+This means that when you run
 
 ```Bash
-reek --todo lib/
+reek -c other_configuration.reek --todo lib/
 ```
 
-'.todo.reek' will get overwritten with a possibly updated configuration.
+`other_configuration.reek` will simply be ignored (as outlined before, Reek
+is supposed to have one configuration file and one file only).
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -404,13 +404,29 @@ via:
 reek -c whatever/.todo.reek lib/
 ```
 
-Note that if you re-run
+It's important to understand that the number one use case of the `--todo` flag
+is to be run once at the beginning of the introduction of Reek to ease the transition.
+If you find yourself re-running Reek with the `--todo` flag in order to silence new warnings
+you're defeating the purpose of both the `--todo` flag and of Reek itself.
+
+As a consequence, this
```
mvz: I would remove the full invocation below, and put something like 'running Reek with the `--todo` flag again'.

README.md
```diff
@@ -404,13 +404,29 @@ via:
...
 
-'.todo.reek' will get overwritten with a possibly updated configuration.
+will overwrite '.todo.reek' without asking (should not be a problem since this file is
+supposed to be versioned) and without taking **any** other configuration file you might
+have into account.
+
+This means that when you run
+
+```Bash
+reek -c config other_configuration.reek --todo lib/
```
mvz: I think the `-c config other_configuration.reek` isn't quite correct.

mvz:
Review done!

README.md
troessner: Good catch, fixed!

README.md
troessner: Ack, fixed.

troessner:
Addressed all comments - lgtm? :)

mvz:
More tweaks of the README may result from our discussion in #916, but looks good for now. Merging.

PR #1109, opened 2016-10-03
Refactor reports.

troessner:
Looong flight to San Francisco, first pull request ;)

This restructures everything around reports and formatters:

* Splits up all the report classes contained in report.rb into separate files
* Moves all formatters (except the code_climate one, see below) into report/formatter
* Puts all reports that are special in the sense that they consist of more than
  just a report class into a separate folder:
  * html_report (because of the template)
  * code_climate (because the formatter is 100% code_climate specific so it should not
    end up in report/formatter and then there is also the yml configuration)
* Given a "foo" directory with the files "bar.rb" and "baz.rb" there will always be a
  "foo.rb" on the same level as the directory which will require "bar" and "baz"


.rubocop.yml
```diff
@@ -69,6 +71,16 @@ RSpec/VerifiedDoubles:
     - 'spec/reek/context/method_context_spec.rb'
     - 'spec/reek/context/module_context_spec.rb'
 
+# FIXME: This seems to be a rubocop-rspec bug. The warning says:
+#   spec/reek/report/code_climate/code_climate_report_spec.rb:8:1: C: RSpec/FilePath: Spec path should end with reek/report/code_climate_report*_spec.rb
+#   RSpec.describe Reek::Report::CodeClimateReport do
```
mvz: I think this means it expects a CodeClimate namespaces to go with the `code_climate` directory.

troessner:
Updated and rebased!

mvz:
I think I mostly agree with the resulting location, but I'm not convinced about having stuff like CodeClimateReport in `code_climate/code_climate_report.rb`. What's the advantage over having it in its 'logical' file?

troessner:
>> What's the advantage over having it in its 'logical' file?

Hmmm, the basic idea was to group reports that have dependent files like html and codeclimate into directories to prevent everything from floating around on the same level.
I found that highly annoying that for instance the erb template was on the same directory level like the html_report.
Do you have another idea how to fix that?

mvz:
> Do you have another idea how to fix that?

Perhaps put the report class in `html_report.rb`, and the template in `html_report/template.html.erb` (or whatever its name is now :smile:).

troessner:
>> Perhaps put the report class in html_report.rb, and the template in html_report/template.html.erb (or whatever its name is now 😄).

Updated. I'd leave the codeclimate stuff as it is right now though - deal?

mvz:
> Updated. I'd leave the codeclimate stuff as it is right now though - deal?

Well, I suppose we can always move it again later :smile:.

Issue #1102, opened 2016-09-26
A more structured approach to "how to validate stuff in source code comments"


troessner:
I'd like to approach the "how to validate stuff in source code comments" more structured and analytical just for once ;)

**Requirements for validation:**

1.) Valid smell detector -> we already do that
2.) No garbage in the config hash which would in turn make `Yaml.load` crash. That's ready for review in #1101 
3.) Validate config keys for a given, valid detector:
3.1) The basic smell detector options
3.2) The detector specific options
4.) Valid datatypes for a given key

So far, [1] and [2] haven been done in CodeComment, but going down that road further will become unmaintainable pretty fast.
The problem is that the CodeComment is actually not **only** validating a comment but also SmellDetectors and configuration and what YAML loads and so on.

I suggest a separate class for doing this called CodeCommentValidator with the following API:

``` Ruby
def initialize(comment:, line:, source:)
end

# Returns true / false and a message if comment is invalid
# @return [Boolean, String]
def valid?
end
```

Usage could then look like this:

``` Ruby
validator = CodeCommentValidator.new(comment: comment, line: 5, source: 'string')
success, message = validator.valid?
raise BadCommentConfiguration, message unless success
```

And we could then catch only BadCommentConfiguration in Examiner as we do already.

WDYT?

troessner:
>> I think CodeCommentValidator can just raise the error by itself, so we only need a #validate method 

Good idea! I'll come up with a wip PR later and then we can discuss the API.

>> I think CodeCommentValidator can just raise the error by itself, so we only need a #validate method 

Link pls?

mvz:
> Link pls?

Burn your getters: http://media.eurucamp.org/eurucamp/2013/jcoglan / https://github.com/jcoglan/burn-your-getters/tree/master

@chastell's talk: https://www.youtube.com/watch?v=pazYe7WRWRU

troessner:
Awesome, thanks!

troessner:
Quick update regarding:

>>
1.) Valid smell detector -> we already do that
2.) No garbage in the config hash which would in turn make Yaml.load crash. That's ready for review in #1101 
3.) Validate config keys for a given, valid detector:
3.1) The basic smell detector options
3.2) The detector specific options

[1] and [2] have been implemented, [3] is wip at the moment.

troessner:
[3] has been done as well. I'm closing this issue to prevent it from becoming more messy and will create a follow-up issue for [4]

PR #1112, opened 2016-10-03
Introduce CodeCommentValidator.

troessner:
PR nr. 3 hash-tag long-flight ;)

A typical configuration via code comment looks like this:

  :reek:DuplicateMethodCall { enabled: false }

There are a lot of ways a user can introduce some errors here:

1.) Unknown smell detector
2.) Garbage in the detector configuration like { thats: a: bad: config }
3.) Unknown configuration keys (e.g. by doing a simple typo: "exclude" vs. "exlude" )
4.) Bad data types given as values for those keys
This class validates [1] and [2] at the moment but will also validate [3]
and [4] in the future.

This is basically a refactoring of the already working [1] and [2] to mitigate working on [3] and [4].
Contributes to #1102

### DIFF
--- a/lib/reek/code_comment.rb
+++ b/lib/reek/code_comment.rb
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
+
 require 'yaml'
+
 require_relative 'smell_detectors/base_detector'
 require_relative 'errors/bad_detector_in_comment_error'
-require_relative 'errors/bad_detector_configuration_in_comment_error'
+require_relative 'errors/garbage_detector_configuration_in_comment_error'
 
 module Reek
   #
@@ -37,16 +39,13 @@ module Reek
       @source            = source
       @config            = Hash.new { |hash, key| hash[key] = {} }
 
-      @original_comment.scan(CONFIGURATION_REGEX) do |detector, _option_string, options|
-        escalate_bad_detector(detector) unless SmellDetectors::BaseDetector.valid_detector?(detector)
-
-        begin
-          parsed_options = YAML.load(options || DISABLE_DETECTOR_CONFIGURATION)
-        rescue Psych::SyntaxError
-          escalate_bad_detector_configuration(detector)
-        end
-
-        @config.merge! detector => parsed_options
+      @original_comment.scan(CONFIGURATION_REGEX) do |detector_name, _option_string, options|
+        CodeCommentValidator.new(detector_name:    detector_name,
+                                 original_comment: original_comment,
+                                 line:             line,
+                                 source:           source,
+                                 options:          options).validate
+        @config.merge! detector_name => YAML.load(options || DISABLE_DETECTOR_CONFIGURATION)
       end
     end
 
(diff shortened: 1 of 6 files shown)


lib/reek/code_comment.rb
```diff
@@ -64,19 +63,77 @@ def sanitized_comment
...
+    # @param options [String] - the configuration options as String for the detector that were
+    #   extracted from the original comment
+    def initialize(detector_name:, original_comment:, line:, source:, options: {})
+      @detector_name    = detector_name
+      @original_comment = original_comment
+      @line             = line
+      @source           = source
+      @options          = options
+      @detector_class   = nil # We only know this one after our first initial checks.
+      @parsed_options   = nil # We only know this one after our first initial checks.
```
mvz: These last two ivars don't seem to be used.

spec/reek/code_comment_spec.rb
```diff
@@ -1,7 +1,6 @@
 require_relative '../spec_helper'
 require_lib 'reek/code_comment'
-require_lib 'reek/errors/bad_detector_configuration_in_comment_error'
-require_lib 'reek/errors/bad_detector_in_comment_error'
+require_lib 'reek/errors'
```
mvz: I don't think this require are needed, since `code_comment` requires the errors already.

lib/reek/code_comment.rb
```diff
@@ -64,19 +63,77 @@ def sanitized_comment
...
+  #
+  # 1.) Unknown smell detector
+  # 2.) Garbage in the detector configuration like { thats: a: bad: config }
+  # 3.) Unknown configuration keys (e.g. by doing a simple typo: "exclude" vs. "exlude" )
+  # 4.) Bad data types given as values for those keys
+  # This class validates [1] and [2] at the moment but will also validate [3]
+  # and [4] in the future.
+  #
+  # :reek:TooManyInstanceVariables: { max_instance_variables: 7 }
+  class CodeCommentValidator
```
mvz: Why not break this out into its own file?

troessner: I wanted to stress that CodeCommentValidator is **only** used by CodeComment.
What do you think about modelling as pseudo-inner class a la:

```Ruby
class CodeComment
  # ...
  class CodeCommentValidator
  end
end
```

This wont prevent anybody from instantiating it but at least it communicates our intent here in a pretty clear way.

spec/reek/code_comment_spec.rb
troessner: Good point, fixed!

lib/reek/code_comment.rb
mvz: Comparing it to how our detectors our organized that does make sense.

lib/reek/errors.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+require_relative 'errors/bad_detector_in_comment_error'
+require_relative 'errors/garbage_detector_configuration_in_comment_error'
```
mvz: This file can be removed now.

lib/reek/code_comment.rb
troessner: Updated!

lib/reek/errors.rb
troessner: Updated.

troessner:

Issue #1111, opened 2016-10-03
Refactor and split up SmellDetector (or BaseDetector hopefully soon)


troessner:
Seriously, that thing is getting way too big.

PR #1113, opened 2016-10-03
Validate configuration keys in code comments

troessner:
Hash-tag #long-flight nr.4 ;)

Branched of of #1112 and heavily wip. Please hold off with a review for now.
I'll stop with this one here until we have #1112 merged and continue from there.

### DIFF
--- a/lib/reek/code_comment.rb
+++ b/lib/reek/code_comment.rb
@@ -4,6 +4,7 @@ require 'yaml'
 
 require_relative 'smell_detectors/base_detector'
 require_relative 'errors/bad_detector_in_comment_error'
+require_relative 'errors/bad_detector_configuration_key_in_comment_error'
 require_relative 'errors/garbage_detector_configuration_in_comment_error'
 
 module Reek
@@ -75,10 +76,10 @@ module Reek
     # 2.) Garbage in the detector configuration like { thats: a: bad: config }
     # 3.) Unknown configuration keys (e.g. by doing a simple typo: "exclude" vs. "exlude" )
     # 4.) Bad data types given as values for those keys
-    # This class validates [1] and [2] at the moment but will also validate [3]
-    # and [4] in the future.
+    # This class validates [1], [2] and [3] at the moment but will also validate
+    # [4] in the future.
     #
-    # :reek:TooManyInstanceVariables: { max_instance_variables: 5 }
+    # :reek:TooManyInstanceVariables: { max_instance_variables: 7 }
     class CodeCommentValidator
       #
       # @param detector_name [String] - the detector class that was parsed out of the original
@@ -94,16 +95,20 @@ module Reek
         @line             = line
         @source           = source
         @options          = options
+        @detector_class   = nil # We only know this one after our first initial checks
+        @parsed_options   = nil # We only know this one after our first initial checks
       end
 
       #
       # Method can raise the following errors:
       #   * Errors::BadDetectorInCommentError
       #   * Errors::GarbageDetectorConfigurationInCommentError
+      #   * Errors::BadDetectorConfigurationKeyInCommentError
       # @return [undefined]
       def validate
         escalate_bad_detector
         escalate_bad_detector_configuration
+        escalate_unknown_configuration_key
       end
 
       private
@@ -112,7 +117,9 @@ module Reek
                   :original_comment,
                   :line,
                   :source,
-                  :options
+                  :options,
+                  :detector_class,
+                  :parsed_options
 
       def escalate_bad_detector
         return if SmellDetectors::BaseDetector.valid_detector?(detector_name)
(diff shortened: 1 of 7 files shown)


lib/reek/smell_detectors/base_detector.rb
```diff
@@ -14,7 +14,9 @@ module SmellDetectors
     # for details.
     #
     # :reek:UnusedPrivateMethod: { exclude: [ smell_warning ] }
+    # :reek:TooManyMethods: { max_methods: 18 }
```
troessner: I hear you screaming and rightfully so. But fixing this whole mess is something I'd really like to tackle in the course of #1111

lib/reek/smell_detectors/data_clump.rb
```diff
@@ -45,6 +45,10 @@ def self.default_config
           MIN_CLUMP_SIZE_KEY => DEFAULT_MIN_CLUMP_SIZE)
       end
 
+      def self.custom_options
+        Set.new [:max_copies, :min_clump_size]
```
troessner: Yes, this is duplicated. Using the constants from above in here would require to convert them to symbols because they are strings, which is also not exactly appealing.
Fixes this in this PR would make it completely explode - I intend to fix this in the course of #1123

troessner:

troessner:
@mvz @waldyr any of you up for reviewing this one?
I'm flying 13 hours tomorrow and I'd love to get started on validating the datatypes of the values and that code will be based on this one...

lib/reek/smell_detectors/base_detector.rb
```diff
@@ -122,6 +124,25 @@ def valid_detector?(detector)
...
+        #
+        def to_detector(detector_name)
+          SmellDetectors.const_get detector_name
+        end
+
+        #
+        # @return [Set<Symbol>] - all configuration keys that are available for this detector
+        #
+        def configuration_keys
+          BASIC_OPTIONS + custom_options
```
mvz: How about implementing this as `Set.new(default_options.keys.map(&:to_sym)`?

lib/reek/smell_detectors/utility_function.rb
```diff
@@ -45,10 +45,12 @@ def self.default_config
         super.merge(PUBLIC_METHODS_ONLY_KEY => PUBLIC_METHODS_ONLY_DEFAULT)
       end
 
-      class << self
-        def contexts # :nodoc:
-          [:def]
-        end
+      def self.contexts
```
mvz: :+1:

lib/reek/code_comment.rb
```diff
@@ -75,10 +76,10 @@ def sanitized_comment
     # 2.) Garbage in the detector configuration like { thats: a: bad: config }
     # 3.) Unknown configuration keys (e.g. by doing a simple typo: "exclude" vs. "exlude" )
     # 4.) Bad data types given as values for those keys
-    # This class validates [1] and [2] at the moment but will also validate [3]
-    # and [4] in the future.
+    # This class validates [1], [2] and [3] at the moment but will also validate
+    # [4] in the future.
     #
-    # :reek:TooManyInstanceVariables: { max_instance_variables: 5 }
+    # :reek:TooManyInstanceVariables: { max_instance_variables: 7 }
```
mvz: :crying_cat_face:

lib/reek/code_comment.rb
```diff
@@ -94,16 +95,20 @@ def initialize(detector_name:, original_comment:, line:, source:, options: {})
         @line             = line
         @source           = source
         @options          = options
+        @detector_class   = nil # We only know this one after our first initial checks
+        @parsed_options   = nil # We only know this one after our first initial checks
```
mvz: I think this is telling us this should be restructured, but I'm not sure how yet.

troessner: Fully agree! I had the exact same thoughts. My initial idea was to create separate classes for each validator, but since the validations are dependent on each other this lead to a huge amount of boilerplate code where we would do nothing else but pass parameters around.

I suggest to keep it like that because there is also one last validation missing: the ones for the datatypes. When we have that in place we're fully done validating source code comments (yay!) and then we have the test coverage in place to refactor.

lib/reek/smell_detectors/base_detector.rb
troessner: Sorry, but I don't follow. Where's `default_options` coming from, where's the benefit of doing it like that and how is `Set.new(default_options.keys.map(&:to_sym)` related to `BASIC_OPTIONS + custom_options`?

mvz: `default_options` is the method each smell detector implements containing option keys and their default values. Since the goal here is to get those keys in symbolized form, it seems most straightforward to take them straight from `default_options`.

troessner: Ah, you were referring to `default_config`, not `default_options` :)
Sorry, was a little confused by this. We also should rename this after this PR.
Good idea, updated the PR.

troessner:
Grmppfff, I need to fix the test coverage :(

troessner:
Ready for re-review! I believe I addressed all comments. lgtm? :)

Issue #1118, opened 2016-10-08
Hide progress bar when STDOUT is not a TTY


mvz:
I don't think it makes sense to show the progress bar when piping to `more` or a file.

troessner:
Ack.

troessner:
Merged #1122 - closing this one.

PR #1119, opened 2016-10-08
Hide progress on non-ttys

mvz:
Implements #1118.

### DIFF
--- a/lib/reek/cli/options.rb
+++ b/lib/reek/cli/options.rb
@@ -37,10 +37,10 @@ module Reek
         @parser             = OptionParser.new
         @report_format      = :text
         @location_format    = :numbers
-        @progress_format    = :dots
+        @progress_format    = tty_output? ? :dots : :quiet
         @show_links         = true
         @smells_to_detect   = []
-        @colored            = color_support?
+        @colored            = tty_output?
         @success_exit_code  = DEFAULT_SUCCESS_EXIT_CODE
         @failure_exit_code  = DEFAULT_FAILURE_EXIT_CODE
         @generate_todo_list = false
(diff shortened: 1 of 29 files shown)


troessner:
Do we now have an actual integration test for the --progress flag left? 
I dont think that just specs are enough...

mvz:
> Do we now have an actual integration test for the --progress flag left? 

I'll take a look.

mvz:
Yes, we do: `features/command_line_interface/show_progress.feature`

mvz:
I'll rebase this once #1121 has been merged.

troessner:
>> Yes, we do: features/command_line_interface/show_progress.feature

The problem I have with this is the following: Suppose I'm an innocent reader of our features (or us in 6 months!), just by reading them top down my impression would be that showing the progress is something that is **false by default** and that you have to enable it explicitly.
I mean, in none of the features it has any impact but in one feature it does, and that's the show_progress.feature where we enable it explicitly.
Or am I missing something here?

lib/reek/cli/options.rb
```diff
@@ -57,8 +57,8 @@ def parse
       private
 
       # :reek:UtilityFunction
-      def color_support?
-        $stdout.tty?
+      def piped_output?
```
troessner: I think the name is confusing. In the cucumber features above nothing gets piped, at least not in the way I - and probably others - understand pipes. Basically the way we run Reek in our features is the default way that involves no pipe. Maybe rename this to "non_tty_output" and put an explanatory comment above?

I really have a shitty understanding of the tty concept which comes from medieval times and I really do not want to read [tty demystified](http://www.linusakesson.net/programming/tty/) again in my life. Like, ever :-/

mvz:
... and it's green!

lib/reek/cli/options.rb
mvz: That makes sense.

mvz:
I see your point, but I'm not sure how we can solve it, since all output of processes running under Aruba will be non-tty. We have a similar problem for `--color`.

troessner:
Maybe add an explanation for this confusing set up in our contributing docs?

mvz:
Ok, I'll add something there, and a comment in the relevant feature files too.

mvz:
I processed all comments, but I'll open another pull request on master for final review and to make Travis do its thing.

PR #1125, opened 2016-10-12
Emit fingerprints in Code Climate reporter

### LINKED ISSUE #1115: Generate specific Code Climate fingerprints for some issues
Hi,

I'm an engineer at @codeclimate, and we've recently gotten some customer confusion from PR comparisons sometimes showing the same effective issue as both "fixed" and "new" on some issues from Reek.

The underlying logic we use in comparisons is based on the [`fingerprint`](https://github.com/codeclimate/spec/blob/master/SPEC.md#data-types) attribute that engines can optionally emit on issues. When a `fingerprint` isn't included, Code Climate generates a default one based on a hash of the code referred to by the issue's `location`, but for some types of issues this doesn't work very well.

For example: https://github.com/almostwhitehat/codeclimatetest/blob/master/example.rb contains some code that Reek detects as `RepeatedConditional`, and [this PR](https://github.com/almostwhitehat/codeclimatetest/pull/3/files) makes a small change that ends up changing the fingerprint because the underlying code changed. (I'm afraid the repo doesn't actually have Code Climate configured, so it's not obvious just from visiting the pages: one way to confirm this is to clone the repo and use the Code Climate CLI, e.g. `code climate analyze -f json`, to see the generated fingerprint.)

In other engines, we've addressed this by generating fingerprints on more stable attributes like the check name & the attached message. Our [RuboCop implementation](https://github.com/codeclimate/codeclimate-rubocop/blob/master/lib/cc/engine/fingerprint.rb) is a good example of the approach we've taken elsewhere, and I think a similar approach might make sense for Reek as well.

Our team is interested in helping fix this issue, so I'm opening this so we can discuss what approach would make sense for you. We're happy to discuss details & help with implementation if we can. I think one thing it would be helpful to discuss here is if there are checks aside from `RepeatedConditional` in Reek that would be good to include fingerprint logic for.

One more small detail: the [codeclimate-engine-rb](https://github.com/andyw8/codeclimate-engine-rb) gem you're using for JSON generation doesn't actually support the `fingerprint` key right now. I opened a [small PR](https://github.com/andyw8/codeclimate-engine-rb/pull/7) this afternoon to fix that, so hopefully that will get merged, but so you know that's a small blocker on fixing this issue fully.

maxjacobson:
Hi! Jumping in for @wfleming. This should close #1115.

This enables Code Climate to more accurately identify whether an issue
is new, as opposed to changed.

I chose to look at the `warning.parameters` value rather than the
`warning.message` value, because the message (for some smell types) will
change over time. For example:

If you have this code:

```ruby
class Alfa
  Bravo = Charlie = Delta = 1
end
```

You will have a warning with message "Alfa has 3 constants".

Later, when you add a 4th constant, it will change to "Alfa has 4
constants".

Ideally, this would be flagged as the same Code Climate issue, but with
a greater remediation points value, so we can flag it as getting worse.
For now, I think it's good to simply treat it as the same issue.

The way this is accomplished is by CodeClimateFingerprint maintaining a
list of parameters to ignore (:count and :depth, currently). I think
this is a good light-weight approach, but am open to the idea of moving
it somewhere else if you're concerend that may fall out-of-sync over
time.

There are two smell types which don't seem to be identifiable based on
smell_type + context + parameters, which are ManualDispatch and
NilCheck. I believe those could become identifiable if they had
parameters, but haven't done that.

What do you think?

### DIFF
--- a/lib/reek/report/code_climate/code_climate_fingerprint.rb
+++ b/lib/reek/report/code_climate/code_climate_fingerprint.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+module Reek
+  module Report
+    # Generates a string to uniquely identify a smell
+    class CodeClimateFingerprint
+      NON_IDENTIFYING_PARAMETERS = [:count, :depth].freeze
+
+      def initialize(warning)
+        @warning = warning
+      end
+
+      def compute
+        return unless warning_uniquely_identifiable?
+
+        identify_warning
+
+        identifying_aspects.hexdigest
+      end
+
+      private
+
+      attr_reader :warning
+
+      def identify_warning
+        identifying_aspects << warning.source
+        identifying_aspects << warning.smell_type
+        identifying_aspects << warning.context
+        identifying_aspects << parameters
+      end
+
+      def identifying_aspects
+        @identifying_aspects ||= Digest::MD5.new
+      end
+
+      def parameters
+        warning.parameters.except(*NON_IDENTIFYING_PARAMETERS).sort.to_s
+      end
+
+      def warning_uniquely_identifiable?
+        # These could be identifiable if they had parameters
+        ![
+          'ManualDispatch',
+          'NilCheck'
+        ].include?(warning.smell_type)
+      end
+    end
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -74,6 +74,7 @@ RSpec/VerifiedDoubles:
 # rubocop-rspec expects a CodeClimate namespace to go with the code_climate directory.
 RSpec/FilePath:
   Exclude:
+    - 'spec/reek/report/code_climate/code_climate_fingerprint_spec.rb'
     - 'spec/reek/report/code_climate/code_climate_formatter_spec.rb'
     - 'spec/reek/report/code_climate/code_climate_report_spec.rb'
 
--- a/lib/reek/report/code_climate.rb
+++ b/lib/reek/report/code_climate.rb
@@ -1,3 +1,4 @@
 # frozen_string_literal: true
+require_relative 'code_climate/code_climate_fingerprint'
 require_relative 'code_climate/code_climate_formatter'
 require_relative 'code_climate/code_climate_report'
--- a/lib/reek/report/code_climate/code_climate_formatter.rb
+++ b/lib/reek/report/code_climate/code_climate_formatter.rb
@@ -15,6 +15,7 @@ module Reek
                             categories: categories,
                             location: location,
                             remediation_points: remediation_points,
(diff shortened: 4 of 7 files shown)


troessner:
This looks promising! I'll check it out when I have later, latest at the weekend (i have a loooong flight on Saturday).

troessner:
I'm flying with f***ing United, so the best that I will get is the Teenage mutant hero turtles since both netflix and amazon won't let me watch stuff offline :(

troessner:
Reviewed, looks excellent (i don't like the shared examples but you basically just used what was there already, so that is ok) - merged.

troessner:
I guess you need a proper release now?

lib/reek/report/code_climate/code_climate_fingerprint.rb
```diff
@@ -0,0 +1,48 @@
...
+      def initialize(warning)
+        @warning = warning
+      end
+
+      def compute
+        return unless warning_uniquely_identifiable?
+
+        identify_warning
+
+        identifying_aspects.hexdigest
```
Drenmi: Would there be any value in freezing the returned digest string for good measure? 😀

troessner:
And released it is - 4.5.0 is out with the changes ;)

maxjacobson:
@troessner Awesome, thank you. I'll update our end today and let you know how it goes.

Also, I thought a little bit more about my comment suggesting that `ManualDispatch` and `NilCheck` could become uniquely-identifiable with parameters. It's not strictly true, because there could still be a method which contains the same smell multiple times:

```ruby
class Alfa
  def bravo(charlie)
    true if charlie.respond_to?(:to_a)
    true if charlie.respond_to?(:to_a)
  end
end
```

In this case, there are two occurrences of the ManualDispatch smell, and even if we passed in `charlie.to_a` as a parameter, that wouldn't identify the smells uniquely.

It may be an improvement anyway, because at least it would protect against more trivial changes to the code such as `false if charlie.respond_to?(:to_a)` being flagged as a different smell.

troessner:
>> In this case, there are two occurrences of the ManualDispatch smell, and even if we passed in charlie.to_a as a parameter, that wouldn't identify the smells uniquely.

But in combination with the line number it's unique so we that would fix it, wouldn't it?

This will not fix stuff like

```Ruby
class Alfa
  def bravo(charlie)
    true if charlie.respond_to?(:to_a); true if charlie.respond_to?(:to_a)
  end
end
```

but I think it's safe to ignore this as total edge case ;)

maxjacobson:
Heh, yep. For that and other reasons (code is shifted up and down without changing), we generally don't consider line numbers in fingerprints

PR #1126, opened 2016-10-12
Improve coverage and implementation for ModuleInitialize

mvz:


spec/reek/smell_detectors/module_initialize_spec.rb
```diff
@@ -16,6 +16,16 @@ def initialize; end
                            source:  'string')
   end
 
+  it 'does not report with for a method with a different name' do
```
troessner: I guess that should be "does not report a method with a different name"

mvz: Yes, or something like that :smile:.

troessner:
@mvz one tiny change and this one is ready to go....

mvz:
I already fixed that but GitHub doesn't tell you about it :disappointed:.

troessner:
>> I already fixed that but GitHub doesn't tell you about it 😞.

Too bad. Do you think thats because of our "squash-and-force-push" workflow?

mvz:
> Do you think thats because of our "squash-and-force-push" workflow?

I don't think so. It's probably hard for GitHub to tell if something was fixed or just changed. OTOH, if you just add commits, GitHub can at least tell you which commits you haven't looked at yet.

troessner:
True, but i was expecting an email when you change a line i reviewed, that is fairly easy to tell....

PR #1128, opened 2016-10-13
Require digest for code climate reporter

maxjacobson:
This is a follow-up to https://github.com/troessner/reek/pull/1125

Started testing this upgrade in our canary and realized that the
context where this is running, `Digest` isn't defined 😞.

While I'm here, I took the suggestion to freeze the digest from Drenmi
in the last PR. Seems sensible to me.

### DIFF
--- a/lib/reek/report/code_climate/code_climate_fingerprint.rb
+++ b/lib/reek/report/code_climate/code_climate_fingerprint.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+require 'digest'
+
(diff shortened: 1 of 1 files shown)


troessner:
Merged. Do you need another release or can you point your canary against latest master to make sure it works now?

maxjacobson:
Thanks for the quick merge! No need to release.

I'll run the canary for d710a251e7041093ce62e2b1c7c144120742a0b1 and let you know how it goes

PR #1129, opened 2016-10-13
Remove dependency on activesupport's Hash#except

maxjacobson:
Sigh, one more (I think/hope). I forgot this was added by ActiveSupport.
It's required somewhere in the specs, so it's available outside of the
specs too, but not in the non-spec-context.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -74,6 +74,7 @@ RSpec/VerifiedDoubles:
 # rubocop-rspec expects a CodeClimate namespace to go with the code_climate directory.
 RSpec/FilePath:
   Exclude:
+    - 'spec/reek/report/code_climate/code_climate_configuration_spec.rb'
     - 'spec/reek/report/code_climate/code_climate_fingerprint_spec.rb'
     - 'spec/reek/report/code_climate/code_climate_formatter_spec.rb'
     - 'spec/reek/report/code_climate/code_climate_report_spec.rb'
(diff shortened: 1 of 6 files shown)


maxjacobson:
(Realizing you may prefer to use pure Ruby for the main library code. Happy to update it, although it is a very convenient method 😄)

maxjacobson:
I updated it :)

maxjacobson:
@troessner Going to push another fix after lunch actually. Testing locally without going thru this feedback cycle uncovered one more thing

maxjacobson:
@troessner OK thanks for your patience. I think this is good now. What do you think?

troessner:
Merged! What next?

maxjacobson:
No more steps! It's shipped. I've updated [our docs](https://docs.codeclimate.com/docs/reek) to reflect that we're using the newer version. Thanks for your help!!

troessner:
Wait, but you still need an official release, right? I mean, you don't want to point your application against a git SHA?

maxjacobson:
The way our system works, for community-maintained engines we always point at a sha. Generally we pick shas that were shipped as official releases, but in this case using https://github.com/troessner/reek/commit/3d108cf67238d251651a47bdf29f377833f9d123 (and calling it 4.5.0 in our docs) seems appropriate to me. If you want to cut it as an official release I'll update the docs to reflect that.

troessner:
Cool, then I'll just leave it like that ;)

PR #1142, opened 2016-11-12
Remove unnecessary require.

troessner:


mvz:
Why is there no conflict emoji :angry:?

troessner:
Rebased!

PR #1148, opened 2016-11-27
Update How-To-Write-New-Detectors.md

iwamot:


troessner:
Can't believe we missed that! Thanks a lot - merged.

iwamot:
Reek is very useful. Thank you, too!

PR #1149, opened 2016-11-27
Fix gem requirements on v2.2.1 to make it compatible with ruby 1.9.3

budmc29:
The problem:
On rubygems.org version 2.2.1 has required ruby version >= 1.9.3, but the unparser gem required ruby 2.1.

The solution:
Use unparser version 0.1.16 which is the last one compatible with ruby 1.9.3.

### DIFF
--- a/reek.gemspec
+++ b/reek.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency 'parser',   '~> 2.2'
   s.add_runtime_dependency 'rainbow',  '~> 2.0'
-  s.add_runtime_dependency 'unparser', '~> 0.2.2'
+  s.add_runtime_dependency 'unparser', '~> 0.1.16'
 
   s.add_development_dependency 'activesupport', '~> 4.2'
   s.add_development_dependency 'aruba',         '~> 0.6.2'


budmc29:
I did not have the pleasure of working with tags in git and don't really know what is the protocol of patching an older tag.
Any help is really appreciated.

troessner:
Your PR goes against master, that won't work ;)
Try to point it to the commit that is associated to the 2.2.1 tag.

budmc29:
Couldn't manage to point the pull request to the v2.2.1 tag.

mvz:
@budmc29 yeah that makes sense since you can only create a pull request against a branch.

My recommendation would be to restore the patch branch you deleted (`patch-tag-v2.2.1`), and point your Gemfile to that or build a custom Reek version from that locally.

troessner:
But he could just point his branch to the commit the tag points to @mvz ?

mvz:
@troessner If there is nothing pointing to that commit it won't be available locally.

budmc29:
@mvz thanks for the tips. I am more comfortable pointing to your guys gem (even if it's 1.1.6) than pointing to my repo or building a local one.

So is there no way to modify an old gem version (I am referring to rubygems.org) to patch it up?

mvz:
@budmc29 We can't replace the current 2.2.1, but we could release a 2.2.2. However, we would have to test it to confirm that it actually works, which could be tricky given you want to target 1.9.3.

Also, the commit introducing unparser 0.2.2 into Reek (just before Reek 1.6.3!) explicitely mentions it supporting 1.9.3, so I wonder what's going on. I can't be believe we'd have so many versions of Reek accidentally not working on 1.9.3 without us noticing (especially since we did test on 1.9.3 through Travis at that time).

mvz:
@budmc29 Ah yes: support for 1.9.3 was dropped from unparser *after* the release of Reek 2.2.1: https://github.com/mbj/unparser/commit/5d982ec2f0683f4ff8061b46c8b8e165905ab815.

The solution is to pin unparser to 0.2.4 or earlier in your Gemfile.

budmc29:
I'll stick with reek 1.1.6 which is ok for me.
I wanted to make this fix (even updating the required ruby from 1.9.3 to 2.1) so others that still have legacy projects pinned to 1.9.3 to be spared this hassle.

Given that there weren't many complains I guess it's ok to forget about this one.

Thank you very much for taking the time to respond.

mvz:
@budmc29 You didn't read my other comment, it seems: Reek 2.2.1 should work fine with Ruby 1.9.3, as long as you install unparser 0.2.4 or earlier, which will satisfy Reek 2.2.1's dependency. It is only unparser 0.2.5 that is not compatible with Ruby 1.9.3.

budmc29:
Great! Thanks.

Issue #1150, opened 2016-12-01
Calls to super with arguments don't get counted as references to self


jhubert:
This code triggers FeatureEnvy:

```ruby
def foo=(bar)
  super(bar.presence)
end
```

It looks like there was a change made in #768 to fix the `zsuper` version from triggering `UtilityFunction` but no such fix was put in place for FeatureEnvy.

mvz:
This was fixed in #1151. Thanks, @jhubert!

PR #1151, opened 2016-12-01
Stop reporting FeatureEnvy with super and arguments

jhubert:
I honestly don't know why this fixes the problem. Based on my inspection of the code, it seems like this may ensure that super is added to the context when the code is parsed so that it can be counted by the `Reek::AST::ReferenceCollector`. 

I didn't see any tests for the `CodeContext` that tested things like this so I didn't add one, but it seems like it would make sense to test these out a little bit.

It would be helpful if someone could explain what's going on here. 😄 

Per #1150

### DIFF
--- a/lib/reek/context_builder.rb
+++ b/lib/reek/context_builder.rb
@@ -19,7 +19,7 @@ module Reek
   # counting. Ideally `ContextBuilder` would only build up the context tree and leave the
   # statement and reference counting to the contexts.
   #
-  # :reek:TooManyMethods: { max_methods: 30 }
+  # :reek:TooManyMethods: { max_methods: 31 }
   # :reek:UnusedPrivateMethod: { exclude: [ !ruby/regexp /process_/ ] }
   class ContextBuilder
     attr_reader :context_tree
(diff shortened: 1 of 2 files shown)


mvz:
The way this works currently is that `ContextBuilder`, in addition to building a tree of contexts, also does some side jobs, such as triggering the counting of use of self. That's what your implementation does: The `record_use_of_self` method does the actual counting.

However, because there is no further processing done of the arguments of `super`, I think we may be missing stuff here.

The same actually goes for `zsuper`, where I *think* we completely ignore any passed block (but I'd have to check the AST if the block is actually inside the `zsuper` node).

jhubert:
So, are you saying that the code should be changed to:

```ruby
def process_super(exp)
  current_context.record_use_of_self
  process(exp)
end
```

and that the `zsuper` one should be:

```ruby
def process_zsuper(exp)
  current_context.record_use_of_self
  increase_statement_count_by(exp.block)
  process(exp)
end
```

I haven't had a lot of time to understand the code, so just trying to figure out the next steps.

mvz:
> So, are you saying that the code should be changed ..

Yes, eventually, but for now I think your code is fine and can be merged. @troessner wdyt?

troessner:
Very nice @jhubert - merged! 

@mvz I am on the road the whole day, could you take care of a bugfix release?

jhubert:
@troessner @mvz 🎉 thanks guys!

troessner:
And released it is! (4.5.3)

Issue #1152, opened 2016-12-03
Report unnecessary suppressions


jhubert:
Rubocop has a nice process where they let you know if you've unnecessarily suppressed a smell. It would be nice to have the same thing in reek so that my code doesn't end up littered with unnecessary comments while we refactor.

mvz:
Some work is already in progress on this. See #961 and #1060.

mvz:
Closing this as a duplicate of #961.

PR #1153, opened 2016-12-03
WIP: Report unnecessary suppression of smells

jhubert:
This branch reports unnecessary suppression, but it's a little too
greedy. It reports class level use of suppression if it's not necessary
for the whole class.

Working on a refactor of that but wanted to get this up here in the
mean time.

Per #1152

### DIFF
--- a/features/configuration_via_source_comments/well_formed_source_comments.feature
+++ b/features/configuration_via_source_comments/well_formed_source_comments.feature
@@ -114,3 +114,19 @@ Feature: Well formed source comments are handled properly
       configure_detector_specific_option.rb -- 1 warning:
         [3]:TooManyInstanceVariables: Alfa has at least 3 instance variables [https://github.com/troessner/reek/blob/master/docs/Too-Many-Instance-Variables.md]
       """
+
+  Scenario: Disable smell detector unnecessarily
+    Given a file named "disable_detector_implicitly.rb" with:
+      """
+      # Test class
+      # :reek:TooManyInstanceVariables
+      class Alfa
+      end
+      """
+    When I run reek disable_detector_implicitly.rb
+    Then the exit status indicates smells
+    And it reports:
+      """
+      disable_detector_implicitly.rb -- 1 warning:
+        []:TooManyInstanceVariables: Alfa is unnecessarily suppressed [https://github.com/troessner/reek/blob/master/docs/Too-Many-Instance-Variables.md]
+      """
--- a/lib/reek/smell_detectors/base_detector.rb
+++ b/lib/reek/smell_detectors/base_detector.rb
@@ -40,10 +40,9 @@ module Reek
       end
 
       def run_for(context)
-        return [] unless enabled_for?(context)
-        return [] if exception?(context)
+        return [] unless config.enabled?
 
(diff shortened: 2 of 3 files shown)


mvz:
The tests are mainly failing because this also reports on Reek's default suppressions.

troessner:
@jhubert are you still on this? Really looking forward to this ;)

jhubert:
Ok, just did another push.

I have to add a few cucumber tests for the different configuration options.

However, there is one test failing that I'm pretty worried about. The situation is that if you have a suppression like this:

```
# :reek:UncommunicativeVariableName
class Alpha
  def bravo
    x = 0
  end

  def charlie
    y = 0
  end

  def delta
    z = 0
  end
end
```

It will report an unnecessary suppression on Alpha. 

In order to confirm that this is a valid suppression, it would need to know that one of the children raised this smell. I'm not sure how I can figure that out in an efficient way.

Any idea?

troessner:
Hmmm, I see your problem. The way we handle comments is confusing unfortunately due to the distinction between module and method scope of detectors.
Would it make things easier for you if suppression via comment would always be bound to where it actually happens? In your example above this would mean that the suppressions would be noted above the separate methods and not above the class definition.

jhubert:
Yeah, that would mean there is no need to do any recursive smell detection on the context child nodes.

Personally, it feels like a bit of a small to suppress at that level via a code comment.  Seems like something that should be in the config file if you want the whole class to skip a method level smell.

troessner:
The problem with this:

>> Would it make things easier for you if suppression via comment would always be bound to where it actually happens?

is that this would require us to change significant parts of our module scope smell detectors.
I am all up for this, since the way it is right now does make sense on an abstract lebel but is confusing for users. Not sure when I will be able to tackle this. Would you be interestd in doing this?

jhubert:
I think we would be better off finding an expensive way to check the child nodes. Otherwise, this would be a breaking change for a lot of people.

jhubert:
I'm thinking that if no smell is detected, but it's locally disabled via a source code comment we re-check all child nodes for that smell. I think the performance hit would be ok in that case because it would only happen in that edge case.

jhubert:
Then, once we've solved that for this version, we can do the work to remove the abstract comment level for a future major release.

troessner:
>> I'm thinking that if no smell is detected, but it's locally disabled via a source code comment we re-check all child nodes for that smell. I think the performance hit would be ok in that case because it would only happen in that edge case.

I like where this is going. The only thing I'm afraid is that you might end up writing a lot of code that is directly tied to our not-so-great current solution and that then again becomes technical debt as well.
But let's give it a shot and see how it turns out :)

jhubert:
I'll try to keep that in mind. Might just be a class that can do the check
and then be removed when we no longer need it.

troessner:
@jhubert that PR is quite old, ok for you if I close it or were you planning to pick this one up again?

jhubert:
I'm still interested in this but I've not been able to find the time to
focus on it. Feel free to close it and I'll revisit if I get a chance to
pick it back up. Sorry :'(

PR #1157, opened 2016-12-17
Update build dependencies

mvz:
- Update rake to 12.0
- Loosen activesupport dependency
- Tell bundler to respect current Ruby version so it picks the correct version of activesupport
- Ensure latest version of bundler on Travis to make the previous point work there


troessner:
Nice!

PR #1159, opened 2016-12-22
Bump dependencies for Code Climate engine

pointlessone:
* Bumped Ruby to 2.3.3 (was 2.2.2p95)
* Bumped all gems to their latest versions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 # Build and run via:
 #   docker build -t codeclimate/codeclimate-reek . && docker run codeclimate/codeclimate-reek
 
-FROM codeclimate/alpine-ruby:b38
+FROM ruby:2.3.3-alpine
 
 MAINTAINER The Reek core team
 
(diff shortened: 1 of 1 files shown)


mvz:
Looks good. @troessner do you know how we can check that this works?

Dockerfile
```diff
@@ -13,16 +13,17 @@ ENV user app
...
+RUN adduser -u 9000 -D ${user}
+USER ${user}
+
 ADD . ${app_dir}
 
 WORKDIR ${app_dir}
 
 RUN bundle install --without debugging development
-RUN adduser -u 9000 -D ${user}
-RUN chown -R ${user}:${user} ${app_dir}
```
troessner: Why is this not needed anymore? Or wasn't it needed in the first place?

Dockerfile
```diff
@@ -13,16 +13,17 @@ ENV user app
...
+#RUN chown -R ${user}:${user} ${app_dir}
```
troessner: No commented out source code please, you can just remove it.

troessner:
>> Looks good. @troessner do you know how we can check that this works?

We can build the image and then run it via CC's cli. I can look into that but I'd first like to sort my questions ;)

Dockerfile
pointlessone: It is needed. I was experimenting with something and forgot to uncomment it. Will push a fix shortly.

pointlessone: Updated.

pointlessone:
> We can build the image and then run it via CC's cli. I can look into that but I'd first like to sort my questions ;)

That's how I test it.

troessner:
Regarding 

>> Bumped all gems to their latest versions

But that is not reflected in this pull request, is it?

pointlessone:
No, not really. There's no `Gemfile.lock` to update. The latest versions of gems will be installed on each image build. I meant to say that Gemfile and gemspec allow installation of latest version and do not need to be updated.

troessner:
Looks good to me, merged, thanks!

PR #1167, opened 2017-01-17
Update rubocop and resolve offenses

mvz:


Gemfile
```diff
@@ -17,7 +17,7 @@ group :development do
   gem 'yard',          '~> 0.9.5'
 
   if RUBY_VERSION >= '2.3'
-    gem 'rubocop',       '~> 0.47.0'
+    gem 'rubocop',   git: 'https://github.com/bbatsov/rubocop.git'
```
Drenmi: A [bugfix release](https://github.com/bbatsov/rubocop/releases/tag/v0.47.1) has been released. 😀

mvz: Awesome 😁

.rubocop.yml
```diff
@@ -49,6 +49,12 @@ RSpec/DescribeClass:
 RSpec/ExampleLength:
   Enabled: false
 
+# FIXME: Find a better way to block output during specs, and fix relevant examples.
+RSpec/ExpectOutput:
```
backus: Interesting. What issues did you run into here?

mvz: RSpec doesn't really provide a way to block output without making an assertion about it. There is some code in Reek that emits a warning and does other things. Our current setup is intended to be able to spec the other things while preventing those warnings from polluting the spec output.

backus: Ah I see. This is a bit ugly but you could do something like this:

```ruby

shared_examples_for 'handling and recording the error' do
  def error
    expect { return src.syntax_tree }.to output.to_stderr
  end

  it 'does not raise an error' do
    expect { error }.to_not raise_error
  end

  it 'returns an empty syntax tree' do
    expect(error).to be_nil
  end

  it 'records the syntax error' do
    expect { src.syntax_tree }.to output(/#{error_class.name}/).to_stderr
  end

  it 'records the source name' do
    expect { src.syntax_tree }.to output(/#{source_name}/).to_stderr
  end

  it 'records the error message' do
    expect { src.syntax_tree }.to output(/#{error_message}/).to_stderr
  end
end
```

The `error` method takes advantage of the RSpec error suppression and returns the value so you can use it in a test

mvz: Thanks, @backus. I'm going with simple suppressions for now and will revisit this later and see if I can make your suggestion work for all cases.

backus: 👍 Feel free to ping me whenever you're doing rubocop-rspec stuff. I get a lot out of seeing how people interact with the cops.

PR #1172, opened 2017-02-05
Make each check for smells a separate spec

mvz:
This has three effects:
- Feedback is much nicer, since RSpec will print the offending file name as part of the spec description.
- Every file will be checked, regardless of earlier failures in other files.
- Each spec takes a lot less time individually, so there's less chance of hitting the 5 second spec timeout.


troessner:
Nice!

Issue #1117, opened 2016-10-08
Report ManualDispatch once with all lines


mvz:
Following up from #1044, #1048 and #1115, I'd like to merge the ManualDispatch warnings for a given context, so we get one warning per method, but with all offending lines listed (similar to e.g., NilCheck).

troessner:
Yes, yes and yes :)

PR #1173, opened 2017-02-05
Combine lines for manual dispatch smells

mvz:
Reports a single warning per method with all smelly lines instead of one warning per smell.

#1117.


troessner:
Nice!

PR #1177, opened 2017-02-05
Raise exceptions from the API

mvz:
This was inspired by the problem fixed by #1171: That particular problem did not result in a build failure because there was only some output during the `test:quality` task, but spec itself passed.

So, what does this do? Instead of always logging exceptions in `Examiner`, we can pass in a handler object. By default, the handler doesn't handle any exceptions so they are all re-raised. However, in the configuration used when we actually run Reek, a logging handler gets passed that handles the exception by logging it, so they exceptions are not re-raised, resulting in the original behavior in that case.


troessner:
Looking good!

PR #1183, opened 2017-02-13
Update parser dependency

mvz:


reek.gemspec
```diff
@@ -22,6 +22,6 @@ Gem::Specification.new do |s|
   s.summary = 'Code smell detector for Ruby'
 
   s.add_runtime_dependency 'codeclimate-engine-rb', '~> 0.4.0'
-  s.add_runtime_dependency 'parser',                '~> 2.3.1', '>= 2.3.1.2'
+  s.add_runtime_dependency 'parser',                '<= 2.5', '>= 2.3.1.2'
```
troessner: Why "<="?

mvz: Good catch, that should be just "<".

mvz:
Updated!

PR #1185, opened 2017-02-18
Re-raise parse errors

mvz:
To keep things consistent, also let parse errors bubble up to Examiner#run, just like config and processing errors.


troessner:
Very nice, merged!

PR #1189, opened 2017-02-25
Improve TodoListCommand specs

mvz:
- Remove specs that duplicate specs for BaseDetector
- Avoid mocking a method on the object under test
- Narrow the spy on File
- Use Reek's own silencer instead of stream re-assignment
- Give cuke feature a name


mvz:
@troessner ready for review!

troessner:
Nice refactoring!

PR #1191, opened 2017-03-06
Remove redundant 'it' in docstrings

dgollahon:
- Discovered these while test driving a new `rubocop-rspec` cop. :)


mvz:
Thanks, @dgollahon!

PR #1197, opened 2017-03-30
Update Rails example configuration to take mailers into account

schmijos:
From the [Rails Guides](http://guides.rubyonrails.org/action_mailer_basics.html):

> Just like controllers, any instance variables we define in the method become available for use in the views.

So I'd like to suggest to losen up the requirements for *ActionMailer* descendants by disabling [`InstanceVariableAssumption`](docs/Instance-Variable-Assumption.md) for `app/mailers`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -557,6 +557,9 @@ Just add this to your configuration file:
     enabled: false
   UtilityFunction:
     enabled: false
+"app/mailers":
+  InstanceVariableAssumption:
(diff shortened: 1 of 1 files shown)


troessner:
Excellent suggestion, merged, thanks!

mvz:
Do mailers have before_filters? I thing those are the reason we have this exclusion for the controllers.

PR #1200, opened 2017-04-02
Implement `--force-exclusion` flag

### LINKED ISSUE #1196: Should `exclude_paths` configuration be respected when checking a file in that path?
Config.reek contents:

```yaml
---
exclude_paths:
  - db/migrate
```

`reek .` returns no smells, but `reek db/migrate/20170328163910_create_features.rb` returns 3 smells. 

Should reek respect the exclude_paths config options when passed in a file in an excluded path?

IanWhitney:
Closes #1196 

There is probably work to do on this (such as creating a new scenario as [@mvz suggests](https://github.com/troessner/reek/issues/1196#issuecomment-290972786)), but I wanted to get this up for review and feedback as soon as possible.

As discussed, this introduces a `--force-exclusion` flag that mirrors the behavior of Rubocop's similarly-named flag. If the config contains an `exclude_path` for `smelly_path` and  CLI is called with

`reek ./smelly_path/file.rb`

Then the behavior of Reek is unchanged and the file will be checked for smell violations

But if the CLI is called with

`reek --force-exclusion ./smelly_path/file.rb`

Then the file will not be checked and Reek will exit with 0.

There's no breaking changes in this commit and all existing tests -- Cucumber and RSpec -- continue to pass.

### DIFF
--- a/spec/reek/source/source_locator_spec.rb
+++ b/spec/reek/source/source_locator_spec.rb
@@ -29,34 +29,66 @@ RSpec.describe Reek::Source::SourceLocator do
       end
     end
 
+    # rubocop:disable RSpec/NestedGroups
     context 'exclude paths' do
       let(:configuration) do
         test_configuration_for(CONFIG_PATH.join('with_excluded_paths.reek'))
       end
 
-      let(:path) { SAMPLES_PATH.join('source_with_exclude_paths') }
+      let(:options) { instance_double('Reek::CLI::Options', force_exclusion?: false) }
 
-      let(:expected_paths) do
-        [path.join('nested/uncommunicative_parameter_name.rb')]
+      context 'when the path is a file name in an excluded directory' do
+        let(:path) { SAMPLES_PATH.join('source_with_exclude_paths', 'ignore_me', 'uncommunicative_method_name.rb') }
+
+        context 'when options.force_exclusion? is true' do
+          before do
+            allow(options).to receive(:force_exclusion?).and_return(true)
+          end
+
+          it 'excludes this file' do
+            sources = described_class.new([path], configuration: configuration, options: options).sources
+            expect(sources).not_to include(path)
+          end
+        end
+
+        context 'when options.force_exclusion? is false' do
+          before do
+            allow(options).to receive(:force_exclusion?).and_return(false)
+          end
+
+          it 'includes this file' do
+            sources = described_class.new([path], configuration: configuration, options: options).sources
+            expect(sources).to include(path)
+          end
+        end
       end
 
-      let(:paths_that_are_expected_to_be_ignored) do
-        [
-          path.join('ignore_me/uncommunicative_method_name.rb'),
-          path.join('nested/ignore_me_as_well/irresponsible_module.rb')
-        ]
-      end
+      context 'when path is a directory' do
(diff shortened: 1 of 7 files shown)


IanWhitney:
I'm not sure why Travis is getting different test results than I am. Full test site continues to run fine for me, as does the ci task:
screenshot 2017-04-02 13 22 12

spec/reek/source/source_locator_spec.rb
```diff
@@ -34,27 +34,57 @@
...
+
+      context 'when path is a file name' do
+        let(:path) { SAMPLES_PATH.join('source_with_exclude_paths','ignore_me','uncommunicative_method_name.rb') }
+
+        context 'when options.force_exclusion? is true' do
+          before do
+            allow(options).to receive(:force_exclusion?).and_return(true)
+          end
+
+          it 'excludes files in ignored directories' do
```
troessner: I found this to be quite confusing. Above it says "when path is a file name". Ok, sounds good. Then it says down here "excludes files in ignored directories" and I'm thinking "wait, what? How can a file name exclude files in directories"?

How about 

>> it 'excludes this file'

to make this clearer?

troessner:
One small comment, rest looks great!

troessner:
@IanWhitney  we're almost there ;)
We just need to fix the rubocop travis failures and squash to one commit, then lgtm

IanWhitney:
I'll poke at the CI problem on Wednesday, probably. Still not sure what's going on there as it seems to be specific to Travis. 

Rubocop should be good. Running it locally shows no problems and codeclimate has marked it as passed.

Once the CI is passing I'll rebase and push.

IanWhitney:
Ah, I hadn't looked at all of the Travis errors, just the first. My fault.

So, it looks like Ruby 2.1 and  2.2 are failing due to a LocalJumpError.

Ruby 2.3 and 2.4 are failing due to Rubocop's, mostly because of nesting. What's the thought on that? The test nesting seems entirely appropriate to me and my inclination is to override Rubocop in this case.

IanWhitney:
a08c47e should fix the LocalJumpError in Ruby 2.1 and 2.2. Turned out to be kind of interesting, if you find new behaviors added to Ruby iterator methods interesting, I guess.

I'll get to the Rubocop stuff in tomorrow or Wednesday.

IanWhitney:
All that's left is Rubocop's flagging of test nesting. I think the nesting is appropriate, so I'm going to disable the `RSpec/NestedGroups` cop for that section of the test file. I'll do that in 1 commit so that it's easy to revert.

IanWhitney:
Yay, all checks have passed. If everyone is good with this I'll rebase.

troessner:
Please squash, then lgtm ;)

IanWhitney:
Done.

ibrahima:
Quick question: is there a way for API consumers to set this option? [pronto-reek](https://github.com/prontolabs/pronto-reek/blob/master/lib/pronto/reek.rb) uses Reek::Examiner directly, and I don't see a way to pass the CLI Options object to that. This option would be very useful for pronto-reek, because right now exclude_paths is effectively useless there.

Thanks!

mvz:
I commented on https://github.com/prontolabs/pronto-reek/issues/19.

PR #1203, opened 2017-04-05
Properly handle absolute paths

### LINKED ISSUE #1202: Exclude paths isn't honored when given an absolute path
Given a config.reek file with

```
---
exclude_paths:
  smelly_code
```

The command `reek -c config.reek --force-exclusion ./smelly_code/smelly.rb` works as expected

But the command

The command `reek -c config.reek --force-exclusion /absolute/path/to/smelly_code/smelly.rb` does not. The exclusion is not honored and code smells are reported.

This comes up if you use an automated runner of reek that converts all relative path to absolute paths. Like Overcommit.

IanWhitney:
Fixes #1202

I came at this a few different ways. First I looked at expanding all
paths when creating a new SourceLocator. That worked, but made a bunch
of tests fail because they were expecting output that contained relative
paths.

So, to fix the bug and not force a rewrite of all the tests I settled on
doing the conversion within the path_excluded method. Doing it here
ensures we're comparing two absolute paths and it prevents the expanded
paths from leaking throughout the app.

I tried implementing a feature test for this but had no success. It's
weirdly hard to get an absolute path for a file you create through
Aruba. But I'm not that familiar with Cucumber and Aruba, so maybe I'm
missing something.

### DIFF
--- a/lib/reek/configuration/app_configuration.rb
+++ b/lib/reek/configuration/app_configuration.rb
@@ -62,7 +62,7 @@ module Reek
       end
 
       def path_excluded?(path)
-        excluded_paths.include?(path)
+        excluded_paths.map(&:expand_path).include?(path.expand_path)
       end
 
       def load_values(configuration_hash)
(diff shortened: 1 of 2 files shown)


troessner:
Looks good to me, thanks a lot!
I'll merge and release a new version once travis is green.

Issue #1198, opened 2017-03-30
Cucumber tests not run by `bundle exec rake`?


IanWhitney:
Looking at the Readme

https://github.com/troessner/reek#developing-reek--contributing

I thought Cucumber tests would run with `bundle exec rake`

But after setting up a Cucumber test that I knew would fail, `bundle exec rake` continued to tell me that all my tests were green.

Running `bundle exec cucumber` ran the features and showed me my failing test as expected.

troessner:
Good catch! Just pushed [a fix](https://github.com/troessner/reek/pull/1199) for this.

mvz:
This seems wrong; I'll look into it more this weekend.

troessner:
Pretty sure that was intentional back then since our cucumber features are slower than a snail on opium :-/

mvz:
I just checked, and when I run `bundle exec rake`, both RSpec and Cucumber tests are run. Something else must be going on here. @IanWhitney does your scenario show up when you run `bundle exec cucumber features --format progress --color` (this is what the rake task does)?

mvz:
> slower than a snail on opium

Thank you for this brilliant simile!

mvz:
@IanWhitney can you give us some more info on this please? Is your scenario run when you do `bundle exec rake ci`? What is the name of your scenario file?

IanWhitney:
This can be closed. Thanks.

IanWhitney:
I must have been typing the wrong command or something. Not sure what was going on here. But from what I see now `bundle exec rake` does run the Cucumber tests as expected. Sorry for the confusion.

mvz:
Thanks for the clarification, @IanWhitney.

PR #1206, opened 2017-04-06
Restore explanation of what our default Rake task is doing

mvz:
#1198.

This reverts commit e7d20b86c95c97314b5ba5ac303634d183173ec8.


mvz:
Just leaving this here so we don't forget to fix the documentation once we figured #1198 out.

mvz:
Since #1198 has been closed, this can be merged, @troessner.

PR #1211, opened 2017-04-26
Use Ruby 2.4 parser for parsing code

pocke:
In Ruby 2.3 or older, the following code is not parseable.

```ruby
if (a, b = foo)
end
```

So, if the above code exists in a repository, reek crashes.
This change fixes the problem.


troessner:
Thank you for that, merged! I'll release an update later today as soon as I'm home.

PR #1212, opened 2017-04-26
Ensure parser gem 2.4.

troessner:
Necessary to make #1211 work since Ruby 2.4. is only supported with the parser of the same version.

@mvz do you actually recall why we put the "< 2.5" in there in #1183 ?


mvz:
> do you actually recall why we put the "< 2.5" in there in #1183 ?

It's because in general I like to avoid open-ended dependencies.

PR #1220, opened 2017-05-30
Syntax

### LINKED ISSUE #1215: [Feature Request] Introduce Syntax smell
Problem
-------

Currently, if project has a ruby file that is with a syntax error, Reek stop the analysis.
This behavior have been introduced by #1185 from reek v4.6.0. 

By this behavior, we can notice syntax errors.
However, we can't get analysis results of ruby files that doesn't have syntax error also.

This behavior has problems when we use reek on editor integrations, code review as a service or something.
The tools and services are not always handle the error, so the users maybe can't get analysis result if syntax error files exist. So, maybe users can't notice the syntax error.

Solution
------

### Introduce Syntax smell

For example, RuboCop provides `Lint/Syntax` cop. https://github.com/bbatsov/rubocop/blob/26bad386fa8e3c85ad3a434faef0ec765ba612dc/lib/rubocop/cop/lint/syntax.rb
RuboCop adds offense of `Lint/Syntax` to files that have syntax error.

I think reek introduces the feature, we can avoid the problem. Users can notice syntax error, and users can get analysis result.

WDYT? @mvz @troessner

pocke:
Close #1215

done
-----

- [x] Add a detector to detect syntax error
- [x] Update some tests
- [x] Add `Syntax` to defaults.reek

Is there anything else to do?


lib/reek/source/source_code.rb
```diff
@@ -82,36 +90,34 @@ def self.from(source)
...
-            begin
-              ast, comments = parser.parse_with_comments(buffer)
-            rescue Racc::ParseError, Parser::SyntaxError => error
-              raise Errors::ParseError, origin: origin, original_exception: error
-            end
+      # @param parser [Parser::Ruby24]
+      # @param source [String] - Ruby code
+      # @return [Anonymous subclass of Reek::AST::Node] the AST presentation
+      #         for the given source
+      def parse(parser, source)
```
troessner: +1 for those changes

spec/reek/source/source_code_spec.rb
```diff
@@ -34,50 +34,31 @@
...
+        parser = instance_double('Parser::Ruby24')
+        allow(parser).to receive(:parse_with_comments).and_raise(error_class, error_message)
+        {
+          parser: parser
+        }
       end
 
-      it 'raises the error' do
-        expect { src.syntax_tree }.to raise_error error_class
+      it 'raiess the error' do
```
troessner: Typo

spec/spec_helper.rb
```diff
@@ -1,5 +1,6 @@
 require 'pathname'
 require 'timeout'
+require 'active_support/core_ext/string/strip'
```
troessner: Why this change?

troessner:
Looks great to me with a couple of smaller questions and comments, thank you for that!
I also just made you a collaborateur to this repository so you can create future pull requests directly against this repo, welcome on board ;)

lib/reek/examiner.rb
```diff
@@ -97,12 +97,15 @@ def smelly?
     #
     # :reek:TooManyStatements { max_statements: 8 }
     def run
-      return [] unless syntax_tree
-      begin
-        examine_tree
-      rescue Errors::BaseError => exception
-        raise unless @error_handler.handle exception
-        []
+      if source.valid_syntax? && syntax_tree
```
mvz: Yeah, I also think this was for empty files. Apparently we don't have specs for this?

troessner: I did check our specs and we have at least [one example](https://github.com/troessner/reek/blob/master/spec/reek/source/source_code_spec.rb#L20) that does check this indeed. I just check out @pocke branch and ran it on both empty and comment-only source files and it worked just. I assume parser's behaviour now has changed to always return a syntax tree.

So seems like we're good with this change ;)

mvz: Excellent!

spec/spec_helper.rb
pocke: Could you see this commit? c67dbaf
This change is needed to run RSpec with one file.

troessner: Ah, got it :)

troessner:
I think we're good to merge this one now.
What's left is:

* Fix that typo ;)
* Squash

I'm happy to merge and do a minor release afterwards. WDYT @mvz ?

mvz:
Yes! 👍 :shipit:

pocke:
I fixed the typo and squashed. I'll merge this pull-request if Travis CI is passed. 
Thanks!

PR #1223, opened 2017-06-02
Use shorthand for README.md code example

petehamilton:
Similar to how `array.map { |x| x.foo }` can be written `array.map(&:foo)`,
`items.sum { |item| item.gross_price }` can be written
`items.sum(&:gross_price)`.

Proposed as I think it helps demonstrate the benefit of the rewrite in terms of conciseness and clarity?

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ class ShoppingCart < ActiveRecord::Base
   has_many :items
 
   def gross_price
-    items.sum { |item| item.gross_price }
+    items.sum(&:gross_price)
   end
 end
 


troessner:
Thanks for the idea. The samples in the README are intentionally "dumb" in the sense that they should as be readable to as many people as possible including absolute beginners to Ruby who may be not familiar with symbol2proc. That might even extend to people who don't speak Ruby at all - e.g. pythonistas - but might still be able to figure out the "dumb" version of the samples as opposed to the "smart" version.

petehamilton:
Makes sense, no problem.

PR #1225, opened 2017-06-03
Include warning.source for syntax smells

maxjacobson:
When this is missing, the Code Climate formatter errors, because the
source is required to be a String.

### DIFF
--- a/features/reports/codeclimate.feature
+++ b/features/reports/codeclimate.feature
@@ -1,0 +1,59 @@
+Feature: Report smells using Code Climate format
+  In order to run as an Engine on Code Climate, output format following their
+  spec.
+
+  Scenario: output is empty when there are no smells
+    Given a directory called 'clean' containing two clean files
+    When I run reek --format code_climate clean
+    Then it succeeds
+    And it reports this Code Climate output:
+    """
+    """
+
+  Scenario: Indicate smells and print them as JSON when using files
+    Given the smelly file 'smelly.rb'
+    When I run reek --format code_climate smelly.rb
+    Then it reports this Code Climate output:
+      """
+      {
+        "type": "issue",
+        "check_name": "UncommunicativeMethodName",
+        "description": "Smelly#x has the name 'x'",
+        "categories": [
+          "Complexity"
+        ],
+        "location": {
+          "path": "smelly.rb",
+          "lines": {
+            "begin": 4,
+            "end": 4
+          }
+        },
+        "remediation_points": 150000,
+        "content": {
+          "body": "An `Uncommunicative Method Name` is a method name that doesn't communicate its intent well enough.\n\nPoor names make it hard for the reader to build a mental picture of what's going on in the code. They can also be mis-interpreted; and they hurt the flow of reading, because the reader must slow down to interpret the names.\n"
+        },
+        "fingerprint": "2b41a3a4bb7de31ac4f5944bf68b7f5f"
+      }
+      NULL_BYTE_CHARACTER
+      {
+        "type": "issue",
+        "check_name": "UncommunicativeVariableName",
+        "description": "Smelly#x has the variable name 'y'",
+        "categories": [
+          "Complexity"
+        ],
+        "location": {
+          "path": "smelly.rb",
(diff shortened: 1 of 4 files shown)


mvz:
The fix seems good, but I would prefer to just add a spec for the Syntax detector, rather than adding three new cucumber scenarios.

lib/reek/smell_detectors/syntax.rb
```diff
@@ -17,21 +17,16 @@ def self.smells_from_source(source)
 
       # :reek:FeatureEnvy
       def smells_from_source(source)
+        context = Struct.new(:exp, :full_name).new(
```
mvz: This does mean creating two new classes every time this method is called. Maybe assign those two structs to constants?

maxjacobson:
> I would prefer to just add a spec for the Syntax detector, rather than adding three new cucumber scenarios

I think covering the formatter with a cuke seems good, since the other formatters are covered. But I take your points that testing the `source` behavior via a unit test rather than a cucumber scenario is probably better. I did try that at first and encountered some errors that I didn't understand, and went in a different direction. If you like I can try again and see if I can figure it out, or maybe report back with more context.

maxjacobson:
Ah, I see what's happening now in my spec. I'm using the `should reek_of` matcher, which alters the behavior of reek in such a way that `Syntax` smells will never be emitted... (see [here](https://github.com/troessner/reek/blob/master/lib/reek/spec/should_reek_of.rb#L15)).

lib/reek/smell_detectors/syntax.rb
maxjacobson: I did make this update btw ^

mvz:
That override is most certainly no longer needed, I'll create a PR to remove it.

mvz:
I agree that it makes sense to cover this report format with a cuke. I'm vaguely worried that this is a format we don't control, so a change to our dependent gem will break these. That's more of a theoretical worry so we can just see how it goes.

mvz:
A spec would still be nice though 😄

maxjacobson:
:) agreed. Just added one. Thanks for clarifying/fixing the blocker there.

samples/syntax_error.rb
```diff
@@ -0,0 +1,2 @@
+class Aruba
```
mvz: This file can go now, I think.

mvz:
Nice! Two last things are left before we can merge this:
* Remove the now-obsolete `syntax_error.rb` sample.
* Squash the commits into one.

maxjacobson:
@mvz done! I split the codeclimate feature out into a separate commit since it felt unrelated to the other changes. Happy to fully squash if you prefer.

samples/syntax_error.rb
maxjacobson: Good eye

troessner:
@maxjacobson I think you mixed up the commits - [this one](https://github.com/troessner/reek/pull/1225/commits/1f48f508e332267d7b8bad0aeeb0968b79bbf8a5) contains stuff that actually belongs in [this one](https://github.com/troessner/reek/pull/1225/commits/c15523ec41bd94d7a4b11476efce973fbc48048b)

features/reports/codeclimate.feature
```diff
@@ -0,0 +1,59 @@
...
+    When I run reek --format code_climate clean
+    Then it succeeds
+    And it reports this Code Climate output:
+    """
+    """
+
+  Scenario: Indicate smells and print them as JSON when using files
+    Given the smelly file 'smelly.rb'
+    When I run reek --format code_climate smelly.rb
+    And it reports this Code Climate output:
```
troessner: s/And/Then ;)

maxjacobson:
@troessner whoops. very good catch. Updated

troessner:
Very nice, merged!

troessner:
@maxjacobson do you need a release or are we good like that?

maxjacobson:
@troessner nope. Just shipped this to production. Thanks for your patience! Let me know if there are any more issues

Issue #1227, opened 2017-06-09
Irresponsible Module doesn't work when only method calls present


AndrewRayCode:
```
$ reek --version

    reek 4.6.1

$ cat file.rb

    class Thing
    end
    
$ reek file.rb

    file.rb -- 1 warning:
        [1]:IrresponsibleModule: Thing has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]

$ cat otherfile.rb

    class Thing
      does :some_thing
    end
    
$ reek otherfile.rb

    Inspecting 1 file(s):
    .
```

troessner:
Wow wow wiwa! That is definitely a problem. I just reproduced this with the following spec that I appended to our existing spec for IrresponsibleModule:

```Ruby
  it 'reports undocumented classes with method calls in it' do
      src = <<-EOS
        class Alfa
          does :some_thing
        end
      EOS

      expect(src).to reek_of(:IrresponsibleModule)
  end
```

I did check if we didn't have an existing spec that would cover this case and indeed we have one that is at least a bit similar, which you can find [here](https://github.com/troessner/reek/blob/master/spec/reek/smell_detectors/irresponsible_module_spec.rb#L175). Looking at the spec:

```Ruby
    it 'does not report constants that are not classes' do
      src = <<-EOS
        #{scope} Alfa
          Bravo = 23
          Charlie = Hash.new
          Delta = ''.freeze
          Echo = Class.new.new
        end
      EOS

      expect(src).not_to reek_of(:IrresponsibleModule)
    end
```

I don't understand why we had this one in the first place. The class in the spec is not just a namespace module so it should have been reported. @mvz can you recall why we configured IrresponsibleModule like this? My first impression is that this looks definitely like a bug that we should fix.

mvz:
It seems the intent of that spec was that Bravo, Charlie, Delta and Echo would not be reported. Something clearly went wrong with Alfa, though.

AndrewRayCode:
It was introduced here https://github.com/troessner/reek/pull/595/files It seems like two of the added specs here are wrong, because two un-commented tests starting with `module Qux` don't reek

mvz:
Yes, it took me a while to understand my line of reasoning from back then, but it's the namespace module thing that's tripping us up here. The specs are not very clear about that part, because it is incidental to what's being tested right there.

I propose the following two fixes:

* Update the specs to make it clearer when there's a namespace module.
[...]

PR #1228, opened 2017-06-11
Improve IrresponsibleModule

mvz:
- Improve clarity of specs
- Only allow constant creation in namespaces

#1227.

PR #1229, opened 2017-06-11
Fixes cases where modules containing only method: thing don't reek

AndrewRayCode:
"Namespace modules" are allowed without documentation. However, reek
incorrectly thinks a module containing method calls like `some_method: thing`
is a namespace module. To fix this, it was suggested in #1227 that if a
module contains *anything* other than a casgn, class or module def, then
it should not be considered a namespace module.

Addresses #1227

### DIFF
--- a/lib/reek/ast/node.rb
+++ b/lib/reek/ast/node.rb
@@ -37,6 +37,22 @@ module Reek
       end
 
       #
+      # Carries out a depth-first traversal of this syntax tree, yielding every
+      # Sexp except any node whose type is listed in the Array `ignoring`. Any
+      # node type in `dont_recurse_children` will be yielded, but its children
+      # will not be walked.
+      #
+      # Examples:
+      #   context.walk_tree() do |node| .... end
+      def walk_tree(ignoring = [], dont_recurse_children = [], &blk)
+        yield self unless ignoring.include?(type)
+        return if dont_recurse_children.include?(type)
+        each_sexp do |elem|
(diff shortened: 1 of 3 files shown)


AndrewRayCode:
This is an attempt to solve the issue, and probably not a great attempt. I will feel no sorrow if this PR is closed.

I didn't see a way to search for *any* node type, to check for *any* node and failing if *any* of them aren't `:cagn/:class/etc` which is why I added a new `walk_tree` method. I considered making an array-like class that always returns `true` for `includes?` to walk every node type.

troessner:
Hey @AndrewRayCode,

I'm afraid you're a bit too late to the party - if I understand this correctly, the issue above is already fixed with @mvz PR [here](https://github.com/troessner/reek/pull/1228), which is close to being merged.

In the PR above you can also see [here](https://github.com/troessner/reek/pull/1228/files#diff-8dec725f4439150ae0f1461c20832402R163) that we added this spec now:

```Ruby
    it "reports #{scope} that contains method calls" do
      src = <<-EOS
        #{scope} Alfa
          bravo :charlie
        end
      EOS

      expect(src).to reek_of(:IrresponsibleModule, context: 'Alfa')
    end
```
which shows that this issue has been fixed.

AndrewRayCode:
Haha I should have looked at the ticket history first. Either way it was a fun exercise to look at the internals of reek. Also good to know that you can just loop over nodes and not use an abstraction and check types there.

PR #1233, opened 2017-07-16
Also report unused uncommunicative parameter names

### LINKED ISSUE #1045: UnusedParameters will swallow UncommunicativeParameterName
> > echo "def m(a); end" | reek
> > STDIN -- 2 warnings:
> >   [1]:UncommunicativeMethodName: m has the name 'm' 
> >   [1]:UnusedParameters: m has unused parameter 'a' 
> > echo "def m(a); a; end" | reek
> > STDIN -- 2 warnings:
> >   [1]:UncommunicativeMethodName: m has the name 'm' 
> >   [1]:UncommunicativeParameterName: m has the parameter name 'a'

mvz:
An unused parameter name should also be reported as uncommunicative, unless it is explicitly marked as unused.

Fixes #1045.

Note that unused parameters marked as such that would become uncommunicative if they became used (e.g., `_foo2`), are not reported.

### DIFF
--- a/features/configuration_files/directory_specific_directives.feature
+++ b/features/configuration_files/directory_specific_directives.feature
@@ -41,7 +41,7 @@ Feature: Directory directives
     And a file named "web_app/app/models/user.rb" with:
       """
       class User < ActiveRecord::Base
-        def logged_in_with_role(r)
+        def logged_in_with_role(role)
           true
         end
       end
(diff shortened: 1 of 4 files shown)


troessner:
Very nice! Are you going to go ahead with a release?

mvz:
I'd like look at a few more points in the coming week and then see if we can release something in the weekend.

troessner:
Sounds good to me.;)

PR #1234, opened 2017-07-18
Initialize a fresh detector for each context

mvz:
The existing implementation used the same detector for all contexts, which resulted in constantly passing around the current context and other values as parameters. This in turn has lead to a large number of
cases of feature envy and similar code smells.

By using each detector instance for just one context, we can migrate to a design where more of the state of the ongoing detection is stored in the detector and less is passed around.

### DIFF
--- a/lib/reek/detector_repository.rb
+++ b/lib/reek/detector_repository.rb
@@ -35,31 +35,25 @@ module Reek
                    configuration: {})
       @configuration = configuration
       @smell_types   = smell_types
-      @detectors     = smell_types.map { |klass| klass.new configuration_for(klass) }
     end
 
     def examine(context)
-      smell_detectors_for(context.type).flat_map do |detector|
-        detector.run_for(context)
+      smell_detectors_for(context.type).flat_map do |klass|
+        detector = klass.new configuration: configuration_for(klass), context: context
+        detector.run
       end
     end
 
     private
 
-    attr_reader :configuration, :smell_types, :detectors
+    attr_reader :configuration, :smell_types
 
     def configuration_for(klass)
       configuration.fetch klass, {}
     end
 
     def smell_detectors_for(type)
-      enabled_detectors.select do |detector|
-        detector.contexts.include? type
-      end
-    end
-
-    def enabled_detectors
-      detectors.select { |detector| detector.config.enabled? }
+      smell_types.select { |detector| detector.contexts.include? type }
     end
   end
 end
--- a/spec/reek/smell_detectors/irresponsible_module_spec.rb
+++ b/spec/reek/smell_detectors/irresponsible_module_spec.rb
@@ -49,17 +49,6 @@ RSpec.describe Reek::SmellDetectors::IrresponsibleModule do
       expect(src).not_to reek_of(:IrresponsibleModule)
     end
 
-    it "does not report re-opened #{scope} in the same file" do
-      src = <<-EOS
-        # This comment describes Alfa
-        #{scope} Alfa; end
-
-        #{scope} Alfa; def bravo; end; end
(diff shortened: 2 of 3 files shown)


mvz:
I think this would be a real improvement to Reek's design. WDYT?

lib/reek/detector_repository.rb
```diff
@@ -35,31 +35,25 @@ def initialize(smell_types: self.class.smell_types,
                    configuration: {})
       @configuration = configuration
       @smell_types   = smell_types
-      @detectors     = smell_types.map { |klass| klass.new configuration_for(klass) }
     end
 
     def examine(context)
-      smell_detectors_for(context.type).flat_map do |detector|
-        detector.run_for(context)
+      smell_detectors_for(context.type).flat_map do |klass|
```
Drenmi: At first glance, it doesn't look like `#smell_detectors_for` would return a two-dimensional array. Could this get away with just `#map`?

lib/reek/detector_repository.rb
```diff
@@ -35,31 +35,25 @@ def initialize(smell_types: self.class.smell_types,
...
 
     def smell_detectors_for(type)
-      enabled_detectors.select do |detector|
-        detector.contexts.include? type
-      end
-    end
-
-    def enabled_detectors
-      detectors.select { |detector| detector.config.enabled? }
+      smell_types.select { |detector| detector.contexts.include? type }
```
Drenmi: There's a cop (which is currently disabled by default) `Style/MethodCallWithArgsParentheses` which checks for the use of parens around method calls with arguments.

It is pretty reasonable if run with `IgnoreMacros: true` (default) and you can selectively whitelist methods as well. I am pushing to make it enabled in the next release of RuboCop, but if you're brave you can enable it now. 🙂

lib/reek/detector_repository.rb
mvz: It doesn't but the block returns an array.

Drenmi: Right. 🙂

spec/reek/smell_detectors/irresponsible_module_spec.rb
```diff
@@ -49,17 +49,6 @@ class Charlie
...
-    it "does not report re-opened #{scope} in the same file" do
-      src = <<-EOS
-        # This comment describes Alfa
-        #{scope} Alfa; end
-
-        #{scope} Alfa; def bravo; end; end
-      EOS
-
-      expect(src).not_to reek_of(:IrresponsibleModule)
-    end
```
troessner: How is this change relevant to this PR?

troessner:
One tiny comment and then we're good to roll. I love this change overall, the "old" way always bothered me.

spec/reek/smell_detectors/irresponsible_module_spec.rb
mvz: This change makes it impossible for detectors to keep state between sniffing one context and the next. Therefore, the behavior specified here can no longer work. I don't think this is a great loss, since:

* Why would you open a class or module twice in a single file, *and* define methods in it both times?
* None of the other detectors keep state for the same kind of case, e.g., TooManyMethods, and UnusedPrivateMethod.

lib/reek/detector_repository.rb
mvz: 🙂

spec/reek/smell_detectors/irresponsible_module_spec.rb
troessner: Gotcha, makes sense :)

mvz:
Awesomeness!

Issue #1207, opened 2017-04-18
Can't comment-suppress UtilityFunction smell for private def


cobbr2:
I *hope* I'm just misspelling something, but I'm seeing the following:

```
# Should suppress utility function smell; does not for me.
class SuppressOnPrivateDef
  # :reek:UtilityFunction
  private def utility_function
    STDOUT.puts "I'm a utility function"
  end
end
```
does *not* suppress, but

```
# Should suppress utility function smell; does.
class SuppressInPrivateSection
  private

  # :reek:UtilityFunction
  def utility_function
    STDOUT.puts "I'm a utility function"
  end
end
```
does. Is there some different magic necessary to get the comment-based suppression to work for `private def` (and presumably `protected def`)?

Full set of my tests and their output is at https://gist.github.com/cobbr2/8ba8523fea3a9c08811211a608a60ab6

$ reek --version
reek 4.6.1
$ ruby --version
ruby 2.1.5p273 (2014-11-13 revision 48405) [x86_64-linux]

troessner:
I can confirm this locally - excellent bug report, thank you for this!

mvz:
In the AST, the comment is probably associated with the `private`, instead of with the `def`.

mvz:
This is fixed in current master.

PR #1236, opened 2017-07-22
Support visibility modifier of method definition

mvz:
This is a run-up to handling #1207: The `private def bla` construction was completely unsupported in Reek.


troessner:
Very cool, merged!

PR #1238, opened 2017-07-23
Remove parent attribute of Reek::AST::Node class

mvz:
This attribute is no longer used, and is also confusing because it is not an object of class Reek::AST::Node.


troessner:
Very nice, merged!

PR #1240, opened 2017-07-23
Handle private def comment

mvz:
This handles #1207.


mvz:
This currently includes #1236.

PR #1244, opened 2017-08-05
Refactor ContextBuilder#handle_send_for_modules

### LINKED ISSUE #1239: Refactor ContextBuilder#handle_send_for_modules
This method currently pretends that it does three things, but in fact either does the first two things, or the third thing, or none of the things. The code should make it clearer that some kind of choice is made here.

For reference, the method currently looks like this:

```ruby
    def handle_send_for_modules(exp)
      method_name = exp.name
      arg_names = exp.arg_names
      current_context.track_visibility(method_name, arg_names)
      current_context.track_singleton_visibility(method_name, arg_names)
      register_attributes(exp)
    end
```

mvz:
- Inline `SendNode#arg_names`. This method was not generally useful for `:send` nodes and did not return argument names in the general case.
- Merge `ModuleContext#track_visibility` and `#track_singleton_visibility`. Having two methods actually made the code more confusing.

Fixes #1239.

### DIFF
--- a/lib/reek/ast/sexp_extensions/send.rb
+++ b/lib/reek/ast/sexp_extensions/send.rb
@@ -23,10 +23,6 @@ module Reek
           ([receiver] + args).compact
         end
 
-        def arg_names
-          args.map { |arg| arg.children.first }
-        end
-
         def module_creation_call?
           object_creation_call? && module_creation_receiver?
         end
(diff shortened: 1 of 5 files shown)


troessner:
Very nice! A lot simpler now indeed.

PR #1250, opened 2017-08-27
Use config file in `reek` matcher

### LINKED ISSUE #1093: reek/spec rspec helper not picking up config.reek
I tested with the following in `config.reek` file in my home directory.

``` yaml

---
"app/controllers":
  IrresponsibleModule:
    enabled: false
  NestedIterators:
    max_allowed_nesting: 2
  UnusedPrivateMethod:
    enabled: false
  InstanceVariableAssumption:
    enabled: false
"app/helpers":
  IrresponsibleModule:
    enabled: false
  UtilityFunction:
    enabled: false
exclude_paths:
  - db
  - config
```

Then in my rails_helper.rb

``` ruby
require 'reek/spec'
```

Then in my `spec/smells/reek_spec.rb` file

``` ruby
require 'rails_helper'

describe 'Code Smells' do
  # Check all ruby files in lib and app directory
  file_patterns = { lib: 'lib/**/*.rb', app: 'app/**/*.rb' }
  file_patterns.each do |directory, file_pattern|

    # within the context of given directory
    context "with #{directory}/ files" do
      Pathname.glob(file_pattern).sort.each do |file|

        # Execute smell check for each file
        context "with #{file} file" do
          it 'contains no code smells' do
            expect(file).not_to reek
          end
        end
      end
    end
  end
end
```

When I run my specs, I get a failure:

``` Shell
$ rspec spec/smells/reek_spec.rb
Failures: 
  1) Sniff Tests with app/ files with app/controllers/foo_bars_controller.rb file contains no code smells
     Failure/Error: expect(file).not_to reek

       Expected no smells, but got:
         InstanceVariableAssumption: FooBarsController assumes too much for instance variable '@foobar'
```

However when I run reek at the command line, I get no errors.

``` Shell
$ reek
0 total warnings
```

mvz:
Fixes #1093.

### DIFF
--- a/features/rspec_matcher.feature
+++ b/features/rspec_matcher.feature
@@ -1,0 +1,32 @@
+Feature: Use reek_of matcher
+  As a developer
+  In order to check code quality as part of my spec suite
+  I want to assert that my code doesn't smell
+
+  Background:
+    Given the smelly file 'smelly.rb'
+    And a file "reek_spec.rb" with:
+    """
+    require 'reek/spec'
+
+    describe 'smelly.rb' do
+      it 'is clean' do
+        expect(Pathname.new('smelly.rb')).not_to reek
+      end
+    end
+    """
+
+  Scenario: Failing on a smelly file
+    When I run `rspec reek_spec.rb`
+    Then stdout should contain:
+    """
+    Failure/Error: expect(Pathname.new('smelly.rb')).not_to reek
+    """
+
+  Scenario: Masking smells with a configuration file
+    Given a configuration file 'full_mask.reek'
+    When I run `rspec reek_spec.rb`
+    Then stdout should contain:
+    """
+    1 example, 0 failures
+    """
--- a/lib/reek/configuration/app_configuration.rb
+++ b/lib/reek/configuration/app_configuration.rb
@@ -17,9 +17,10 @@ module Reek
       include ConfigurationValidator
       EXCLUDE_PATHS_KEY = 'exclude_paths'.freeze
 
-      # Instantiate a configuration via given path.
+      # Instantiate a configuration via given path, or the default path.
       #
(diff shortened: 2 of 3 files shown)


features/rspec_matcher.feature
```diff
@@ -0,0 +1,43 @@
+Feature: Use reek_of mathcer
```
troessner: Typo

features/rspec_matcher.feature
```diff
@@ -0,0 +1,43 @@
...
+
+  Scenario: Pick up configuration file
+    Given the smelly file 'smelly.rb'
+    And a configuration file 'full_mask.reek'
+    And a file "reek_spec.rb" with:
+    """
+    require 'reek/spec'
+
+    describe 'smelly.rb' do
+      it 'is fragrant' do
```
troessner: Should "is NOT" i guess?

features/rspec_matcher.feature
```diff
@@ -0,0 +1,43 @@
...
+        expect(Pathname.new('smelly.rb')).not_to reek
+      end
+    end
+    """
+    When I run `rspec reek_spec.rb`
+    Then it succeeds
+    And stdout should contain:
```
troessner: The second scenario is almost identical with a couple of easy to miss differences. Would be great if we could make the differences more obvious.

features/rspec_matcher.feature
mvz: Fixed

features/rspec_matcher.feature
mvz: I've changed the scenarios around so I could extract more common parts. Let me know what you think.

troessner:
Yes, I like those cucumber scenarios a lot better, merged!

Issue #1131, opened 2016-10-22
FeatureEnvy too curious about what's inside lambda expressions


andreaspixsy:
I ran reek on https://github.com/tueksta/webstats/blob/master/importer.rb
and I got this smell warning:

```
importer.rb -- 4 warnings:
  [10, 11, 12, 13, 14, 15, 16, 17, 18]:FeatureEnvy: Importer#initialize refers to 'arg' more than self (maybe move it to another class?) [https://github.com/troessner/reek/blob/master/docs/Feature-Envy.md]
...
```

In my opinion this is a false positive, as the args are not referred to by Importer#initialize at all: the lambda expressions should be considered by reek as yet another data entry as they will never be executed within the context of Importer#initialize - inspecting the the insides needs a different context, the one where it is actually called.

tueksta:
whoops, wrong account :)

well, I can also imagine not snooping on lambdas at all. is there some configuration to have them ignored?

mvz:
Hi, thanks for your bug report.

This is an interesting issue. First of all, I think Reek should probably consider each `arg` of each lambda as a separate entity. This would make the warning go away, since each `arg` only has one method called on it.

Second,

> as they will never be executed within the context of Importer#initialize

I'm afraid it's not really possible to determine this: If you're using `#each` with a block, that block is in a sense called in the context of the containing method. But some other method may take a block and just store it for later use. The same goes with lambda's.

In the end, though, it doesn't really matter when the block or lambda is called: If it doesn't really use any part of the Importer instance, why does it need to be in that class? Maybe it's better off in another class. In this case, you can consider creating classes `GoodReadsImportSource`, `ImdbImportSource`, etc.

tueksta:
I agree, what I did is not a good final solution. (I also want to avoid specialized importers, as this is considered data and not logic - only with strict separation I retain the flexibility of adding and removing ImportSources without changing code for better scaling. Moving away from lambdas to a more proper representation of what they are trying to achieve will also get rid of the reek warning.)

But anyway, even as an interims solution, I did not expect this smell to appear - by distinguishing variables according to the lambdas/procs/blocks they are created and used in the smell detection by reek would be much more refined, in my opinion.

troessner:
>> First of all, I think Reek should probably consider each arg of each lambda as a separate entity. This would make the warning go away, since each arg only has one method called on it.

Agreed, let's fix this.

troessner:
The original code is not available anymore.
[...]

PR #1252, opened 2017-09-04
[WIP] Handle lambda expressions in Feature Envy

mvz:
#1131.

For now, this just show what I think is the desired behavior. @troessner do you agree that these specs reflect how this should work?

### DIFF
--- a/spec/reek/smell_detectors/feature_envy_spec.rb
+++ b/spec/reek/smell_detectors/feature_envy_spec.rb
@@ -152,6 +152,36 @@ RSpec.describe Reek::SmellDetectors::FeatureEnvy do
       and reek_of(:FeatureEnvy, name: 'bravo')
   end
 
+  it 'reports calls inside lambda expressions' do
+    src = <<-EOS
+      def alfa(bravo)
+        [
+          -> { bravo.charlie },
+          -> { bravo.delta },
+          -> { bravo.echo },
+          foxtrot
+        ]
+      end
+    EOS
+
+    expect(src).to reek_of(:FeatureEnvy)
+  end
+
+  it 'does not report different lambda parameters with the same name' do
+    src = <<-EOS
+      def alfa
+        [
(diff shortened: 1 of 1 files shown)


spec/reek/smell_detectors/feature_envy_spec.rb
```diff
@@ -152,6 +152,36 @@ def alfa
       and reek_of(:FeatureEnvy, name: 'bravo')
   end
 
+  it 'reports calls inside lambda expressions' do
+    src = <<-EOS
+      def alfa(bravo)
+        [
+          -> { bravo.charlie },
+          -> { bravo.delta },
+          -> { bravo.echo },
```
troessner: We shouldn't need this lambda to actually trigger FeatureEnvy, now shouldn't we? The first 2 should be enough following current FeatureEnvy semantics.

spec/reek/smell_detectors/feature_envy_spec.rb
```diff
@@ -152,6 +152,36 @@ def alfa
...
+    expect(src).to reek_of(:FeatureEnvy)
+  end
+
+  it 'does not report different lambda parameters with the same name' do
+    src = <<-EOS
+      def alfa
+        [
+          ->(bravo) { bravo.charlie },
+          ->(bravo) { bravo.delta },
+          ->(bravo) { bravo.echo },
```
troessner: Same here, let's remove this lambda.

troessner:
Couple of smaller comments, overall I agree that this is the way it should behave.

spec/reek/smell_detectors/feature_envy_spec.rb
mvz: It seemed two was not enough, but I'll check again.

troessner:
@mvz this PR seems stale, wdyt about closing it for now?

mvz:
@troessner yes, agreed.

Issue #1231, opened 2017-07-11
DuplicateMethodCall gives false positive for variables in lamdas


oliverklee:
With this class …

```Ruby
class CurrencyInput < SimpleForm::Inputs::CollectionInput
  def input(wrapper_options = nil)
    input_options = merge_wrapper_options(input_html_options, wrapper_options)

    collection = Money::Currency.table
    label_method = lambda do |currency|
      "#{currency[1].fetch(:name)} (#{currency[1].fetch(:iso_code)})"
    end
    value_method = ->(currency) { currency[1].fetch(:iso_code) }

    @builder.collection_select(
      attribute_name, collection, value_method, label_method,
      input_options, input_options
    )
  end
end
```

Reek complains about this.

```
  [20, 20, 22]:DuplicateMethodCall: CurrencyInput#input calls 'currency[1]' 3 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
```

However, as these are lambdas, there are not 3 calls, but 2 + 1.

oliverklee:
And this warning is the same problem:

  [20, 22]:DuplicateMethodCall: CurrencyInput#input calls 'currency[1].fetch(:iso_code)' 2 times [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]

troessner:
Oh yeah, let's fix this. @mvz do you want to tackle this? If not I can also take a stab at it after the release of Reek 5.

mvz:
@troessner this is what #1253 was for. It turned out to be quite difficult to fix. I'm actually more and more inclined to *not* fix this or #1131, because there are many more cases where you're calling the same method on what technically is a different object, e.g.:

```
foo = make_me_a_foo
foo.bla
foo = make_me_another_foo
foo.bla
```

We can't win, really.

troessner:
I understand that your example is kind of not fixable for us, but what about the special case with the lambdas - would that be easier?

mvz:
Clearer, but not a lot easier. Given our current architecture, you would need to make each block its own context.

mvz:
... and even then, if the call is not on one of the block parameters, you have to treat the whole method as one context.

troessner:
Ok, i see. What do you think about making this a prominent part of our DuplicateMethodCall documentation?

mvz:
That sounds like an excellent plan.

troessner:
[...]

PR #1253, opened 2017-09-04
[WIP] Handle block params in DuplicateMethodCall

mvz:
#1231.

For now, this just show what I think is the desired behavior. @troessner do you agree that these specs reflect how this should work?

### DIFF
--- a/spec/reek/smell_detectors/duplicate_method_call_spec.rb
+++ b/spec/reek/smell_detectors/duplicate_method_call_spec.rb
@@ -126,6 +126,40 @@ RSpec.describe Reek::SmellDetectors::DuplicateMethodCall do
     end
   end
 
+  context 'with repeated method calls inside blocks' do
+    it 'does not report calls to parameters of different blocks' do
+      src = <<-EOS
+        def alfa(bravo, charlie)
+          bravo.each { |delta| delta.echo }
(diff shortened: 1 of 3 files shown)


spec/reek/smell_detectors/duplicate_method_call_spec.rb
```diff
@@ -126,6 +126,40 @@ def alfa
...
+          charlie.each { |delta| delta.echo }
+        end
+      EOS
+      expect(src).not_to reek_of(:DuplicateMethodCall)
+    end
+
+    it 'does not report calls inside lambda expressions' do
+      src = <<-EOS
+        def alfa
+          bravo = ->(delta) { delta.echo.foxtrot }
```
troessner: Why do you need the "foxtrot" at the end? We can drop that, can't we?

troessner:
Same as with #1252 - I agree this is the way to go.

spec/reek/smell_detectors/duplicate_method_call_spec.rb
mvz: Probably :-)

troessner:
@mvz this PR seems stale, wdyt about closing it for now?

mvz:
@troessner Agreed. Closing.

PR #1258, opened 2017-09-13
Make prima donna method report the method definition line

### LINKED ISSUE #1256: Prima Donna method should report method definition lines
The prima donna method detector currently reports the line of the defining class, but should report the line of the method definition.

mvz:
This improves the Prima Donna Method detector by making it report the line where the method is defined, rather than the (mostly useless) line where the class starts.

This fixes #1256.


chastell:
(*wakes from ages-old slumber* *clears throat* ummm, yeah, sorry, I think my name was Piotr…)

LGTM, thanks for catching and fixing this – merging!

PR #1259, opened 2017-09-17
Update rubocop(-rspec) and fix offenses

mvz:
Noteworthy parts:
* I only corrected HeredocDelimiterNaming in `lib/`, not in `spec/`.
* I disabled InterpolationCheck and BooleanSymbol.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -36,6 +36,15 @@ Layout/MultilineOperationIndentation:
 Lint/AmbiguousBlockAssociation:
   Enabled: false
 
+# Some of our source examples include interpolation explicitely.
+Lint/InterpolationCheck:
+  Exclude:
+    - 'spec/**/*'
+
+# We use :true and :false as AST node types.
+Lint/BooleanSymbol:
+  Enabled: false
+
 Lint/HandleExceptions:
   Exclude:
     - 'spec/reek/configuration/configuration_file_finder_spec.rb'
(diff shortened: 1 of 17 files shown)


.rubocop.yml
```diff
@@ -63,6 +72,15 @@ Metrics/LineLength:
 Metrics/ParameterLists:
   CountKeywordArgs: false
 
+Naming/AccessorMethodName:
+  Exclude:
+    - 'lib/reek/context/visibility_tracker.rb'
+
+# EOS is a fine name to use in our specs
```
Drenmi: Just curious what the reason is. Asking because I use an editor that will infer which syntax highlight to use based on the delimiter. I guess this could potentially be undesirable if the fixture adds too much noise?

mvz: In this case, 'fine' means that I don't particularly care about which delimiter is used. I actually have half a mind to disable this cop for the entire project.

I don't have any experience with editors that highlight based on the delimiter used, but I can imagine that it becomes rather confusing if the strings represent Ruby code and are highlighted as such.

Drenmi: It mostly saves me when working with custom SQL queries in Rails apps, but I agree this is a minor convenience (or inconvenience depending on your preference.) 🙂

troessner:
Looks good to me - let's squash it then I'll happily merge :)

mvz:
Squashed!

Issue #1254, opened 2017-09-06
Error when run in POSIX locale


lorn:
Hi, I just found a weird bug.

I have a local Docker for my project and `rubycritic` was not running on them, but it is running fine in the gitlab-ci.

Digging a little bit I found that gitlab-ci uses the `ruby:2.4-alpine` image and the local Dockerfile was using `ruby:2.4.1` image.

Searching for this kind of bug I found this issue whitesmith/rubycritic/issues/244 and here I'm to give you some evidences:

```
$ docker run -it ruby:2.4.1 bash
root@fac826472069:/# gem install reek
Fetching: thread_safe-0.3.6.gem (100%)
Successfully installed thread_safe-0.3.6
Fetching: descendants_tracker-0.0.4.gem (100%)
Successfully installed descendants_tracker-0.0.4
Fetching: equalizer-0.0.11.gem (100%)
Successfully installed equalizer-0.0.11
Fetching: coercible-1.0.0.gem (100%)
Successfully installed coercible-1.0.0
Fetching: ice_nine-0.11.2.gem (100%)
Successfully installed ice_nine-0.11.2
Fetching: axiom-types-0.1.1.gem (100%)
Successfully installed axiom-types-0.1.1
Fetching: virtus-1.0.5.gem (100%)
Successfully installed virtus-1.0.5
Fetching: codeclimate-engine-rb-0.4.0.gem (100%)
Successfully installed codeclimate-engine-rb-0.4.0
Fetching: ast-2.3.0.gem (100%)
Successfully installed ast-2.3.0
Fetching: parser-2.4.0.0.gem (100%)
Successfully installed parser-2.4.0.0
Fetching: rainbow-2.2.2.gem (100%)
Building native extensions.  This could take a while...
Successfully installed rainbow-2.2.2
Fetching: reek-4.7.2.gem (100%)
Successfully installed reek-4.7.2
12 gems installed
root@fac826472069:/# reek
Inspecting 2440 file(s):
[...]

PR #1263, opened 2017-10-19
Handle encoding error when parsing the source

### LINKED ISSUE #1262: When erroring out because of an invalid byte-sequence, reek doesn't show what file it crashed on.
```
reek -ns

Inspecting 183 file(s):
SSSSSS....SS.....SSSS.S..SS.SSS..S./usr/local/bundle/gems/parser-2.4.0.0/lib/parser/source/buffer.rb:164:in `source=': invalid byte sequence in US-ASCII (EncodingError)
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/source/source_code.rb:99:in `parse'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/source/source_code.rb:33:in `initialize'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/source/source_code.rb:50:in `new'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/source/source_code.rb:50:in `from'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/examiner.rb:40:in `initialize'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/cli/command/report_command.rb:26:in `new'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/cli/command/report_command.rb:26:in `block in populate_reporter_with_smells'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/cli/command/report_command.rb:25:in `each'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/cli/command/report_command.rb:25:in `populate_reporter_with_smells'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/cli/command/report_command.rb:17:in `execute'
	from /usr/local/bundle/gems/reek-4.7.2/lib/reek/cli/application.rb:29:in `execute'
	from /usr/local/bundle/gems/reek-4.7.2/bin/reek:14:in `<top (required)>'
	from /usr/local/bundle/bin/reek:17:in `load'
	from /usr/local/bundle/bin/reek:17:in `<main>'
```

I would love it if reek told me what file it crashed on.

Anyway, thanks for all the hard work you do. Let me know if you need any info from me.

troessner:
Fixes #1262 - still wip though.

### DIFF
--- a/lib/reek/errors/encoding_error.rb
+++ b/lib/reek/errors/encoding_error.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require_relative 'base_error'
+
+module Reek
+  module Errors
+    # Gets raised when Reek is unable to process the source due to an EncodingError
+    class EncodingError < BaseError
+      ENCODING_ERROR_TEMPLATE = <<-MESSAGE.freeze
+        !!!
+        Source '%s' can not be processed by Reek due to an encoding error in the source file.
+
+        This is a problem that is outside of Reek's scope and should be fixed by you, the
+        user, in order for Reek being able to continue.
+        Check out this article for an idea on how to get started:
+        https://www.justinweiss.com/articles/3-steps-to-fix-encoding-problems-in-ruby/
+
+        Exception message:
+
+        %s
+
+        Original exception:
+
+        %s
+
+        !!!
+      MESSAGE
+
+      def initialize(origin:, original_exception:)
+        message = format(ENCODING_ERROR_TEMPLATE,
+                         origin,
+                         original_exception.message,
+                         original_exception.backtrace.join("\n\t"))
+        super message
+      end
+    end
+  end
+end
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -95,17 +95,25 @@ module Reek
     #
     # @return [Array<SmellWarning>] the smells found in the source
     #
-    # :reek:TooManyStatements { max_statements: 8 }
+    # :reek:TooManyStatements { max_statements: 6 }
     def run
       if source.valid_syntax? && syntax_tree
-        begin
-          examine_tree
-        rescue Errors::BaseError => exception
-          raise unless @error_handler.handle exception
-          []
-        end
+        examine_tree
       else
         SmellDetectors::Syntax.smells_from_source(source)
+      end
+    rescue StandardError => exception
+      wrapper = wrap_exception exception
+      raise wrapper unless @error_handler.handle wrapper
+      []
+    end
+
+    def wrap_exception(exception)
+      case exception
+      when Errors::BaseError
+        exception
+      else
+        Errors::IncomprehensibleSourceError.new(origin: origin, original_exception: exception)
       end
     end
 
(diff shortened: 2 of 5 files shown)


troessner:
@mvz can you think of something to put [here](https://github.com/troessner/reek/pull/1263/files#diff-f63572ef48cf20a9eac97e84f1474209R44) so that I can trigger an EncodingError in our specs? If you have no idea as well then I'd go ahead and create a bad source file like describe [here](https://stackoverflow.com/questions/6374756/why-do-i-get-an-invalid-byte-sequence-in-utf-8-error-reading-a-text-file) and would use that one.

mvz:
It may be hard to trigger that error because it may have to do with the locale set in the environment (see #1254).

How about using a double for the buffer and add a mock to have it raise some error? This would require some refactoring to be able to inject the buffer object, of course.

troessner:
Hmmm, the specs fail with

>>
        expected Exception with message matching /Looks like bad source/, got #<RSpec::Mocks::MockExpectationError: #<Double (anonymous)> received unexpected message :source= with ("Code that raises an EncodingError")> with backtrace:
         # ./lib/reek/source/source_code.rb:100:in `parse'
         # ./lib/reek/source/source_code.rb:33:in `initialize'

Which is weird. I mean, obviously this line [here](https://github.com/troessner/reek/pull/1263/files#diff-f9dae46a54ebb72c2c734ff15d80d26eR99):

```Ruby
buffer = Parser::Source::Buffer.new(origin, 1)
```

returns the proper which I have instructed to accept the "source" as method call via 

```Ruby
allow(buffer).to receive(:source) { raise EncodingError }
```

Not really sure what's going on here. @mvz do you have any idea?

mvz:
It turns out you need  to stub `:source=` instead of `:source`.

troessner:
Haha, what a stupid mistake of me, thanks for the hint.
I am in lovely Marocco for the next 3 weeks without laptop, could you possibly finish this one?

mvz:
Sure!

mvz:
I'm not entirely sure about this solution anymore (I know, I proposed it 😞). I thought we could do something like we do in `Examiner`, but `Examiner` actually does something quite complicated: It catches all exceptions and then either prints stuff or re-raises. We should definitely not do that in `SourceCode` as well. Ideally, we would catch errors in `SourceCode#parse` in `Examiner#run`, but in fact it is already executed inside `Examiner#initialize`. If we can delay calling `#parse` until either `#valid_syntax?` or `#syntax_tree` is called, that problem would be solved. That way, all tricky code runs inside `Examiner#run`, and we can just modify that to catch all errors happening inside it.

mvz:
I've updated Examiner to handle these errors properly during `#run`.

Because of #1254 I think we cannot yet say the encoding error needs to be fixed by the user.

troessner:
Very nice. Looks good to me. Wanna squash?

mvz:
I'll see if I can make Travis behave.

mvz:
Squashed and green.

PR #1277, opened 2017-12-02
Only show long error message when exception is handled

mvz:
This changes the error processing in Examiner so that raise wrapping error classes pick up the original exception in their `cause` attribute.

This makes error reporting when the error is not handled (e.g., when running specs during development) more compact.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -3,6 +3,12 @@ engines:
     enabled: true
   reek:
     enabled: true
+  duplication:
+    enabled: true
+    config:
+      languages:
+        ruby:
+          mass_threshold: 29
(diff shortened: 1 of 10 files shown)


troessner:
Did a quick pre-review, this looks very promising!

mvz:
Ready now. Not sure how to handle codeclimate 😕

troessner:
Couldnt you just silence those warnings like depicted [here](https://docs.codeclimate.com/v1.0/docs/disabling-individual-checks)?

mvz:
Oddly, I don't get the nice link there. However, I did manage to suppress this warning, so all is green now.

PR #1281, opened 2017-12-03
Unify node iterators

mvz:
This unifies `AST::Node#each_node` and `AST::Node#find_nodes`.

Fundamentally, each_node was recursive and never skipped root, while find_nodes was non-recursive and could skip root. These properties were not reflected in the method names.

The resulting `#each_node` has slightly more complex behavior than strictly desired. It is hoped that future refactoring can remedy this.

### DIFF
--- a/lib/reek/ast/node.rb
+++ b/lib/reek/ast/node.rb
@@ -34,54 +34,38 @@ module Reek
       end
 
       #
-      # Carries out a depth-first traversal of this syntax tree, yielding
-      # every Sexp of type `target_type`. The traversal ignores any node
-      # whose type is listed in the Array `ignoring`.
+      # Carries out a depth-first traversal of this syntax tree, yielding every
+      # Sexp of the searched for type or types. The traversal stops at any node
+      # whose type is listed in `ignoring`.
+      #
+      # If a type is searched for *and* listed in ignoring, it will be yielded
+      # but traversal will not continue to its children.
+      #
+      # If the root's type is ignored, traversal does not stop, unless the root
+      # is of a target type.
+      #
       # Takes a block as well.
       #
-      # target_type - the type to look for, e.g. :send, :block
-      # ignoring - types to ignore, e.g. [:casgn, :class, :module]
-      # blk - block to execute for every hit. Gets passed in the matching element itself.
+      # @param target_types [Symbol, Array<Symbol>] the type or types to look
+      #     for
+      # @param ignoring [Array<Symbol>] types to ignore
+      # @param blk block to execute for every hit. Gets passed in the
+      #     matching element itself.
       #
-      # Examples:
-      #   context.each_node(:send, [:mlhs]) do |call_node| .... end
-      #   context.each_node(:lvar).any? { |it| it.var_name == 'something' }
+      # @example
+      #   node.each_node(:send, [:mlhs]) do |call_node| .... end
+      #   node.each_node(:lvar).any? { |it| it.var_name == 'something' }
+      #   node.each_node([:block]).flat_map do |elem| ... end
       #
       # Returns an array with all matching nodes.
-      # TODO: without a block, this doesn't do what one might expect
-      def each_node(target_type, ignoring = [], &blk)
-        if block_given?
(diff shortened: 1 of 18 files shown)


lib/reek/ast/node.rb
```diff
@@ -35,52 +35,36 @@ def line
...
-      # Takes a block as well.
+      # every Sexp of the searched for type or types. The traversal stops at any node
+      # whose type is listed in `ignoring`.
       #
-      # target_type - the type to look for, e.g. :send, :block
-      # ignoring - types to ignore, e.g. [:casgn, :class, :module]
-      # blk - block to execute for every hit. Gets passed in the matching element itself.
+      # If a type is searched for *and*
+      # listed in ignoring, it will be yielded but traversal will not continue
+      # to its children.
```
chastell: Can we rewrap this comment? The current ragged edge looks like some docs were removed by accident.

```ruby
      # If a type is searched for *and* listed in ignoring, it will
      # be yielded but traversal will not continue to its children.
```

(Look, it wraps to bricktext!)

lib/reek/ast/sexp_extensions/methods.rb
```diff
@@ -30,8 +30,8 @@ def ends_with_bang?
         end
 
         def body_nodes(types, ignoring = [])
-          if body
-            body.find_nodes(types, ignoring)
+          if body && !ignoring.include?(body.type)
+            body.each_node(types, ignoring | types)
           else
             []
           end
```
chastell: I’m not sure whether I wouldn’t prefer the more explicit handling of edge cases:
```ruby
def body_nodes(types, ignoring = [])
  return [] if ignoring.include?(body.type)
  return [] unless body
  body.each_node(types, ignoring | types)
end
```
(In a better world `body` would be non-nilable and an ‘empty’ body would return `[]` on `each_node` – at that point the second `return` line could be dropped altogether.)

spec/reek/ast/node_spec.rb
```diff
@@ -25,8 +25,8 @@
         end
       end
 
-      it 'returns an empty array of ifs when no block is passed' do
-        expect(ast.each_node(:if, [])).to be_empty
+      it 'returns a enumerator when no block is passed' do
```
chastell: _an_ enumerator

mvz:
Yes, the entire `AST::Node` class is private. It looks like the [docs on rubydoc.info](http://www.rubydoc.info/gems/reek/Reek/AST/Node) aren't picking up the custom markers we have for that.

lib/reek/ast/node.rb
```diff
@@ -34,53 +34,38 @@ def line
...
+        return enum_for(:each_node, target_types, ignoring) unless block_given?
+
+        look_for(Array(target_types), ignoring, &blk)
       end
 
       def contains_nested_node?(target_type)
-        look_for_type(target_type) { |_elem| return true }
+        each_node(target_type) { |_elem| return true }
         false
       end
```
chastell: A quick test suggest
```ruby
      def contains_nested_node?(target_type)
        each_node(target_type).any?
      end
```
works just as well.

lib/reek/ast/node.rb
```diff
@@ -120,22 +105,22 @@ def source
       protected
 
       # See ".each_node" for documentation.
-      def look_for_type(target_type, ignoring = [], &blk)
+      def look_for(target_types, ignoring, &blk)
```
chastell: I’m trying to understand why do we need both `look_for` and `look_for_recurse` – they only differ in that `look_for` returns only if the current node is _both_ of a target type _and_ ignored type, while `look_for_recurse` returns as soon as the current node’s type is one of the ignored ones. I tried to end up with just one of those two, but tests blew up. 😉 

Is there a way to conflate those two into one or is this an inescapable difference?

lib/reek/ast/node.rb
mvz: Nice!

chastell:
This is a great improvement, thanks! Merging!

PR #1282, opened 2017-12-13
Bump RuboCop and fix new offences

chastell:
I’m partial to disabling IfUnlessModifier (at least in this case).

I’m also totally open to disabling FormatStringToken, but I think this actually adds to the readability and portability of the format strings (the `format` calls no longer need to maintain strict positional order of the template format sequences).

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -9,7 +9,7 @@ plugins:
     config:
       languages:
         ruby:
-          mass_threshold: 29
+          mass_threshold: 35
 exclude_patterns:
   - coverage/
   - docs/
--- a/Gemfile
+++ b/Gemfile
@@ -16,8 +16,8 @@ group :development do
   gem 'yard',          '~> 0.9.5'
 
   if RUBY_VERSION >= '2.3'
-    gem 'rubocop',       '~> 0.51.0'
-    gem 'rubocop-rspec', '~> 1.20.0'
+    gem 'rubocop',       '~> 0.52.0'
+    gem 'rubocop-rspec', '~> 1.20'
   end
 
   platforms :mri do
--- a/lib/reek/configuration/configuration_file_finder.rb
+++ b/lib/reek/configuration/configuration_file_finder.rb
@@ -20,8 +20,8 @@ module Reek
     module ConfigurationFileFinder
       TOO_MANY_CONFIGURATION_FILES_MESSAGE = <<-MESSAGE.freeze
 
-        Error: Found multiple configuration files %s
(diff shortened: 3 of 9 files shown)


chastell:
(I’ll squash the commits once we make a call on whether either or both go in.) 😉

troessner:
I'm all up for those changes especially the format ones.

mvz:
Same here. The format string changes really aid readability.

chastell:
Excellent. 😄 

How should I address [the Code Climate grievances](https://codeclimate.com/github/troessner/reek/pull/1282)? I’d rather keep the repetition than start constructing templates from template fragments, but I’m not sure how to tell Code Climate to chill about this particular duplication.

mvz:
In `.codeclimate.yml` you can up the `mass_threshold` setting. Or we can disable this check entirely.

mvz:
Oh, it helps to run codeclimate locally to find the correct setting.

troessner:
Check out the PR [here](https://github.com/troessner/reek/pull/1277) for  the CC problem.

chastell:
> Oh, it helps to run codeclimate locally to find the correct setting.

I’m sorry, I tried to figure out and couldn’t. 😕  How can I do this? 🙏

mvz:
First, you need to have Docker installed.

Then, in the Reek directory so you pick up the config, install codeclimate and fetch its engines:
```
gem install codeclimate
codeclimate engines:install
```

Then, to run the duplication analysis:
```
codeclimate analyze -e duplication
```

chastell:
@mvz Thanks so much for your patience, but the codeclimate gem (at least the newest version, 0.70.6) does not install the `codeclimate` executable. Which version does the executable come from?

mvz:
Oh, I forgot about that. You'll need to follow the instructions here: https://github.com/codeclimate/codeclimate#packages

chastell:
Whew, 3 GB of `/var/lib/docker` later I got this to work. Thanks for the patience, ready to merge!

troessner:
@chastell how about updating our Readme so we dont forget the single steps?

mvz:
@troessner steps for what? codeclimate? Normal users shouldn't need to run codeclimate locally.

troessner:
No, for us. If "what to do when running into problems with codeclimate?" was perfectly documented it wouldn't have been such a pain in the ass to get it up and running for @chastell - or am I missing something?

mvz:
Ah, that makes sense. Perhaps we can improve codeclimate's documentation and then point to that.

PR #1284, opened 2017-12-26
Travis: Add Ruby 2.5 and peg JRuby to newest 9.1

chastell:
I tried to go with `jruby-9.1` and `jruby-9.1.15`, but they  both seem to resolve to the (non-existent) `jruby-9.1.15.0200`.

It seems `2.5` installs CRuby 2.5.0-preview1 which subsequently blows up; I’ll wait till Travis supports 2.5.0 and retry then.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 dist: trusty
 cache: bundler
 language: ruby
+before_install: gem update --system
 bundler_args: --without debugging
 script: bundle exec rake ci
 rvm:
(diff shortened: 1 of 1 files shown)


chastell:
it’s downright fascinating – once Travis started resolving `2.5` to `2.5.0` the problem with `2.5.0-preview1` went away but now it breaks with a missing `rake`, despite it’s clearly there in the `development` group 🤔

chastell:
Ah, the good old ‘let’s upgrade RubyGems by hand’ is here again! 🎉  Seems fixed.

JRuby 9.1.15 seems to have some of its own ideas on how relative should relative `Pathnames` be. 🤷‍♂️

Issue #1137, opened 2016-11-12
False positives of ModuleInitialize


PeterCamilleri:
Two closely related issues with the following fragment of code:

```
module Mysh
  def self.reset_host
    exec_class = Class.new do
      #Set up a new execution environment
      #<br>Endemic Code Smells -- False positive.
      #* :reek:ModuleInitialize
      def initialize
        #Code deleted for brevity.
      end

      #Balance of class definition omitted.
    end

    #Balance of reset_host method deleted for brevity.
end
```
(1) This generates the following warning:

```
lib/mysh/expression.rb -- 1 warning:
  [9]:ModuleInitialize: Mysh has initialize method [https://github.com/troessner/reek/blob/master/docs/Module-Initialize.md]
1 total warning
```
even though it is a class initialize method and not a module initialiaze method.

(2) The :reek:ModuleInitialize directive is ignored. The only way to silence this report is to turn it off in the project's reek config file. I think this is a poor option since it prevents the detection of legitimate code smells.

troessner:
Yeah, that's not something we can detect at the moment.
The AST for dynamically creating classes is vastly different than for creating it "regularly" (if there is such a thing as "regular" with Ruby).

```
 ruby-parse -e 'module M; def m; Class.new; end; end'
(module
  (const nil :M)
  (def :m
    (args)
    (send
      (const nil :Class) :new)))
```

vs.

```
ruby-parse -e 'module M; class C; end; end'
(module
  (const nil :M)
  (class
    (const nil :C) nil nil))
```

Or in other words, we'd have to detect `Class.new` as well. That might open up a few more edge cases as well. WDYT @mvz ?

>> (2) The :reek:ModuleInitialize directive is ignored. The only way to silence this report is to turn it off in the project's reek config file. I think this is a poor option since it prevents the detection of legitimate code smells.

You have to put the comment above the module like this:

```Ruby
      # :reek:ModuleInitialize
      module Alfa
        def initialize; end
      end
```

That's because the smell detector works on module level, not on method level. 
Something we failed to document properly in our [detector docs](https://github.com/troessner/reek/blob/master/docs/Module-Initialize.md), I'll update this and all similar pages over the next days.

mvz:
We are actually detecting `Class.new` (and `Struct.new`) in a few cases. It would be nice if we could consistently handle them the same way as `class Foo`. It will be tricky, since we assume the 'regular' way in quite a few places.

[...]

PR #1287, opened 2018-01-05
Do not detect ModuleInitialize for nested dynamic classes

mvz:
#1137

### DIFF
--- a/lib/reek/context/module_context.rb
+++ b/lib/reek/context/module_context.rb
@@ -43,9 +43,9 @@ module Reek
         AttributeContext
       end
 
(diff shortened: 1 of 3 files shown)


chastell:
LGTM, thanks! I reran the rogue Travis jobs and they’re all green now. 🙌

chastell:
Gaah, there should probably be a matching `CHANGELOG.md` entry. @mvz can you add it? Sorry!

mvz:
Well, we do tend to only add those when we release, but I'll make a proposal.

PR #1298, opened 2018-03-26
Change `before_filter` to `before_action`

edenthecat:
Updated to use before_action instead of before_filter so that this doc is consistent with Rails >=5.1


troessner:
Good catch, thanks!

PR #1299, opened 2018-04-02
Loosen dependency on rainbow

mvz:
Reek works fine with rainbow 2.x. This change resolves a conflict with
dependencies of pronto.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,9 @@ sudo: false
 dist: trusty
 cache: bundler
 language: ruby
-before_install: gem update --system
+before_install:
+  - gem update --system
+  - gem install bundler
 bundler_args: --without debugging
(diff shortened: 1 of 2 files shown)


troessner:
Lgtm. Oh, travis ;)

chastell:
Hrmgh, supposedly the `before_install: gem update --system` we have in `.travis.yml` should work around the build issue, but it apparently doesn’t. :disappointed:

@mvz from reading around [the relevant issue](https://github.com/travis-ci/travis-ci/issues/8969): can you try rebuilding with `cache: bundler` dropped? If that doesn’t help, can you try adding `gem install bundler` to `before_install`?

mvz:
Fixed the build :tada:. Adding `gem install bundler` helped :confused:.

PR #1302, opened 2018-05-16
Remove unneeded gender reference to a method

tibbon:
Methods don't have genders, so I'm proposing removing the reference to it as `her`.

### DIFF
--- a/docs/Prima-Donna-Method.md
+++ b/docs/Prima-Donna-Method.md
@@ -18,7 +18,7 @@ An exclamation mark in method names means (the explanation below is taken from
 > dangerous version of exit. flatten! is the dangerous version of flatten. And
 > so forth.
 
-Such a method is called _Prima Donna Method_ if and only if her non-bang
+Such a method is called _Prima Donna Method_ if and only if the non-bang
 version does not exist and this method is reported as a smell.
 
 ## Example


mikegee:
Unfortunately, "Prima Donna" remains gendered. So maybe we should change the name of the whole smell?

tibbon:
I agree, the smell name overall is highly gendered and I figured that was a larger conversation.

If we do want to address that, it might be better in another PR, but a few suggestions for a better smell name:

- **Unmatched Mutate Method**
- **False Mutate Method** (Not all methods without `!` are free of side effects)
- **Celebrity Method** 
- **False Warning Method**
- **Uncertain Danger Method**
- **Whiz-bang Method** (has a neat whimsy, and points to the fact that there's a `!` in the offending method)
- **Alternate Method**
- **Scary Method**

mvz:
Hey @tibbon, thanks for your patch. I definitely agree with the proposed change to the wording. I'll leave this unmerged for a bit so we have a place to discuss a possible new name for this smell.

troessner:
I'm fine with changing this as well. I like whiz-bang :)

troessner:
>> I like "whiz bang" too, but this may also be a good opportunity to change the name to something descriptive

I believe it is quite descriptive :)

mvz:
(As an aside, it seems the term "prima donna" can, in the English language, refer to both men and women, so its use here is not so strange. I still don't like the smell name very much, since it's not very descriptive.)

mvz:
> I believe it is quite descriptive :)

Well, it mentions 'bang', I'll grant you that :stuck_out_tongue:.

mikegee:
"Missing Safe Method"?

mvz:
@mikegee nice. It reverses the naming logic: Instead of naming what is there, it names what is, well, missing.

tibbon:
I agree "Missing Safe Method" is quite descriptive.

troessner:
>> Along that line, I'd suggest SelfImportantMethod, but there might be potential confusion with keyword self

Yep, that would be problematic probably :)

I think we should go with "Missing Safe Method".

>> If a decision to change the name is made, I'd be happy to make an additional commit for that.

That would be great @tibbon - I'd probably would be better though if it was a new pull request since now the scope has changed significantly from a fix patch to a minor version update including a different branch name (I'd suggest "change-PrimaDonnaMethod-to-MissingSafeMethod").

mvz:
> I'd probably would be better though if it was a new pull request

Agreed. Let's merge this one first.

PR #1304, opened 2018-05-17
Change PrimaDonnaMethod to MissingSafeMethod

tibbon:
Renames all instances of PrimaDonnaMethod to MissingSafeMethod for greater clarity and a more gender neutral approach.

If configuration options have been set for PrimaDonnaMethod, they will need to now reference MissingSafeMethod. I think configuration will fail silently in this case, but I'm unsure how this fits in with the versioning protocol, and also unsure if it's ideal to have a backwards compatible alias, as least until the next major version. I didn't otherwise intentionally make any changes to functionality. 

Prior discussion around naming can be found in https://github.com/troessner/reek/pull/1302


lib/reek/smell_detectors/missing_safe_method.rb
```diff
@@ -48,20 +48,20 @@ def self.contexts # :nodoc:
...
         context.node_instance_methods.select do |method_sexp|
-          prima_donna_method?(method_sexp)
+          missing_safe_method?(method_sexp)
         end.map do |method_sexp|
           name = method_sexp.name
           smell_warning(
             context: context,
             lines: [method_sexp.line],
-            message: "has prima donna method '#{name}'",
+            message: "has missing safe method '#{name}'",
```
mvz: The missing safe method is `name` without the bang. Perhaps `has dangerous method '#{name}' without a safe version`.

defaults.reek
```diff
@@ -44,6 +44,9 @@ LongYieldList:
 ManualDispatch:
   enabled: true
   exclude: []
+MissingSafeMethod:
```
mvz: This trips up codeclimate at the moment. I'm not sure what we should do. This would probably be an issue whenever we add a new smell anyway.

lib/reek/smell_detectors/missing_safe_method.rb
tibbon: I was thinking the wording there (and in a few other places) was awkward, but was attempting to limit the scope of changes I made. Can amend this and other awkward phrasings.

troessner: I like @mvz suggestion :)

troessner:
>> and also unsure if it's ideal to have a backwards compatible alias, as least until the next major version.

Not a huge fan of this especially given that we rarely have the need for this, this pull request being the exception ;)

>> I think configuration will fail silently in this case, but I'm unsure how this fits in with the versioning protocol

We are doing semantic versioning so this would be a breaking change and thus a major patch. 

I think this warrants a major version upgrade. wdyt @mvz ?

troessner:
@mvz this would also allow us get all those [breaking changes](https://github.com/troessner/reek/labels/breaking-change) out <3

tibbon:
Higher numbers are better. I like this idea!

mvz:
@tibbon can you rebase on current master? That should at least clean up the Travis build failures.

tibbon:
@mvz Rebased!

tibbon:
I see the Travis file that was failing and have just pushed a fix for it. There was a cucumber test I somehow overlooked.

troessner:
Merged, thanks!

PR #1306, opened 2018-05-18
List all smells.

### LINKED ISSUE #1165: Introduce a --list option
As per the [discussion here](https://github.com/troessner/reek/pull/1162#discussion_r96138695), it would be nice to have a `--list` option to print all the smells to aid exploration and configuration without having to visit the documentation.

The first question is what information is interesting to the user, and how much of it to display. 😀

troessner:
Fixes #1165


Drenmi:
Might be subjective, but I find it easier to parse lists when the items have a leading `-`. 🙂

lib/reek/cli/options.rb
```diff
@@ -4,6 +4,7 @@
 require 'rainbow'
 require_relative '../version'
 require_relative 'status'
+require_relative '../../../lib/reek/detector_repository'
```
mvz: You can leave out `../../lib/reek/`.

lib/reek/cli/options.rb
```diff
@@ -213,6 +214,15 @@ def set_utility_options
           puts parser
           exit
         end
+        parser.on_tail('-l', '--list', 'List all available smell detectors') do
+          puts 'All available smell detectors:\n'
+          DetectorRepository.smell_types.map do |full_name|
+            full_name.to_s.split('::').last
+          end.each(&method(:puts))
```
mvz: I'm not liking this way of printing an entire array. How about `puts the_array`?

lib/reek/cli/options.rb
```diff
@@ -213,6 +214,15 @@ def set_utility_options
           puts parser
           exit
         end
+        parser.on_tail('-l', '--list', 'List all available smell detectors') do
+          puts 'All available smell detectors:\n'
+          DetectorRepository.smell_types.map do |full_name|
+            full_name.to_s.split('::').last
```
mvz: Don't we have a method on `SmellDetector` that gives the short name already?

lib/reek/cli/options.rb
```diff
@@ -213,6 +214,15 @@ def set_utility_options
           puts parser
           exit
         end
+        parser.on_tail('-l', '--list', 'List all available smell detectors') do
+          puts 'All available smell detectors:\n'
+          DetectorRepository.smell_types.map do |full_name|
```
mvz: For improved testability, maybe move the generation of the list to another class?

lib/reek/cli/options.rb
troessner: Whoops, that's what I get for using Rubymine's autocompletion :)

lib/reek/cli/options.rb
troessner: No problem, I can change that.

lib/reek/cli/options.rb
troessner: Good catch! And we do [indeed](https://github.com/troessner/reek/blob/master/lib/reek/smell_detectors/base_detector.rb#L94) ;)

lib/reek/cli/options.rb
troessner: How about creating

>> DetectorRepository#print_available_detectors

?

mvz: Perhaps `DetectorRepository.available_detector_names`, so you can do `puts DetectorRepository.available_detector_names`?

troessner: Yeah, good idea. I'd suggest to also rename `DetectorRepository.smell_types` to `detectors` since the method returns ..... detectors and not smell_types. We kind of fucked up our naming a bit here and there ;)

mvz: Yes, that makes sense as well :-).

troessner: Fixed. I'll fix DetectorRepository.smell_types in another pull request so I don't blow up this one.

troessner:
Adressed all comments <3

mvz:
Can you rebase this on current master?

troessner:
Done.

troessner:
Should be green now.

PR #1307, opened 2018-05-18
Drop Ruby 2.1 support.

### LINKED ISSUE #1285: Drop support for Ruby < 2.3
It's time 😄

troessner:
Fixes #1285


troessner:
For v5 ;)

mvz:
We should also update the gemspec and the rubocop configuration.

troessner:
Good catch, done :)

Issue #1248, opened 2017-08-27
Require argument for AppConfiguration#from_path and #from_hash


mvz:
I don't like that these have default arguments.

* For `.from_hash`: Why not pass `{}`?
* For `.from_path`: It looks odd to call this with no argument: Which path is it, then? I would prefer a method `.from_default_path` or something.

mvz:
I'm going to take a stab at this one now.

PR #1310, opened 2018-05-20
Update AppConfiguration API

mvz:
For #1248. This fixes the API as envisioned there.

There are some remaining issues are with this code:

- AppConfiguration has a hard-coded dependency on ConfigurationFileFinder. This is hard to break since we want to be able to easily say 'use the default configuration given the current working directory'.
- ConfigurationFileFinder.find smells of ControlParameter.

However, these are internal issues that don't affect the public API, so they can be picked up later.

### DIFF
--- a/lib/reek/configuration/app_configuration.rb
+++ b/lib/reek/configuration/app_configuration.rb
@@ -18,18 +18,26 @@ module Reek
       EXCLUDE_PATHS_KEY = 'exclude_paths'.freeze
       DIRECTORIES_KEY = 'directories'.freeze
 
-      # Instantiate a configuration via given path, or the default path.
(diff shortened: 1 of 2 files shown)


lib/reek/configuration/app_configuration.rb
```diff
@@ -79,6 +83,10 @@ def load_values(configuration_hash)
         end
       end
 
+      def initialize(configuration_hash: nil)
```
troessner: I'd consider encoding the data type "hash" into the variable name an antipattern in Ruby.

mvz: Fixed. Ready for another round.

troessner:
Looks good to me, merged!

PR #1311, opened 2018-05-21
Name detectors by just their name

### LINKED ISSUE #1163: Deprecate / Drop usage of fully qualified detector names for API users
Coming from #1114 

Let's drop this (quoting from [here](https://github.com/troessner/reek/blob/master/docs/API.md)):

```
Instead of the smell detector names you can also use the full detector class in your configuration hash, for example:

config_hash = { Reek::SmellDetectors::IrresponsibleModule => { 'enabled' => false } }
```

This was an error to begin with since it exposes way to much of our internals to the world.
Just using the smell names (that are very unlikely to change) would have been fine.

mvz:
Fixes #1163.

### DIFF
--- a/lib/reek/spec.rb
+++ b/lib/reek/spec.rb
@@ -65,7 +65,7 @@ module Reek
     # raise an ArgumentError to give you a hint that you passed something that doesn't make
     # much sense.
     #
-    # @param smell_type [Symbol, String, Class] The "smell type" to check for.
+    # @param smell_type [Symbol, String] The "smell type" to check for.
     # @param smell_details [Hash] A hash containing "smell warning" parameters
     #
     # @example Without smell_details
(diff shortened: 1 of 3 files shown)


lib/reek/configuration/configuration_validator.rb
```diff
@@ -10,26 +10,14 @@ module ConfigurationValidator
...
-          true
-        when String
-          begin
-            Reek::SmellDetectors.const_defined? key
-          rescue NameError
-            false
-          end
-        end
+        Reek::SmellDetectors.const_defined? key
+      rescue NameError
```
Drenmi: If `key` is now expected to be a string or a symbol, is it still necessary to account for references to uninitialized constants? 🤔 Would make more sense to me if the `false` was something like exiting with a deprecation message.

(I'm not sure what this did in the original implementation, since `key` was already known to be a string.)

troessner:
Very nice!

lib/reek/configuration/configuration_validator.rb
mvz: Key may have a form like `lib/foo.rb` in this case so a NameError could still be thrown.

Drenmi: Thanks for clarifying! That didn't cross my mind. 🙂

mvz: It didn't cross mine either, so I removed it and the specs failed :slightly_smiling_face:.

PR #1315, opened 2018-05-29
Add yml extension to Reeks configuration file.

### LINKED ISSUE #1232: .reek file should use .yml extension
The default configuration file should use a `.yml` extension so intelligent text editors can provide YAML syntax checking and functionality ([a la Rubocop](http://rubocop.readthedocs.io/en/latest/configuration/))

My particular use case: I generate a reek TODO file for my project and want to comment out various sections so I can work on those issues. Being able to select multiple lines in RubyMine and comment them out with Command / would be helpful instead of having to do that manually (and error prone)

troessner:
Fixes #1232 

What this PR does is:

1.) It propagates using ".reek.yml" instead of just ".reek" while allowing both ways.
2.) It renames "defaults.reek" to ""defaults.reek.yml"


troessner:
Ready for review!

docs/API.md
```diff
@@ -89,7 +89,7 @@ XMLReport
 
 ## Configuration
 
-Given you have the following configuration file called `config.reek` in your root directory:
+Given you have the configuration file called `.reek.yml` in your root directory:
```
mvz: I think it's clearer with the word 'following' included, or some other indication that what follows is the contents of `.reek.yml`.

mvz:
Since we're making breaking changes anyway, and we're actually changing the contents of the configuration file as well, why not remove support for the plain `.reek` extension?

Also, how about picking up just `.reek.yml` by default, and only swiching to another file if `--config some_file.yml` is passed?

docs/API.md
troessner: ok

troessner:
>> Also, how about picking up just .reek.yml by default, and only swiching to another file if --config some_file.yml is passed?

I like it! Updating....

troessner:
Ready for another review! Oh boy, what a battle ;)

features/configuration_loading.feature
```diff
@@ -30,19 +30,71 @@ Feature: Offer different ways how to load configuration
...
+
+  Scenario: Two opposing configuration files but we ignore the first one due to an improper name
+    Given the smelly file 'smelly.rb' in a subdirectory
+    And a configuration file 'partial_mask.reek' in a subdirectory
+    And a configuration file '.reek.yml'
     When I run "reek smelly.rb" in a subdirectory
+    Then it reports no errors
+    And it succeeds
+
+  Scenario: Do not use the default config file when we explicitly specify one configuration file
```
mvz: This is essentially the same scenario as `Two opposing configuration files and we stop after the first one` above, isn't it?

features/configuration_loading.feature
```diff
@@ -30,19 +30,71 @@ Feature: Offer different ways how to load configuration
...
+      """
+    When I run reek -c config.reek smelly.rb
     Then the exit status indicates smells
     And it reports:
       """
       smelly.rb -- 1 warning:
         [5]:UncommunicativeVariableName: Smelly#x has the variable name 'y' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Variable-Name.md]
       """
+
+  Scenario: Prefer the default config file over other configuration files
```
mvz: .. and I think this is the same as `Two opposing configuration files but we ignore the first one due to an improper name`.

troessner: There is a subtle difference: Both other scenarios talk about subdirectories. But overall I think you're right, the scenarios are too close together to warrant their existence. I removed the "old ones" and kept the new ones since they are more expressive and make it more obvious what happens.

mvz:
Ok, let's merge this!

PR #1318, opened 2018-06-02
Move directory directives under a special key.

### LINKED ISSUE #1078: Move directory directives under a special key
Looking into #1062 i realize that it is very hard to validate our configuration because we  deliberately made it very ambiguous if something is a directory directive or a smell detector.
Validating our configuration AND handling most edgecases will be a nightmare.
I think we can make it easier:
- introduce a new top level key called "directories" analoguous to [exclude_paths](https://github.com/troessner/reek#configuration-options) under which we put the directives
- the change should be trivial for users but would make our lifes easier. 
- users can still fuck up the config but since we can make solid assumptions (e.g. everything under "directories" is a directory directive) we can show the users far better and helpful error messages that will help him figure it out on his own. E.g. ""foo/bar/baz is not a valid smell detector. Are you trying to set up a directory directive? Please see...."
- this would be a breaking change of course

Wdyt?

troessner:
Fixes #1078

### DIFF
--- a/spec/reek/configuration/app_configuration_spec.rb
+++ b/spec/reek/configuration/app_configuration_spec.rb
@@ -26,19 +26,21 @@ RSpec.describe Reek::Configuration::AppConfiguration do
     end
 
     let(:directory_directives_value) do
-      { 'samples/three_clean_files' =>
-        { 'UtilityFunction' => { 'enabled' => false } } }
+      { described_class::DIRECTORIES_KEY =>
+        { 'samples/three_clean_files' =>
+          { 'UtilityFunction' => { 'enabled' => false } } } }
     end
 
     let(:exclude_paths_value) do
-      ['samples/two_smelly_files',
-       'samples/source_with_non_ruby_files']
+      { described_class::EXCLUDE_PATHS_KEY =>
+        ['samples/two_smelly_files',
+         'samples/source_with_non_ruby_files'] }
     end
 
     let(:combined_value) do
       directory_directives_value.
         merge(default_directive_value).
-        merge('exclude_paths' => exclude_paths_value)
+        merge(exclude_paths_value)
     end
 
     describe '#from_path' do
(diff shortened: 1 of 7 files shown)


troessner:
Ready for review!

spec/reek/configuration/app_configuration_spec.rb
```diff
@@ -99,11 +102,10 @@
...
       let(:configuration_as_hash) do
-        {
-          directory => { TooManyStatements: { max_statements: 8 } },
-          :IrresponsibleModule => { enabled: false },
-          :TooManyStatements => { max_statements: 15 }
-        }
+        { described_class::DIRECTORIES_KEY =>
+            { directory => { TooManyStatements: { max_statements: 8 } } },
+          'IrresponsibleModule' => { enabled: false },
+          'TooManyStatements' => { max_statements: 15 } }
```
mvz: I'm not sure why RuboCop isn't complaining but the formatting here is a little off.

spec/reek/configuration/app_configuration_spec.rb
```diff
@@ -99,11 +102,10 @@
...
 
       let(:configuration_as_hash) do
-        {
-          directory => { TooManyStatements: { max_statements: 8 } },
-          :IrresponsibleModule => { enabled: false },
-          :TooManyStatements => { max_statements: 15 }
-        }
+        { described_class::DIRECTORIES_KEY =>
+            { directory => { TooManyStatements: { max_statements: 8 } } },
+          'IrresponsibleModule' => { enabled: false },
```
mvz: Is there a particular reason for changing the key from symbol to string here?

(I notice that we may have an inconsistency here, in that symbols work for the smell types, but perhaps not for the other keys).

troessner: >> Is there a particular reason for changing the key from symbol to string here?

No, good point, changed it back.

spec/reek/configuration/app_configuration_spec.rb
troessner: Moved around the brackets a bit - better?

mvz: Yes, better!

troessner:
@mvz I also added two comments to the README so we do not confuse Reek 4 users that come back to this documentation to look stuff up:
* [comment nr. 1](https://github.com/troessner/reek/pull/1318/files#diff-04c6e90faac2675aa89e2176d2eec7d8R317)
* [comment nr. 2](https://github.com/troessner/reek/pull/1318/files#diff-04c6e90faac2675aa89e2176d2eec7d8R347)

We can then remove those comments in a year or so. Normally this would be the point to introduce version dependent documentation (apart from the official release notes) but I'd like to stay away from this as long as possible since this will complicate the documentation quite a bit. Since this seems to be the only breaking change that might confuse returning Reek 4 users I think we're good.

PR #1308, opened 2018-05-18
Drop legacy code comment separator.

troessner:
Fixes #1237

PR #1322, opened 2018-06-05
Use strings not regexes in our configuration file.

### LINKED ISSUE #1321: Use strings not regexes in our configuration file.
Coming from #1317  I suggest to drop the ruby specific extensions to yaml. 
This would mean to do this:

```yaml
UncommunicativeParameterName:
  reject:
  - foo
  - bar
```

rather than this:

```yaml
UncommunicativeParameterName:
  reject:
  - !ruby/regexp /foo/
  - !ruby/regexp /bar/
```

Turns out the necessary changes to our codebase are quite minimal, see the linked PR.

troessner:
Fixes #1321


features/configuration_files/reject_setting.feature
```diff
@@ -16,43 +16,8 @@ Feature: `reject` configuration setting
...
-    """
-    smelly.rb -- 5 warnings:
-      [4]:UncommunicativeMethodName: Dummy#awesome_helper has the name 'awesome_helper' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Method-Name.md]
-      [6]:UncommunicativeMethodName: Dummy#little_helper has the name 'little_helper' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Method-Name.md]
-      [2]:UncommunicativeModuleName: Dummy has the name 'Dummy' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Module-Name.md]
-      [4]:UncommunicativeParameterName: Dummy#awesome_helper has the parameter name 'solid_argument' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Parameter-Name.md]
-      [6]:UncommunicativeParameterName: Dummy#little_helper has the parameter name 'nifty_argument' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Parameter-Name.md]
-    """
-
-  Scenario: reject regexes as list and as single item
```
troessner: The git diff is very unfortunate here - very hard to see what's going on here. Basically we had the same scenario twice, only the second time with "reject regexes as list and as single item". I just removed the now obsolete 2n scenario.

lib/reek/cli/options.rb
```diff
@@ -13,9 +13,9 @@ module CLI
     #
     # See {file:docs/Command-Line-Options.md} for details.
     #
-    # :reek:TooManyInstanceVariables: { max_instance_variables: 12 }
-    # :reek:TooManyMethods: { max_methods: 18 }
-    # :reek:Attribute: { enabled: false }
```
troessner: I missed those occurences and about 20 others in #1308 big time.

lib/reek/configuration/default_directive.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative './configuration_validator'
```
troessner: Boyscout rule.

spec/reek/code_comment_spec.rb
```diff
@@ -161,13 +161,6 @@
           build(:code_comment, comment: comment)
         end.not_to raise_error
       end
-
-      it 'does not raise on regexps' do
```
troessner: Now obsolete.

troessner:
Ready for review!

docs/Uncommunicative-Method-Name.md
```diff
@@ -21,22 +21,24 @@ Reek's _Uncommunicative Method Name_ detector supports the
 
 | Option         | Value       | Effect  |
 | ---------------|-------------|---------|
-| `reject` | array of regular expressions or strings | The set of patterns / names that Reek uses to check for bad names. Defaults to `[/^[a-z]$/, /[0-9]$/, /[A-Z]/]`. |
-| `accept` | array of regular expressions or strings | The set of patterns / names that Reek will accept (and not report) even if they match one of the `reject` expressions. |
+| `reject` | array of strings | The set of names that Reek uses to check for bad names. Defaults to `[/^[a-z]$/, /[0-9]$/, /[A-Z]/]`. |
```
mvz: This is now a little confusing since we say they're strings but the default is a bunch of regexes.

spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -97,6 +97,20 @@
...
+    context 'with strings for accept and reject settings' do
+      it 'properly converts them' do
+        expected = {
+          'UnusedPrivateMethod' => { 'exclude' => [/exclude regexp/] },
+          'UncommunicativeMethodName' => { 'reject' => [/reject regexp/], 'accept' => [/accept regexp/] },
+          'UncommunicativeModuleName' => { 'reject' => [/reject regexp/], 'accept' => [/accept regexp/] },
+          'UncommunicativeParameterName' => { 'reject' => [/reject regexp/], 'accept' => [/accept regexp/] },
+          'UncommunicativeVariableName' => { 'reject' => [/reject regexp/], 'accept' => [/accept regexp/] }
+        }
+        configuration = described_class.load_from_file(CONFIG_PATH.join('accepts_and_rejects.reek'))
```
mvz: I think we should specifically document the behavior if people use special regex characters in these strings. Users may  want to match `foo` but not `food`. Can they use `foo$` in this case?

troessner: >>  Users may want to match foo but not food

The way this works now, no, they can't. Is this a loss? I don't think so. Looking at the way we use disabling smells in our own codebase I can see only 1 case where we actually use regexes:

>> # :reek:UnusedPrivateMethod: { exclude: [ !ruby/regexp /process_/ ] }

and in this case the regex wouldn't have even been necessary.

My point is that while I can see this might be useful in some cases, I think those are edge cases and the majority of users will be fine with us removing this since they don't use it.

I think this is a small price to pay for having a much simpler configuration set up.

Note that you can still achieve this:

>> Users may want to match foo but not food

quite easily - you'll just have to spell it out.

docs/Uncommunicative-Method-Name.md
troessner: Good point - what do you think about actually saying:

>> By default Reek rejects one-letter names, names ending in a number or names that have a capital letter in them

and then updating the other pieces of documentation that are similar to this in a similar fashion?

spec/reek/configuration/configuration_file_finder_spec.rb
troessner: Yes :)
With "spelling it out" i meant: If you have "foo" and "food" then "foo" would match both. This definitely might be a huge problem.
Maybe implement something along @chastell idea 

```
I’m not sure how backward-(in)compatible would be to try to guess whether a given string is a name or a regex, but maybe we could just go with a simple heuristic to the tune of strings starting and ending with / are always regexes.
```

right away?

mvz: Yes, go for it!

troessner:
Ready for review! The last step before releasing Reek 5 <3

spec/reek/configuration/configuration_file_finder_spec.rb
```diff
@@ -97,6 +97,43 @@
       configuration = described_class.load_from_file(CONFIG_PATH.join('full_mask.reek'))
       expect(configuration).to eq(sample_configuration_loaded)
     end
+
+    context 'with exclude, accept and reject settings' do
+      let(:configuration) { described_class.load_from_file(CONFIG_PATH.join('accepts_rejects_and_excludes.reek')) }
```
mvz: Shouldn't this be a .yml filename?

defaults.reek
```diff
@@ -90,35 +90,35 @@ UncommunicativeMethodName:
   enabled: true
```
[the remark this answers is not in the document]
mvz: Are you sure? I just moved it back and there were no problems.

lib/reek/cli/options.rb
mvz: How come these don't show up as smells when running reek on itself? :confused:

mvz: Ah! It just ignores everything after the third `:`.

defaults.reek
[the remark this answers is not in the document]
troessner: Weird. I can't reproduce this. I moved it back.

spec/reek/configuration/configuration_file_finder_spec.rb
troessner: Good point, renamed it.

troessner:
Addressed all comments - lgtm?

mvz:
Once Travis is green, :shipit:!

troessner:
Had forgotten to push one line, should work in 5 min ;)

PR #1329, opened 2018-06-10
Fix our documentation until we release Reek 5.

### LINKED ISSUE #1327: Documentation: default config given in README has the wrong name.
In the README, it says that the default name for the config file is ".reek.yml", but apparently it's anything ending in ".reek".

troessner:
Fixes #1327

### DIFF
--- a/docs/Rake-Task.md
+++ b/docs/Rake-Task.md
@@ -33,7 +33,7 @@ An more sophisticated rake task that would make use of all available configurati
 ```Ruby
 Reek::Rake::Task.new do |t|
   t.name          = 'custom_rake' # Whatever name you want. Defaults to "reek".
-  t.config_file   = 'config/.reek.yml' # Defaults to nothing.
+  t.config_file   = 'config/config.reek' # Defaults to nothing.
   t.source_files  = 'vendor/**/*.rb' # Glob pattern to match source files. Defaults to lib/**/*.rb
   t.reek_opts     = '-U'  # Defaults to ''. You can pass all the options here in that are shown by "reek -h"
   t.fail_on_error = false # Defaults to true
--- a/README.md
+++ b/README.md
@@ -96,7 +96,8 @@ demo.rb -- 2 warnings:
 
 Reek is officially supported for the following CRuby versions:
 
-  - 2.2 (EOL)
+  - 2.1
+  - 2.2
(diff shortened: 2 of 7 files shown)


mvz:
Oh man I'm torn about this.

I was going to say if we're going for the solution where we have to change the README just before the release we can just put a pointer to the v4.8.1 README, so documentation an code actually match for the given repo state.

However, this PR also solves the problem that all our `-U` links link to the master branch. The really terrible thing is that even after we release 5.0 there will be people using 4.x and getting broken links.

troessner:
>> Putting back the PrimaDonnaMethod documentation, ideally with a notice that it will be renamed in 5.0.

Just did that.

>> Putting just a notice in the README saying 'This is the README for master, please look here for the currently released version'

Unless we make that <marquee> and <h1> i doubt people are reading that, especially if you quickly want to look something up (what most of our users are doing i suppose) you won't catch that.

docs/Rake-Task.md
```diff
@@ -33,7 +33,7 @@ An more sophisticated rake task that would make use of all available configurati
 ```Ruby
 Reek::Rake::Task.new do |t|
   t.name          = 'custom_rake' # Whatever name you want. Defaults to "reek".
-  t.config_file   = 'config/.reek.yml' # Defaults to nothing.
+  t.config_file   = 'config/config.reek' # Defaults to nothing.
```
mvz: This can stay as it is I think.

troessner: While it's true that we're not checking for the file extension explicitly when specifying a config file I still think it's very confusing that all of a sudden we have a configuration file with a yaml extension.

mvz:
> Unless we make that "marquee" and "h1" i doubt people are reading that, especially if you quickly want to look something up (what most of our users are doing i suppose) you won't catch that.

I like to think I always notice, but there's no way of knowing, is there? Anyway, let's just merge this so we can think of the Awesomest Best Solution for Pointing People to the Right Documentation (ABSPPRD) at leisure.

troessner:
And merged it is ;)

docs/Rake-Task.md
mvz: Oh wait I missed that this was also in the docs. Please ignore my remark for this line.

PR #1335, opened 2018-06-10
Fix list output

### LINKED ISSUE #1334: Output of `--list` is broken
On master, `reek --list` outputs literal `\n`. This also means there's something wrong with how our cucumber scenarios check output.

mvz:
Fixes #1334 (except the cukes don't catch it).


troessner:
Good catch!

PR #1340, opened 2018-06-16
Reduce own smell suppressions

mvz:
Two refactorings to make Reek more fragrant.


troessner:
A couple of wonderful changes :)

PR #1343, opened 2018-06-17
Add support for --stdin-filename

### LINKED ISSUE #1296: Add a --stdin-filename option
[ALE](https://github.com/w0rp/ale) wants to check for problems with Vim buffers while you type, and doing so requires checking text via stdin, or temporary files. This poses a problem when you want to write a configuration file for ignoring some rules for certain files, etc. Other tools like ESLint solve this problem by allowing you to specify the path to a file being checked when the text is being checked via stdin, so you can apply whatever is in configuration files and also check for problems while you type.

I propose adding an `--stdin-filename` option so you can write `reek --stdin-filename=foo/bar.rb < tempoary_file` or similar, so you can both check for problems while you type and apply settings from configuration files specific to the files.

paul:
When other tools, like automatic linters within editors, run a sample
through reek, they typically do so via stdin. However, if there are reek
config exceptions to apply to that file, then reek has no way of knowing
what the original filename was, triggering false positives.

This adds an `--stdin-filename` CLI option that takes the
original filename, and passes it through to the Examiner and SourceCode
file to use as the `origin` attribute, allowing the configuration
directives to apply.

Fixes #1296 

@troessner I'm probably missing a few additional places where this might need to get additional tests. I added a cuke feature to start with, and a spec in `application_spec.rb` that passed as soon as I added the option, but before I implemented the changes in `ReportCommand` and `Examiner`. Do you have any pointers on where to look that you'd want specs added?

### DIFF
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -33,12 +33,14 @@ module Reek
     #   The configuration for this Examiner.
     #
     # @public
+    # :reek:LongParameterList { max_params: 6 }
     def initialize(source,
+                   stdin_filename: nil,
                    filter_by_smells: [],
                    configuration: Configuration::AppConfiguration.default,
                    detector_repository_class: DetectorRepository,
                    error_handler: NullHandler.new)
-      @source              = Source::SourceCode.from(source)
+      @source              = Source::SourceCode.from(source, filename: stdin_filename)
       @smell_types         = detector_repository_class.eligible_smell_types(filter_by_smells)
       @detector_repository = detector_repository_class.new(smell_types: @smell_types,
                                                            configuration: configuration.directive_for(description))
--- a/lib/reek/examiner.rb
+++ b/lib/reek/examiner.rb
@@ -11,6 +11,7 @@ module Reek
   # Applies all available smell detectors to a source.
   #
   # @public
+  # :reek:TooManyInstanceVariables { max_instance_variables: 5 }
   class Examiner
     # Handles no errors
     class NullHandler
@@ -33,14 +34,16 @@ module Reek
     #   The configuration for this Examiner.
     #
     # @public
+    # :reek:ControlParameter
     # :reek:LongParameterList { max_params: 6 }
     def initialize(source,
-                   stdin_filename: nil,
+                   origin: nil,
                    filter_by_smells: [],
                    configuration: Configuration::AppConfiguration.default,
                    detector_repository_class: DetectorRepository,
                    error_handler: NullHandler.new)
-      @source              = Source::SourceCode.from(source, filename: stdin_filename)
+      @source              = Source::SourceCode.from(source)
+      @origin              = origin || @source.origin
       @smell_types         = detector_repository_class.eligible_smell_types(filter_by_smells)
       @detector_repository = detector_repository_class.new(smell_types: @smell_types,
                                                            configuration: configuration.directive_for(description))
@@ -50,17 +53,13 @@ module Reek
     # @return [String] origin of the source being analysed
     #
     # @public
-    def origin
-      @origin ||= source.origin
-    end
+    attr_reader :origin
 
     # @return [String] description of the source being analysed
     #
     # @public
     # @deprecated Use origin
-    def description
-      origin
-    end
+    alias description origin
 
     #
     # @return [Array<SmellWarning>] the smells found in the source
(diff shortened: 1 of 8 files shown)


troessner:
This looks excellent to me. However travis is failing due to rubocop :)
Other than that I have no additional comments. @mvz ?

paul:
@troessner Yeah, I haven't had time to fix it. It's rubocop complaining about the complexity since I added the `override || fallback` to the case statement. I tried a couple different easy things to make the warning go away, but it resulted in (imo) worse code. 

Probably the best thing to do is have sub-classes of `SourceCode` (`IOSourceCode`, `FileSourceCode`, etc...) but that seems like a bigger change that would cascade into further refactoring, and I was hoping to not have to be that invasive for this feature. 

The quick and dirty thing to make rubycop happy but the code probably worse would be to extract a couple helper methods, something like this:

```ruby
def self.from(source, filename: nil)
  new(code: code_from_source(source), origin: origin_from_source(source, override: filename))
end
```

...but, I personally think "private helper class methods" like `code_from_source` and `origin_from_source` to be a worse smell than the rubocop "Complexity" metric in this case. @troessner thoughts?

troessner:
@paul i agree - in this case you can add an exception to our [Rubocop configuration](https://github.com/troessner/reek/blob/master/.rubocop.yml)

lib/reek/examiner.rb
```diff
@@ -33,12 +33,14 @@ def handle(_exception)
     #   The configuration for this Examiner.
     #
     # @public
+    # :reek:LongParameterList { max_params: 6 }
     def initialize(source,
+                   stdin_filename: nil,
```
mvz: Since we're in the process of releasing Reek 5, we could also require `source` to be a `Source::SourceCode` and avoid passing in this extra parameter.

Notice that `stdin_filename` is passed to `Source::SourceCode.from` and then read back through `source.origin`, so an alternative would be to *not* pass it to `Source::SourceCode` but store it and use it in `Examiner#description`.

spec/reek/cli/application_spec.rb
```diff
@@ -44,6 +44,18 @@
                configuration: Reek::Configuration::AppConfiguration,
                options: Reek::CLI::Options)
       end
+
+      context 'when a stdin filename is provided' do
+        let(:app) { described_class.new ['--stdin-filename', 'foo.rb'] }
+
+        it 'assumes that filename' do
```
mvz: I don't see the specific filename being checked in the implementation of this spec?

lib/reek/source/source_code.rb
```diff
@@ -59,12 +59,13 @@ def initialize(code:, origin:, parser: self.class.default_parser)
...
-      def self.from(source)
+      # :reek:ControlParameter
+      def self.from(source, filename: nil)
         case source
         when File     then new(code: source.read,           origin: source.path)
-        when IO       then new(code: source.readlines.join, origin: IO_IDENTIFIER)
+        when IO       then new(code: source.readlines.join, origin: filename || IO_IDENTIFIER)
         when Pathname then new(code: source.read,           origin: source.to_s)
-        when String   then new(code: source,                origin: STRING_IDENTIFIER)
+        when String   then new(code: source,                origin: filename || STRING_IDENTIFIER)
```
mvz: Why is `filename` used with both `IO` and `String`? For STDIN use I think only `IO` would be needed, so if we're going to add this to others, we might as well add this to all of them.

However, see my comment on `Examiner` about not passing filename to `SourceCode` at all.

mvz:
@paul will you be able to take a look at my comments soon?

paul:
@mvz Yup, sorry, seems like my only free time these days is on weekends. I'll get to it today hopefully.

spec/reek/cli/application_spec.rb
paul: I should probably just remove this spec. I wanted somewhere in the regular specs (not just features) that would test the `--stdin-filename` option, and this seemed like the most logical place to do so. However, once I dug in, everything in here is so mocked out I wasn't sure how to actually test it, or even what I should be testing. I'm also surprised it passed, had it failed I probably would have just removed it.

paul:
I had to rebase master to get CodeClimate to not error. Do you want me to also squash these into a single commit? I left it separate for now for easier review.

troessner:
Looks good to me as well, merging. Thanks @paul !

stevenharman:
I'm trying to get Reek working with Ale (https://github.com/w0rp/ale), but the `--stdin-filename` doesn't seem to be respecting the `exclude_paths` in my config. Is that intentional with this change? If not, I'll open a new Issue for further discussion, but wanted to check first.

Thanks!

PR #1347, opened 2018-06-19
Refactor samples.

### LINKED ISSUE #1346: Clean up our /samples directory
Goals:

* Remove obsolete files
* Create a consistent directory structure
* Unify configuration files
* Inline sample and configuration files to increase cohesion of our tests. Right now quite often you have to jump between samples and test

troessner:
Fixes #1346


troessner:
Ready for review!  What I've done:

* Remove obsolete files
* Create a consistent directory structure
* Unify configuration files
* Inline sample and configuration files to increase cohesion of our tests. Right now quite often you have to jump between samples and test

features/configuration_loading.feature
```diff
@@ -30,7 +30,14 @@ Feature: Offer different ways how to load configuration
 
   Scenario: Configuration file in working directory
     Given the smelly file 'smelly.rb'
-    And a configuration file '.reek.yml'
+    And a file named "config.reek.yml" with:
```
mvz: Reek should not pick this up since it's not `.reek.yml`, but it does. https://github.com/troessner/reek/pull/1315#issuecomment-394142481

samples/paths.rb
```diff
@@ -1,4 +1,14 @@
...
-SMELLY_FILE = SAMPLES_PATH.join('smelly.rb')
+SAMPLES_DIR = Pathname.new(__dir__).relative_path_from(Pathname.pwd)
+
+# Configuration paths
+CONFIGURATION_DIR = SAMPLES_DIR.join('configuration')
+REGULAR_CONFIGURATION_DIR = CONFIGURATION_DIR.join('regular_configuration')
+EMPTY_SUB_DIRECTORY_DIR = REGULAR_CONFIGURATION_DIR.join('empty_sub_directory')
+REGULAR_CONFIGURATION_FILE = REGULAR_CONFIGURATION_DIR.join('regular.reek.yml')
+HOME_DIR = CONFIGURATION_DIR.join('home')
+HOME_CONFIGURATION_FILE = HOME_DIR.join('home.reek.yml')
```
mvz: Having more constants here makes the specs harder to follow I think.

features/configuration_files/masking_smells.feature
```diff
@@ -34,7 +34,17 @@ Feature: Masking smells using config files
       """
 
   Scenario: provide extra masking inline in comments
-    Given the smelly file 'smelly_with_inline_mask.rb'
+    Given a file named "smelly_with_inline_mask.rb" with:
```
mvz: Why do you inline this but not others?

features/configuration_files/masking_smells.feature
```diff
@@ -53,9 +63,28 @@ Feature: Masking smells using config files
...
   Scenario: Disable UtilityFunction for non-public methods
-    Given the smelly file 'smelly_with_modifiers.rb'
-    And a configuration file 'non_public_modifiers_mask.reek'
-    When I run reek -c non_public_modifiers_mask.reek smelly_with_modifiers.rb
+    Given a file named "smelly_with_modifiers.rb" with:
+      """
+      # Smelly class for testing purposes
+      #
+      # Not necessary for the feature per se but for
+      # removing distracting output. :reek:UnusedPrivateMethod
```
mvz: The UnusedPrivateMethod detector is disabled by default, so this comment can be removed.

troessner: Good catch! Removed.

features/configuration_loading.feature
mvz: I added an issue for this: #1363.

samples/paths.rb
troessner: What are you suggesting as alternative?  For me having to parse 

```Ruby
SAMPLES_DIR.join('configuration').join('....')
```

potentially dozen of times in a spec file is even harder ;)

troessner: Especially things like 

```Ruby
SAMPLES_DIR.join('.reek.yml')
```

are not expressive at all. What you really mean by that is the "configuration that you consider the right one if all things work out". But that line doesn't give you the slightest hint about that. We can argue about the naming, but at least REGULAR_CONFIGURATION_FILE seems a lot more helpful than that.

features/configuration_files/masking_smells.feature
troessner: The reason for that was that the configuration file for this scenario was *only* used for this scenario while the other scenarios in this feature were talking about smelly.rb:

```Cucumber
Given the smelly file 'smelly.rb'
```

which is *always* talking about the same file. I guess you could call this the reference file for smelly code. 
I could inline all of those calls but that would be a *lot* of duplication and if we ever want to change our understanding of this reference file a huge. I know it's a tradeoff, but for now I'd not inline "smelly.rb" but only the files which we use for one specific scenario only. Wdyt?

features/configuration_loading.feature
troessner: And as proof for your hypothesis for that it now fails after I rebased against master 😆

samples/paths.rb
mvz: > potentially dozen of times in a spec file is even harder ;)

How about moving these definitions to a `let` in the relevant spec file(s)?

> at least REGULAR_CONFIGURATION_FILE seems a lot more helpful than that.

I agree, but having it defined in some other file is less helpful. So having

```ruby
let(:regular_configuration_file) { SAMPLES_DIR.join('.reek.yml') }
```

at the top of the spec file would work for me.

features/configuration_files/masking_smells.feature
mvz: Ok, that makes perfect sense.

troessner:
Addressed all comments - merge?

PR #1350, opened 2018-06-20
Update docs for DuplicateMethodCall.

### LINKED ISSUE #1231: DuplicateMethodCall gives false positive for variables in lamdas

troessner:
We can then close #1231

### DIFF
--- a/docs/Duplicate-Method-Call.md
+++ b/docs/Duplicate-Method-Call.md
@@ -42,7 +42,55 @@ The approach you take will depend on balancing other factors in your code.
 
 ## Current support in Reek
 
-Reek's Duplicate Method Call detector checks for repeated identical method calls within any one method definition. This is intended to complement the checks performed by tools such as [Flay](http://ruby.sadi.st/Flay.html) and [Simian](http://www.redhillconsulting.com.au/products/simian/).
+Reek's Duplicate Method Call detector checks for repeated identical method calls within
+any one method definition. This is intended to complement the checks performed by tools
+such as [Flay](http://ruby.sadi.st/Flay.html) and [Simian](http://www.redhillconsulting.com.au/products/simian/).
+
+## Edge cases
+
+Be aware that there are some edge cases like this code:
+
+```Ruby
+class Foo
+  def bar(switch)
+    case switch
+    when :a
+      ->(arg) { arg.call_me(:maybe); do_something }
+    when :b
+      ->(arg) { arg.call_me(:maybe); do_something_else }
+    when :c
+      ->(arg) { arg.call_me(:maybe); do_something_different }
+    end
+  end
+end
+```
+
+Reek cannot reliably detect that each call's receiver is a different arg and will report:
+
+```
+[5, 7, 9]:DuplicateMethodCall: Foo#bar calls 'arg.call_me(:maybe)' 3 times
+```
+  
+If you're running into this problem you can disable this smell detector for this method either via
+configuration:
+
+```Yaml
+---
+DuplicateMethodCall:
+  exclude:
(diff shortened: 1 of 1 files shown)


docs/Duplicate-Method-Call.md
```diff
@@ -42,7 +42,56 @@ The approach you take will depend on balancing other factors in your code.
 
 ## Current support in Reek
 
-Reek's Duplicate Method Call detector checks for repeated identical method calls within any one method definition. This is intended to complement the checks performed by tools such as [Flay](http://ruby.sadi.st/Flay.html) and [Simian](http://www.redhillconsulting.com.au/products/simian/).
+Reek's Duplicate Method Call detector checks for repeated identical method calls within
+any one method definition. This is intended to complement the checks performed by tools
+such as [Flay](http://ruby.sadi.st/Flay.html) and [Simian](http://www.redhillconsulting.com.au/products/simian/).
+
+Unfortunately there are some edge cases Reek can't reliably detect due to Ruby's dynamic nature.
```
mvz: I would change 'Unfortunately' to 'Be aware', so it sounds less apologetic.

docs/Duplicate-Method-Call.md
```diff
@@ -42,7 +42,56 @@ The approach you take will depend on balancing other factors in your code.
...
+      ->(arg) { arg.call_me(:maybe); do_something_else }
+    when :c
+      ->(arg) { arg.call_me(:maybe); do_something_different }
+    end
+  end
+end
+```
+
+In regards to duplicate method calls this should be fine since this happens inside a lambda expression.
+However as mentioned above Reek can't reliably detect such cases and will report:
```
mvz: I wouldn't say that the code is fine. Maybe change it to something like: 'Reek cannot reliably detect that each call's receiver is a different `arg`, and will report:'

docs/Duplicate-Method-Call.md
```diff
@@ -42,7 +42,56 @@ The approach you take will depend on balancing other factors in your code.
 
 ## Current support in Reek
 
-Reek's Duplicate Method Call detector checks for repeated identical method calls within any one method definition. This is intended to complement the checks performed by tools such as [Flay](http://ruby.sadi.st/Flay.html) and [Simian](http://www.redhillconsulting.com.au/products/simian/).
+Reek's Duplicate Method Call detector checks for repeated identical method calls within
+any one method definition. This is intended to complement the checks performed by tools
+such as [Flay](http://ruby.sadi.st/Flay.html) and [Simian](http://www.redhillconsulting.com.au/products/simian/).
+
```
mvz: Let's add a heading here, like 'Edge cases'.

troessner:
Excellent points, all addressed now ;)

mvz:
I think we can resolve #1131 in a similar way.

PR #1345, opened 2018-06-19
Remove useless check.

troessner:
Was neither tested in any way nor does it seem to make sense.

PR #1351, opened 2018-06-21
Make defaults.reek only relevant as documentation

mvz:
This stops in-development configuration changes interfering with running external versions of Reek. Case in point: CodeClimate.


mvz:
Nope, that didn't help.

mvz:
I still think this is a good change, though.

mvz:
It gets weirder and weirder since I can run codeclimate's reek engine just fine locally.

mvz:
Hm. It seems #1345 broke this.The check is relevant if no configuration file can be found at all.

troessner:
Change looks good to me as well but travis disagrees as well ;)

troessner:
Cant merge on my phone cause the build is failing because of CC. This is getting annoying. Lets quickly fix our CC integration or kill it.

mvz:
Ok, I'll merge this from here and see if that helps.

mvz:
I think that fixed it: https://codeclimate.com/github/troessner/reek/builds

troessner:
Awesome!

PR #1353, opened 2018-06-21
Revert "Remove useless check."

troessner:
Reverts troessner/reek#1345


mvz:
Ideally, we should have some test for this ...

troessner:
>> Ideally, we should have some test for this ...

Fully agree. For now, I just wanted to make sure that we don't accidentally release Reek without this code. We have an issue already for writing test so we can tackle this right after :)

PR #1354, opened 2018-06-21
Update docs for FeatureEnvy.

### LINKED ISSUE #1131: FeatureEnvy too curious about what's inside lambda expressions

troessner:
Fixes #1131 in analogy to #1350

### DIFF
--- a/docs/Feature-Envy.md
+++ b/docs/Feature-Envy.md
@@ -40,6 +40,50 @@ belongs to the Item class, not the Warehouse.
 
 _Feature Envy_ reports any method that refers to self less often than it refers to (ie. send messages to) some other object.
 
+## Edge cases
+
+Be aware that there are some edge cases like this code:
+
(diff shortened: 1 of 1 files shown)


mvz:
:tada:

PR #1356, opened 2018-06-21
[WIP] Remove Syntax smell detector.

### LINKED ISSUE #1355: Getting rid of the "Syntax" smell detector
So I started to look into this and I  can't really recall why we added that in the first place. The whole thing seems bogus:

1.) Broken syntax is detected by the parser gem wayyyyy before it actually reaches any detector. At first I thought this was something special we did for stuff that is passed via STDIN but it all goes through the same code. Long story short, we could have just relied on the parser gem and then catching the exception.
2.) I started to look into [report_syntax_errors](https://github.com/troessner/reek/blob/master/lib/reek/examiner.rb#L101) - when is that actually called in our tests? Funny enough it is *only* called when you pass in an empty source file which is ironic, since this isn't an error for sure.

Long story short: Should be easy to remove it - wip.

troessner:
Fixes #1355


mvz:
@troessner I'm continuing this PR now.

mvz:
Rebased and ready for review!

lib/reek/examiner.rb
```diff
@@ -113,14 +116,10 @@ def syntax_tree
     end
 
     def examine_tree
-      return unless syntax_tree
+      return [] unless syntax_tree
```
troessner: Can we add a comnent here that explains that this happens on a source file with no code? I do not consider this obvious.

troessner:
One detail + travis and we are good to go!

lib/reek/examiner.rb
mvz: Agreed.

mvz: I refactored the code so this whole line is gone :tada:.

PR #1364, opened 2018-06-24
Fix nonsensical ConfigurationFileFinder specs

### LINKED ISSUE #1349: Non sensical / not working examples in configuration_file_finder_spec.rb
We have 2 non sensical / not working examples in configuration_file_finder_spec.rb:

```
'returns the file in home if traversing from the current dir fails' # nr 1
'does not traverse up from :home' # nr 2
```

mvz:
This fixes the specs to actually test what is in the description.

Fixes #1349.


troessner:
Nice!

PR #1366, opened 2018-06-25
Add simple spec to document nil file behavior

### LINKED ISSUE #1352: Test case of no configuration found at all
#1345 removed a seemingly useless check but it turns out it is there for the case when no configuration file can be found. This was fixed in #1353, but we should add a test for this case.

mvz:
Fixes #1352.


troessner:
Nice

PR #1367, opened 2018-06-25
Remove Report::Formatter

mvz:
The Report::Formatter module itself only contained some methods that are better placed elsewhere. This moves those methods, and removes the Formatter namespace, since its contained classes are also named SomethingFormatter.


mvz:
GitHub isn't picking up my pushed changes. :hourglass_flowing_sand:  :sleeping:

mvz:
All conflicts fixed...

Issue #1357, opened 2018-06-23
Output hidden in Heroku CI


mdi:
I run reek as part of a custom script on Heroku CI, which isn't a TTY, so the output is hidden during the test run. It would be nice to have a way to force output in those kind of situations.

It looks like this issue was introduced in https://github.com/troessner/reek/pull/1122https://github.com/troessner/reek/pull/1122

mvz:
Hi @mdi, you can add de `--progress` command line option to enable progress reporting on non-TTYs.

mdi:
The `--progress` option seems to have no effect.

mvz:
Ok, something else must be going on. Since `reek --progress | more` works fine for me, I'm going to need a little bit more information on how you're running reek.

mdi:
I have a script that runs reek followed by rubocop followed by my test suite. When I run this script locally in a terminal, reek displays output (my terminal is a TTY).

<img width="1428" alt="mdi_michaels-mbp____src_mdi_fitrackr-api" src="https://user-images.githubusercontent.com/524808/41818587-ec816c08-777f-11e8-8e3c-2e4b68348c11.png">

But when this script is run on Heroku CI, reek displays no output. Note that rubocop and the test suite are displaying output

![fitrackr_ _pipeline___heroku](https://user-images.githubusercontent.com/524808/41818508-6d9658be-777e-11e8-871f-5954e50a5bfb.png)

The script is:

```
#!/usr/bin/env bash

set -x

function test_rb {
  bundle exec rails test "$@"
}

function rubocop_rb {
  bundle exec rubocop "$@"
}

function reek_rb {
  bundle exec reek --progress "$@"
}

reek_rb "$@" && rubocop_rb "$@" && test_rb "$@"
```

I first contacted Heroku support about it, and after a couple of days of back & forth with them, we came to the realization that the issue is Heroku is a non-TTY environment and reek hides output when in a non-TTY environment.

<img width="1214" alt="heroku_help" src="https://user-images.githubusercontent.com/524808/41818533-fe464860-777e-11e8-9740-a8604c2ea578.png">

mvz:
Even in a non-TTY output, Reek should at least output `0 total warnings`, and even that is missing. Something else must be going on.

mvz:
@mdi To eliminate any potential confusion: Which version of Reek are you using?

mdi:
v4.8.1

mvz:
[...]

PR #1368, opened 2018-06-25
Improve silently

mvz:
This does two things:
* Make `Silencer.silently` correctly restore the original situation.
* Removes use of `Silencer.silently` where it seems no longer necessary.

### DIFF
--- a/lib/reek/ast/node.rb
+++ b/lib/reek/ast/node.rb
@@ -2,9 +2,7 @@
 
 require_relative '../cli/silencer'
 
-Reek::CLI::Silencer.silently do
-  require 'parser'
-end
+require 'parser'
 
 module Reek
   module AST
--- a/lib/reek/cli/silencer.rb
+++ b/lib/reek/cli/silencer.rb
@@ -8,17 +8,20 @@ module Reek
     module Silencer
       module_function
 
-      # :reek:TooManyStatements { max_statements: 7 }
+      # :reek:TooManyStatements { max_statements: 9 }
       def silently
         old_verbose = $VERBOSE
+        old_stderr = $stderr
+        old_stdout = $stdout
+
         $VERBOSE = false
         $stderr = StringIO.new
         $stdout = StringIO.new
         yield
       ensure
         $VERBOSE = old_verbose
-        $stderr = STDERR
-        $stdout = STDOUT
+        $stderr = old_stderr
+        $stdout = old_stdout
       end
     end
   end
--- a/spec/reek/cli/silencer_spec.rb
+++ b/spec/reek/cli/silencer_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe Reek::CLI::Silencer do
       end.to output("there!\n").to_stdout
     end
 
-    it 'blocks output on $stderr after the block' do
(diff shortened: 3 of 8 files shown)


mvz:
My future ideal would be to pass IO down from `bin/reek` and never call `Kernel.puts` or `Kernel.warn`.

lib/reek/ast/node.rb
```diff
@@ -2,9 +2,7 @@
 
 require_relative '../cli/silencer'
 
-Reek::CLI::Silencer.silently do
-  require 'parser'
-end
+require 'parser'
```
troessner: So that problem with the parser gem was fixed upstream? How did you notice?

lib/reek/cli/silencer.rb
```diff
@@ -8,17 +8,20 @@ module CLI
...
+        old_stdout = $stdout
+
         $VERBOSE = false
         $stderr = StringIO.new
         $stdout = StringIO.new
         yield
       ensure
         $VERBOSE = old_verbose
-        $stderr = STDERR
-        $stdout = STDOUT
```
troessner: Nice catch.

spec/reek/cli/silencer_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    it 'restores output on $stdout after the block' do
+      expect do
+        described_class.silently { puts 'Hi!' }
+        puts 'there!'
+      end.to output("there!\n").to_stdout
+    end
+
+    it 'blocks output on $stderr after the block' do
```
troessner: Shouldn't this also read "restores output ...." like above?

spec/reek/configuration/excluded_paths_spec.rb
```diff
@@ -12,9 +12,7 @@
       let(:paths) { [smelly_source_dir, file_as_path] }
 
       it 'raises an error if one of the given paths is a file' do
-        Reek::CLI::Silencer.silently do
-          expect { exclusions.add(paths) }.to raise_error(Reek::Errors::ConfigFileError)
-        end
```
troessner: Interesting. I can't even recall why we had that there.

spec/reek/tree_dresser_spec.rb
```diff
@@ -1,7 +1,4 @@
-require_lib 'reek/cli/silencer'
-Reek::CLI::Silencer.silently do
-  require 'parser/ruby22'
-end
+require 'parser/ruby22'
```
troessner: How about making this consistent to above and use ruby25 in regard to the good old boyscout rule? ;)

lib/reek/ast/node.rb
mvz: I removed the silencer and noticed no extra output.

mvz: This whole PR came from my hunch on what's going on in #1357.

spec/reek/configuration/excluded_paths_spec.rb
mvz: The method emitted a warning and called `exit` at some point.

mvz:
I processed your comments, @troessner.

PR #1378, opened 2018-06-28
Convert strings to regexes for directory directives.

### LINKED ISSUE #1377: Converting strings to regexes with our new syntax is not working for directory directives
This will work:

```Yaml
---
detectors:
  UnusedPrivateMethod:
    exclude:
      - "/exclude regexp/"
```

but this won't:

```Yaml
---
directories:
  "app/controllers":
    UnusedPrivateMethod:
      exclude:
        - "/exclude regexp/"
```
Pull request incoming.

troessner:
Fixes #1377


troessner:
Alright, I fixed the problem with an absolutely hideous solution that makes our test suite green. Tomorrow I'll make it nice and pretty ;)

lib/reek/configuration/rake_task_converter.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Reek
+  module Configuration
+    # Responsible for converting configuration values coming from the outside world
+    # to whatever we want to use internally.
+    module RakeTaskConverter
```
mvz: This module seems tested but not used ...

lib/reek/configuration/configuration_converter.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+require_relative './configuration_validator'
+
+module Reek
+  module Configuration
+    # Responsible for converting marked strings coming from the outside world
+    # into proper regexes.
+    class ConfigurationConverter
```
mvz: This module seems used but not tested ...

lib/reek/configuration/configuration_converter.rb
```diff
@@ -0,0 +1,110 @@
...
+      def strings_to_regexes_for_directories
+        return unless configuration[AppConfiguration::DIRECTORIES_KEY]
+
+        configuration[AppConfiguration::DIRECTORIES_KEY].tap do |directories|
+          directories.keys.each do |directory|
+            directories[directory].each do |detector, configuration|
+              convertible_attributes(configuration).each do |attribute|
+                directories[directory][detector][attribute] = directories[directory][detector][attribute].map do |item|
+                  to_regex item
+                end
```
mvz: Amazing nesting :laughing:.

lib/reek/configuration/rake_task_converter.rb
troessner: It is used at the rake task at the [very end](https://github.com/troessner/reek/pull/1378/files#diff-09c36c2af56d3a3b653a7c7a4e31bb71R12) of this PR ;)

lib/reek/configuration/configuration_converter.rb
troessner: I know! 😆 

I tried to cut it down as much as possible but since our configuration is a deeply nested hash our source code that deals with it is some kind of reflection of it. A way to improve would be to use a proper class for the whole configuration with properly nested sub components via composition. Basically the "AppConfiguration" we have right now on steroids. This is something I have been thinking about for while because we seem to suffer from [PrimiviteObsession](https://refactoring.guru/smells/primitive-obsession) but that would be a *huge* change.

Certainly something though I'd like to discuss in the future :)

lib/reek/configuration/configuration_converter.rb
troessner: It's tested via configuration_file_finder_spec because that is where the tests resided for the existing "string to regex" feature. I'm happy to refactor this if you think that this rather belongs into configuration_converter_spec but I'd appreciate it if we could do this via separate issue / PR to not make this PR more messy than it already is.

lib/reek/configuration/rake_task_converter.rb
mvz: Ah, I don't know how I missed that.

PR #1382, opened 2018-07-01
[WIP] Show configuration path.

### LINKED ISSUE #1342: Have Reek print out which configuration file it is using
I have quite often the problem when toying around with Reek in development that some config file is in the way but I am not sure which one due to our "traverse up" policy. Let's just output "using configuration file ....".

troessner:
Fixes #1342


README.md
```diff
@@ -355,8 +355,18 @@ All options that go beyond the [Basic Smell Options](docs/Basic-Smell-Options.md
...
+
+#### Debugging trouble with the configuration
+
+Given Reeks dynamic mechanism of finding a configuration file you might run into a situation where you are not
+100% sure what configuration file Reek is using. E.g. you have a project specific configuration file in your
+project root and also another Reek configuration in your HOME directory that you use for all your other projects
+and for whatever reasons Reek seems to be using another configuration file that you assumed it would.
+
+In this case you can pass the flag `--show-configuration-path` to Reek which will cause Reek to show ???
+TODO: Does this even make sense? Why not show it by default? It's just one line that's very helpful.
```
troessner: @mvz wdyt?

mvz: In any case where we show progress by default we're probably fine showing this, other cases maybe not (e.g., it would mess up JSON output).

Too bad we called that flag `--progress` :slightly_smiling_face:, otherwise we could have just stuck this under there.

mvz: Oh, actually, we could add it to the JSON, YAML and HTML output as well, in the correct format!

lib/reek/configuration/configuration_file_finder.rb
```diff
@@ -42,6 +42,8 @@ def find_and_load(path: nil)
         #
         # @quality :reek:ControlParameter
         def find(path: nil, current: Pathname.pwd, home: Pathname.new(Dir.home))
+          # TODO: Here we need to return the configuration path as well, that means changing find_by_dir and find_in_dir
+          # as well.
```
mvz: Isn't the configuration path exactly what it is returning?

troessner: Yeah, I'm not fully sure anymore why I wrote down this TODO, thanks for the pointer 😆

features/configuration_files/show_configuration_file.feature
```diff
@@ -0,0 +1,46 @@
+Feature: Show configuration file
+
+  Scenario: No configuration file present
+    Given the clean file 'clean.rb'
+    When there is no configuration file and I run reek --show-configuration-path clean.rb
```
troessner: @mvz i just cant get this one scenario to pass because it will always find the .reek.yml in our root directory. The corresponding "allow" in the step definition seems to have no effect. Any idea how to fix that? If we cant fix it i see the following options:

* Drop the scenario. Not great.
* Test that scenario via rspec. Also not great because this will be super confusing
* Extend application_spec or use sth like rspec_command and move the whole feature there. This could also be the first step of getting rid of cucumber in our test suite

mvz: Here are my thoughts:

* The reason this doesn't work is that Cucumber runs the actual Reek script, which won't know anything about any `allow` called in the steps.
* We should definitely test this, either in Cucumber or RSpec.
* I don't think `rspec_command` will buy us anything, since it will have the same problem (but see below).
* If we run the scenarios in some temp dir outside of regular `$HOME`, we can really simulate the case where no file is found. I *think* Aruba lets you do this, but I'm not sure. If `rspec_command` does this out of the box, that would be a reason to switch.
* In general I prefer to just run the 'happy path' in a cucumber scenario, and test the edge cases in RSpec. This would avoid the entire issue, since we can just test this the same way we test `ConfigurationFileFinder`, and we could have just one scenario showing the output in the 'normal' case of a project-specific file.

So, my recommendation is: Use just one scenario to exercise the feature, and test the rest in RSpec.

features/configuration_files/schema_validation.feature
```diff
@@ -5,7 +5,7 @@ Feature: Validate schema
 
   Scenario: Our generated default configuration
     Given our default configuration file
-    And the clean file 'clean.rb'
+    And the clean file "clean.rb"
```
troessner: Boyscout rule here and below, you should be able it whatever you like and double quotes are cucumber convention.

lib/reek/cli/options.rb
```diff
@@ -32,21 +32,23 @@ class Options
...
+        @location_format         = :numbers
+        @progress_format         = tty_output? ? :dots : :quiet
+        @show_links              = true
+        @smells_to_detect        = []
+        @colored                 = tty_output?
+        @success_exit_code       = Status::DEFAULT_SUCCESS_EXIT_CODE
+        @failure_exit_code       = Status::DEFAULT_FAILURE_EXIT_CODE
+        @generate_todo_list      = false
+        @force_exclusion         = false
+        @show_configuration_path = false
```
troessner: Changes above are indentation changes.

troessner:
Ready for review @mvz !

features/step_definitions/sample_file_steps.rb
```diff
@@ -4,8 +4,8 @@
   write_file(filename, SAMPLES_DIR.join('smelly_source').join(filename).read)
 end
 
-Given(/^the clean file 'clean.rb'$/) do
-  write_file('clean.rb', CLEAN_FILE.read)
+Given(/^the clean file "(.*)"$/) do |filename|
+  write_file(filename, CLEAN_FILE.read)
```
mvz: :+1:

spec/reek/cli/application_spec.rb
```diff
@@ -136,4 +136,44 @@
...
+            say argument
+          end
+        end
+      EOS
+    end
+
+    around do |example|
+      Dir.mktmpdir('/tmp') do |tmp|
+        Dir.chdir(tmp) do
+          File.write('clean.rb', clean_file)
```
mvz: I think this spec will also work without writing a file here.

troessner: Just passed '.' as sources and removed both clean.rb and the File.write.

Issue #1381, opened 2018-07-01
ControlParameter inordinately slow on certain files


retrography:
Reek is very slow on large files containing lots of comments. These are often files containing few lines of code, for instance API interfaces, with many lines of comments documenting the functions.

Examples are `bench/example.ruby` in [here](https://github.com/rubychan/coderay/raw/master/bench/example.ruby) and `gems/aws-sdk-ec2/lib/aws-sdk-ec2/client.rb` in [here](https://raw.githubusercontent.com/aws/aws-sdk-ruby/master/gems/aws-sdk-ec2/lib/aws-sdk-ec2/client.rb).

troessner:
How are you running Reek and in what version?
When I use the example file from above with the latest Reek 5.0.0.  the parser gem crashes on the file:

```
  ( $ ) reek example.rb
Inspecting 1 file(s):
/Users/timo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/parser-2.5.1.0/lib/parser/builders/default.rb:1625: warning: regular expression has redundant nested repeat operator '*'
        !!!
        Source 'example.rb' cannot be processed by Reek due to a syntax error in the source file.

        This is a problem that is outside of Reek's scope and should be fixed by you, the
        user, in order for Reek being able to continue.

        Exception message:

        #<Parser::SyntaxError: unexpected token tINTEGER>

        Original backtrace:

        /Users/timo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/parser-2.5.1.0/lib/parser/diagnostic/engine.rb:73:in `process'
	/Users/timo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/parser-2.5.1.0/lib/parser/base.rb:262:in `on_error'
	/Users/timo/.rbenv/versions/2.3.1/lib/ruby/2.3.0/racc/parser.rb:259:in `_racc_do_parse_c'
```

retrography:
You are right, it crashes. The slowness seems to have been a result of some incompatibility after the ruby upgrade. I reinstalled all the gems and now it works (meaning it crashes!).

retrography:
OK, I found the culprit. It is strange. A directory structure with relatively small files. Each file separately has a small processing time, but the directory in its entirety takes a very long time. Would you mind checking [this](https://www.dropbox.com/s/dv9w8d0o2wuwrpu/scanners.zip?dl=1)? Thanks.

[...]

PR #1389, opened 2018-07-09
Improve control parameter recursion performance

mvz:
Simple fix for #1381. Could use some (performance) tests. Based on the research by @troessner described in https://github.com/troessner/reek/issues/1381#issuecomment-403267108.


troessner:
Oh, this is even explicitly mentioned in our method docs for `Node.each_node`:

```
      # If a type is searched for *and* listed in ignoring, it will be yielded
      # but traversal will not continue to its children.
```

Merged!

PR #1390, opened 2018-07-10
Parser 2.5.1.1 seems broken. Disallow it.

mvz:
See https://github.com/whitequark/parser/issues/518 and https://github.com/rubocop-hq/rubocop/issues/6092.


troessner:
That means we should also release a new version, on it.

PR #1391, opened 2018-07-10
Prepare v5.0.2.

troessner:


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 5.0.2 (2018-07-10)
+
+* (mvz) Parser 2.5.1.1 seems broken. Disallow it.
```
mvz: #1389 should be added here.

troessner: That seems like an internal change that is opaque to our users. AFAIR we don't put internal changes into our release notes to not confuse our users - i'm a bit confused why you want to add this now?

troessner: Updated.

troessner:
And released it is ;)

PR #1396, opened 2018-07-10
Refactor ControlParameter.

troessner:


troessner:
Ready for review!

troessner:
Addressed all of your feedback!

lib/reek/ast/node.rb
```diff
@@ -102,6 +104,31 @@ def source
...
+      #   so in this case we would return this
+      #
+      #     s(:send, nil, :charlie,
+      #       s(:lvar, :bravo))
+      #
+      #   as condition.
+      #
+      def condition
+        return unless CONDITIONAL_NODE_TYPES.include?(type)
+        super
```
mvz: I think you can remove this whole body, since this method will be overridden by the code in the `IfNode`, `CaseNode`, and `LogicOperatorBase` sexp extensions.

troessner: Unfortunately we can't remove this method because otherwise dozens of our specs will crash with errors like:

```
     # ------------------
     # --- Caused by: ---
     # NoMethodError:
     #   undefined method `condition' for #<#<Class:0x00007f82459254b8>:0x00007f82459253a0>
     #   ./lib/reek/smell_detectors/control_parameter_helpers/call_in_condition_finder.rb:47:in `uses_param_in_call_in_condition?'
```

mvz: Not the method, just the body :slightly_smiling_face:.

lib/reek/ast/node.rb
```diff
@@ -102,6 +104,9 @@ def source
         loc.expression.source_buffer.name
       end
 
+      # Method will be overridden by the code in the IfNode, CaseNode, and LogicOperatorBase sexp extensions.
+      def condition; end
```
Drenmi: As I've been working on RuboCop node extensions, I've sometimes come across cases where having a `NullNode` abstraction would simplify the implementation of cops. I'm still not sure if this will be more help or hinderance though. Have you come across cases where this would be helpful? 🙂 

This is a pretty non-specific question, so I'm okay with an equally non-specific answer. 😬

lib/reek/ast/node.rb
```diff
@@ -9,6 +9,8 @@ module AST
     # methods to ease the transition from Sexp to AST::Node.
     #
     class Node < ::Parser::AST::Node
+      CONDITIONAL_NODE_TYPES = [:if, :case, :and, :or].freeze
```
mvz: This constant is no longer needed.

lib/reek/ast/node.rb
Drenmi: Thanks for the input @mvz! One of the downsides I've considered is it could cause code that should (rightfully) fail horribly to instead sort of pass. In RuboCop, we have *a lot* of:

```ruby
if node && node.foo_type?
```

so my first, primitive implementation would be just a node responding `false` to `#*_type?`.

lib/reek/ast/node.rb
troessner: Fixed.

lib/reek/ast/node.rb
mvz: Yeah, you would have to be very careful with how it interacts with existing code.

PR #1400, opened 2018-08-05
Absorb Ataru code into specs

### LINKED ISSUE #1260: What about Ataru?
So, Ataru seems dead upstream, and they're now behind on the `thor` dependency. What shall we do? Some options (there may be others):

1. Drop Ataru.
2. Find some replacement for Ataru and use that.
3. Fork Ataru.
4. Not care about `thor`.

mvz:
Fixes #1260.

This shows how simple Ataru really is. I prefer to have it as spec code in Reek, because:

* We can fix things
* We can run the specs and see what is actually tested (hint: not README.md)
* And no outdated thor!


spec/quality/documentation_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+RSpec.describe 'Documentation' do
+  files = Dir.glob(File.expand_path('../../**/*.md', __dir__))
+  files.each do |file|
+    describe "from #{file}" do
+      text = File.read(file)
+      doc = Kramdown::Document.new(text, input: 'GFM')
+      blocks = doc.root.children
+      blocks.each do |block|
+        next unless [:codeblock, :codespan].include?(block.type)
+        next unless block.attr['class'] == 'language-ruby'
```
mvz: This means that blocks annotated with 'Ruby' are skipped. Ideally, I want to test those too, but it breaks for the README.md because one of the examples loads up `pry`.

mvz:
:scream: `vendor/`

mvz:
Comment processed!

troessner:
Oh, very nice, merged!

PR #1401, opened 2018-08-06
Fix JRuby build

mvz:
Fixes the JRuby build and makes it required.

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -9,6 +9,8 @@ end
 
 SimpleCov.at_exit do
   SimpleCov.result.format!
-  SimpleCov.minimum_coverage 98.88
-  SimpleCov.minimum_coverage_by_file 81.4
+  unless RUBY_ENGINE == 'jruby'
+    SimpleCov.minimum_coverage 98.88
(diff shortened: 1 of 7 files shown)


mvz:
Still to do: SimpleCov.

mvz:
`Dir.mktmpdir(prefix)` seems broken on JRuby :scream:.

mvz:
Actually, seems like we're using undefined behavior by calling it as `Dir.mktmpdir('/tmp')`. The first argument is supposed to be a prefix of the created directory name, so `foo` would result in `/tmp/foosomething`. The second argument would be where you put the directory, so `Dir.mktmpdir(nil, '/foo')` would result in `/foo/dsomething`.

What we're doing creates `/tmp/tmpsomething` on MRI, and `/tmpsomething` on JRuby. Neither behavior is documented.

troessner:
Nice clean up. I'm happy to merge this as it is. Do you want to keep the current number of commits?

mvz:
> Do you want to keep the current number of commits?

Yes, they carry some nice comments on the changes.

PR #1410, opened 2018-09-08
[WIP] Introduce Law of Demeter detector.

troessner:


troessner:
@mvz forget about the code, i am still figuring things out, but what do you think about the specs? I want to focus on parameters only for the first version of the detector and then build on that.

troessner:
Seems hopeless (see #1412), closing this one.

Issue #1411, opened 2018-09-09
A strategy for making our --todo feature more useful


troessner:
As I mentioned in #1406 I have been working on improving our todo list functionality so that the TodoCommand recognises if there is a configuration file present (either the default one or whatever is passed in via -c option) and if there is, append to that instead of creating a new file.

Unfortunately I hit a snag recently:  I thought I could just add another `detectors` block to our configuration file like this:

```yaml
      ---
      detectors:
        UncommunicativeMethodName:
          exclude:
          - Smelly#x
      # Auto generated by Reeks --todo flag
      detectors:
        UncommunicativeVariableName:
          exclude:
          - Smelly#x
```

but this doesn't work because `ConfigurationFileFinder.load_from_file` just discards all non-unique keys.

*I see 3 options to fix this:*

1.) Find the current `detectors` block and append the todo list configuration there. This might become very messy very fast, since we'd have take into account all different configuration file layouts there could. Looking for markers like "last 4 characters indented line in the first detectors block" sounds already incredibly hacky and error-prone.

2.) Read in the existing configuration, append the new todo list configuration and then write it out again. This might kill the formatting of the original file and delete comments. Both not acceptable obviously.

3.) Add the feature to Reek that allows to include additional configuration file via `include` directive.

*What I'd like to suggest:*

1.) Update the todo list generation so that it will only update the .reek.yml based on no existing configuration (which corresponds to what we have already since the TodoList command does not take into account any pre-existing configuration). The task will just abort with a corresponding message if there already is a .reek.yml present. So super simple and just suited for the use case where you introduce Reek into a legacy project. This will also make our todo list feature extremely simple to understand

2.)  Add the feature to Reek that allows to include additional configuration file via `include` directive and reverse-merge whatever it finds in those files.

3.) Come back to this feature and basically continue to work on what I start in the PR #1408 which would make the todo list feature very powerful since you could not just use it for introducing it to legacy projects but also to continuously work on bigger projects where you might not have control over people fixing warnings and you need to get back to a clean slate to continue.

WDYT?

mvz:
That sounds like an excellent plan.

troessner:
Glad you like it, I'll work on this the next couple of weeks. [1] is trivial and [3] is easy given [2]. So [2] is gonna be the big one ;)

troessner:
I'll tackle [2] in the next couple of days / weeks ;)

PR #1413, opened 2018-09-10
Simplify todo list feature.

troessner:
Implements the first part of #1411


troessner:
Ready for review!

lib/reek/cli/command/todo_list_command.rb
```diff
@@ -12,17 +12,19 @@ module Command
       # file that can serve as a todo list.
       #
       class TodoListCommand < BaseCommand
-        FILE_NAME = '.todo.reek'
+        DEFAULT_CONFIGURATION_FILE_NAME = Configuration::ConfigurationFileFinder::DEFAULT_FILE_NAME
+        DETECTOR_KEY = Configuration::AppConfiguration::DETECTORS_KEY
```
mvz: This level of indirection seems undesirable to me. I have a feeling that if it seems necessary, it probably makes sense for these constants to live in some module where they're readily accessible for all modules.

lib/reek/cli/command/todo_list_command.rb
```diff
@@ -44,6 +46,14 @@ def groups
               todos.inject(&:merge)
             end
         end
+
+        def write_to_file
+          File.open(DEFAULT_CONFIGURATION_FILE_NAME, 'w') do |configuration_file|
+            configuration_file.write HEADER
+            configuration_file.write({ DETECTOR_KEY => groups }.to_yaml)
+          end
+          puts "\n'#{DEFAULT_CONFIGURATION_FILE_NAME}' generated! You can now use this as a starting point."
```
mvz: I'd move the puts out of `#write_to_file`, so all output is concentrated in `#execute`.

lib/reek/cli/command/todo_list_command.rb
```diff
@@ -12,17 +12,19 @@ module Command
...
+        DETECTOR_KEY = Configuration::AppConfiguration::DETECTORS_KEY
+        HEADER = "# Auto generated by Reeks --todo flag\n"
+        EXISTING_FILE_MESSAGE = "\nExisting '#{DEFAULT_CONFIGURATION_FILE_NAME}' detected - aborting.\n"
+        DO_NOTHING_MESSAGE = "\nNo smells found - nothing to do, exiting.\n"
 
         def execute
           if smells.empty?
-            puts "\n'.todo.reek' not generated because "\
-                    'there were no smells found!'
+            puts DO_NOTHING_MESSAGE
```
mvz: Let's rename this to `NO_SMELLS_FOUND_MESSAGE` to match `EXISTING_FILE_MESSAGE`, since both messages indicate a situation where nothing is done.

troessner:
@mvz I addressed all of your comments and I moved all other key names out of AppConfiguration into Reek::Constants. As you can see in the diff, all of them were used outside of AppConfiguration already  so this feels like a good idea even regardless of this PR.
Please re-review.

lib/reek/constants.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Reek
+  #
+  # This module holds some constants that are used across Reek.
```
mvz: If we move these up into `Reek` we don't have to namespace them in the rest of the code. WDYT?

spec/reek/cli/command/todo_list_command_spec.rb
```diff
@@ -2,67 +2,95 @@
...
-          { 'UncommunicativeMethodName' => { 'exclude' => ['Smelly#x'] },
-            'UncommunicativeVariableName' => { 'exclude' => ['Smelly#x'] } }
-        }.to_yaml
-        expect(File).to have_received(:write).with(described_class::FILE_NAME, expected_yaml)
-      end
-    end
+      context 'when exists already' do
+        it 'does not update the configuration' do
+          write_existing_configuration default_configuration_file_name
+          command = todo_command configuration: default_configuration
```
mvz: Does it matter for the test what is passed in `configuration`?

spec/reek/cli/command/todo_list_command_spec.rb
```diff
@@ -2,67 +2,95 @@
...
-      it 'does not write a todo file' do
-        Reek::CLI::Silencer.silently { command.execute }
-        expect(File).not_to have_received(:write)
-      end
+    def write_smelly_file
+      File.write(SMELLY_FILE.basename, smelly_file)
+    end
+
+    def write_existing_configuration(file_name)
+      File.write(file_name, existing_configuration)
```
mvz: I'm not convinced this needs to be a method, since it is only called once. How about inlining it?

spec/reek/cli/command/todo_list_command_spec.rb
```diff
@@ -2,67 +2,95 @@
...
+                          sources: sources,
+                          configuration: configuration
+    end
 
-      it 'does not write a todo file' do
-        Reek::CLI::Silencer.silently { command.execute }
-        expect(File).not_to have_received(:write)
-      end
+    def write_smelly_file
+      File.write(SMELLY_FILE.basename, smelly_file)
```
mvz: I'm not convinced this needs to be a method, since it is only called once. How about inlining it?

lib/reek/constants.rb
troessner: Yes, great idea, will change that.

spec/reek/cli/command/todo_list_command_spec.rb
troessner: Ack

spec/reek/cli/command/todo_list_command_spec.rb
troessner: Ack

spec/reek/cli/command/todo_list_command_spec.rb
troessner: Good catch, removed.

troessner:
Addressed all your comments @mvz ;)

troessner:
@mvz I'll be in lovely Israel for the next 2 weeks without a laptop starting tomorrow, would be great if you could re-review today so we can hopefully merge this one before ;)

lib/reek/constants.rb
```diff
@@ -0,0 +1,13 @@
...
+  #
+  module Constants
```
mvz: This module can go now, right?

mvz:
Just one more thing and then it's good to go.

troessner:
Fixed ;)

.simplecov
```diff
@@ -9,6 +9,6 @@ end
 
 SimpleCov.at_exit do
   SimpleCov.result.format!
-  SimpleCov.minimum_coverage 98.9
+  SimpleCov.minimum_coverage 98.88
```
troessner: I feel like this PR is well tested enough :)

PR #1414, opened 2018-09-10
Add pry-byebug aliases to spec_helper.

troessner:
Loading them via .pryrc file stopped working for me for no apparent reason. Other than I see no harm in having those in our spec_helper.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -9,10 +9,23 @@ require_relative '../lib/reek/configuration/app_configuration'
 
 require_relative '../samples/paths'
 
+# :nocov:
 begin
-  Reek::CLI::Silencer.without_warnings { require 'pry-byebug' }
+  Reek::CLI::Silencer.without_warnings do
+    require 'pry-byebug'
+
+    Pry.commands.alias_command 'c', 'continue'
+    Pry.commands.alias_command 's', 'step'
+    Pry.commands.alias_command 'n', 'next'
+    Pry.commands.alias_command 'f', 'finish'
+
+    Pry::Commands.command /^$/, "repeat last command" do
+      _pry_.run_command Pry.history.to_a.last
+    end
+  end
 rescue LoadError # rubocop:disable Lint/HandleExceptions
 end
+# :nocov:
 
 require 'factory_bot'
 FactoryBot.find_definitions


troessner:
Errrrm, travis fails with

>> File (/home/travis/build/troessner/reek/spec/spec_helper.rb) is only (77.78%) covered. This is below the expected minimum coverage per file of (81.40%).

but we configured Simplecov to only track files under lib:

```Ruby
SimpleCov.start do
  track_files 'lib/**/*.rb'
end
```

I'm a tad confused that it fails like that. Locally this doesn't happen for me. Could this somehow be a travis issue?

troessner:
Ah, seems like I misunderstood "track_files". Quoting from [here](https://www.rubydoc.info/gems/simplecov/SimpleCov%2FConfiguration:track_files) this means:

>> Coverage results will always include files matched by this glob, whether or not they were explicitly required. Without this, un-required files will not be present in the final report.

I thought adding a "root" to the configuration like this:

```Ruby
SimpleCov.start do
  root 'lib/'
  track_files 'lib/**/*.rb'
  add_filter 'lib/reek/version.rb' # version.rb is loaded too early to test
  add_filter 'lib/reek/cli/options.rb' # tested mostly via integration tests
end
```
would fix it but if I do it like that it seems the filters below are ignored since it now fails for me locally:

>> File (/Users/timo/code/reek/lib/reek/version.rb) is only (0.00%) covered. This is below the expected minimum coverage per file of (81.40%).

So it's probably not travis or whatever but it's rather our configuration was not perfect right from the beginning but we never noticed. Weird. I'll keep investigating.

troessner:
This:

>> would fix it but if I do it like that it seems the filters below are ignored since it now fails for me locally:

seems like a simplecov bug. I now surrounded the section in question with # :nocov: - lets see if this helps and then it might make sense to create a bug report at simplecov.

mvz:
> Loading them via .pryrc file stopped working for me for no apparent reason.

That seems to work fine for me here. Maybe it's worth investigating the reason?

mvz:
Turning off the silencer might aid in debugging here.

troessner:
I'll pause the investigation until #1416 is solved, my guess here is that those are related since it works for you here as well.

troessner:
>> Turning off the silencer might aid in debugging here.

I think this was not only needed for debugging, this *was* the problem. Lesson learned ;)

PR #1420, opened 2018-09-18
Improve our simplecov set up.

### LINKED ISSUE #1419: Simplecov reports lack of coverage in spec files?
In the scope of [this pull request](https://github.com/troessner/reek/pull/1382) I added this stub of a future spec:

```Ruby
require_lib 'reek/cli/application'

RSpec.describe 'show configuration path' do
  let(:clean_file) do
    <<-EOS
      # Well documented Clean class.
      class Clean
        def hello(argument)
          say argument
        end
      end
    EOS
  end

  around do |example|
    Dir.mktmpdir('/tmp') do |tmp|
      Dir.chdir(tmp) do
        File.write('clean.rb', clean_file)
        example.run
      end
    end
  end
end
```

Surprisingly for me, simplecov fails now with:

```
Coverage report generated for RSpec to /Users/timo/code/reek/coverage. 6724 / 6771 LOC (99.31%) covered.
File (/Users/timo/code/reek/spec/reek/cli/show_configuration_path_spec.rb) is only (50.00%) covered. This is below the expected minimum coverage per file of (81.40%)
```

already which I find very odd. Is this the idea behind this that we write tests for set up code in specs as well? If yes then I'd like to question this configuration and rather have simplecov not cover our specs since this seems like too much to me.

I added a [root](https://www.rubydoc.info/gems/simplecov/SimpleCov%2FConfiguration:root) to the configuration like this:

```Ruby
SimpleCov.start do
  root 'lib/'
  track_files 'lib/**/*.rb'
  add_filter 'lib/reek/version.rb' # version.rb is loaded too early to test
  add_filter 'lib/reek/cli/options.rb' # tested mostly via integration tests
end
```

to prevent simplecov from checking anything outside of lib/ but if I do it like that it seems the filters below are ignored since simplecov fails again:

```
Coverage report generated for RSpec to /Users/timo/code/reek/lib/coverage. 2922 / 2985 LOC (97.89%) covered.
Coverage (97.89%) is below the expected minimum coverage (98.90%).
```

Looking at the coverage report I can see the version.rb up there with 0% coverage, so it seems like adding the root fixed the spec problem but introduced the new problem that our defined filters for the version file are not working anymore.

Is this a problem you have come across already @mvz ? It seems like a simplecov bug to me at the moment but maybe I'm missing something.

troessner:
Fixes #1419

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -2,6 +2,9 @@ SimpleCov.start do
   track_files 'lib/**/*.rb'
   add_filter 'lib/reek/version.rb' # version.rb is loaded too early to test
   add_filter 'lib/reek/cli/options.rb' # tested mostly via integration tests
+  add_filter 'spec/'
+  add_filter 'samples/'
+  coverage_dir 'tmp/coverage'
 end
 
 SimpleCov.at_exit do


.simplecov
```diff
@@ -1,7 +1,7 @@
 SimpleCov.start do
-  track_files 'lib/**/*.rb'
```
mvz: See https://github.com/troessner/reek/issues/1419#issuecomment-422386621 on why this line is needed.

troessner: Re-added.

mvz:
I tried this out and it has the unintended effect of putting the coverage report in `lib/coverage`. How about just adding `add_filter /^spec\//` instead?

troessner:
>>  How about just adding add_filter /^spec\// instead?

Even easier, simplecov supports [coverage_dir](https://www.rubydoc.info/gems/simplecov/SimpleCov/Configuration#coverage_dir-instance_method) which I just added to this PR, please re-review.

mvz:
Hm, that's a nice workaround, but I think your use of `root` is off:

> The root for the project. This defaults to the current working directory.

Making this lib will also mess with [project_name](https://www.rubydoc.info/gems/simplecov/SimpleCov/Configuration#project_name-instance_method). The generated file now says 'Coverage for Lib' at the top :slightly_smiling_face:.

I'm amazed that SimpleCov doesn't have a setting for 'only look at code in this directory' :confused:.

troessner:
Updated, please re-review. Excluding "samples/" was necessary as well, otherwise simplecov would analyze samples/path.rb.

.simplecov
```diff
@@ -2,6 +2,9 @@ SimpleCov.start do
   track_files 'lib/**/*.rb'
   add_filter 'lib/reek/version.rb' # version.rb is loaded too early to test
   add_filter 'lib/reek/cli/options.rb' # tested mostly via integration tests
+  add_filter 'spec/'
+  add_filter 'samples/'
+  coverage_dir 'tmp/coverage'
```
mvz: Do you prefer this over just `coverage/`?

troessner: I never look at the report unless it breaks, and even then the console output is helpful enough to adress the problem. Having "coverage" at the root level makes it seems very important which it isnt, at least for me. If you feel differently i am happy to change it back.

mvz: Makes sense :+1:

PR #1425, opened 2018-10-08
Enable exclusion of single files.

### LINKED ISSUE #1393: Can't exclude a single file
There is no way to add a single file to ignore list right now. `exclude_paths` does not work with file.

troessner:
Fixes #1393

### DIFF
--- a/features/configuration_files/exclude_paths_directives.feature
+++ b/features/configuration_files/exclude_paths_directives.feature
@@ -3,25 +3,24 @@ Feature: Exclude paths directives
   As a user
   I want to be able to exclude specific paths from being checked
 
-  Scenario: Exclude some paths
-    Given a file named "bad_files_live_here/smelly.rb" with:
-      """
-      # A smelly example class
-      class Smelly
-        def alfa(bravo); end
-      end
-      """
+  Scenario: Exclude paths
+    Given the smelly file "smelly.rb" in the directory "smelly_sources"
+    And the smelly file "smelly.rb" in the directory "smelly_as_well"
+    And the smelly file "smelly.rb" in the directory "smelly_as_well_2"
     When I run reek .
     Then the exit status indicates smells
     Given a file named "config.reek" with:
       """
       ---
       exclude_paths:
-        - bad_files_live_here
+        - smelly_sources/smelly.rb
+        - smelly_as_well/
+        - smelly_as_well_2
       """
     When I run reek -c config.reek .
     Then it succeeds
     And it reports nothing
+
   Scenario: Using a file name within an excluded directory
     Given a file named "bad_files_live_here/smelly.rb" with:
       """
--- a/features/configuration_files/schema_validation.feature
+++ b/features/configuration_files/schema_validation.feature
@@ -35,8 +35,8 @@ Feature: Validate schema
 
     exclude_paths:
       - lib/legacy
-      - lib/rake/legacy_tasks
     """
+    And a directory named "lib/legacy"
     And the clean file "clean.rb"
     When I run reek -c config.reek clean.rb
     Then it succeeds
--- a/lib/reek/configuration/excluded_paths.rb
+++ b/lib/reek/configuration/excluded_paths.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative './configuration_validator'
+require_relative '../errors/config_file_error'
 
 module Reek
   module Configuration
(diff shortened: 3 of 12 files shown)


features/configuration_files/schema_validation.feature
```diff
@@ -35,8 +35,8 @@ Feature: Validate schema
 
     exclude_paths:
       - lib/legacy
-      - lib/rake/legacy_tasks
     """
+    And a directory named "lib/legacy"
```
troessner: Necessary, because now we are actually rasing on non-existent directories.

spec/reek/configuration/app_configuration_spec.rb
```diff
@@ -7,48 +7,65 @@
...
 RSpec.describe Reek::Configuration::AppConfiguration do
   describe 'factory methods' do
-    let(:expected_excluded_paths) do
-      [SAMPLES_DIR.join('two_smelly_files'),
-       SAMPLES_DIR.join('source_with_non_ruby_files')]
-    end
-
-    let(:expected_default_directive) do
-      { Reek::SmellDetectors::IrresponsibleModule => { 'enabled' => false } }
+    around do |example|
```
troessner: I think that in hindsight it was a huge mistake that we structured our specs so that they rely on external sample directories and external configuration directories since that makes it very hard to reason about them. Not to mention that a single spec is not self-reliant but has a whole dependency tree of files and directories it expects. I'm trying to reverse this development and I started with this file in the scope of this feature. As you can see, this spec is fully self-reliant and there are no external dependencies.

troessner:
Ready for review!

lib/reek/configuration/excluded_paths.rb
```diff
@@ -10,10 +11,13 @@ module Configuration
...
       include ConfigurationValidator
 
-      # @quality :reek:NestedIterators { max_allowed_nesting: 2 }
+      # @param paths [String]
+      # @return [undefined]
       def add(paths)
-        paths.each do |path|
-          with_valid_directory(path) { |directory| self << directory }
+        paths.map { |path| Pathname(path) }.each do |path|
+          raise Errors::ConfigFileError, "Exclude path #{path} does not exist" unless path.exist?
```
mvz: I'm not sure we want to do that since often people may add, e.g., editor artifacts to such exclusion lists, and they may or may not exist.

spec/reek/configuration/app_configuration_spec.rb
mvz: Yes, I think it's a good idea to make this change. Perhaps we should add a note to the `CONTRIBUTING.md` file?

features/configuration_files/exclude_paths_directives.feature
```diff
@@ -22,6 +22,60 @@ Feature: Exclude paths directives
...
+        def alfa(bravo); end
+      end
+      """
+    When I run reek .
+    Then the exit status indicates smells
+    Given a file named "config.reek" with:
+      """
+      ---
+      exclude_paths:
+        - bad_files_live_here/
```
mvz: Can we combine all these cases into a single scenario? I think it becomes a bit tedious otherwise.

lib/reek/configuration/excluded_paths.rb
troessner: I think it's advantageous for the user that we warn him with a bang instead of silently ignoring it.
Reek is all about code quality, why not enforce "configuration quality" as well?
I can't think of a good use case when a user would not want to be alerted of a useless configuration setting - do you have an example?

spec/reek/configuration/app_configuration_spec.rb
troessner: Good idea, added a paragraph.

features/configuration_files/exclude_paths_directives.feature
troessner: Done ;)

troessner:
Adressed all comments, ready for re-review.

lib/reek/configuration/excluded_paths.rb
mvz: I have, for example, a `tmp/` directory in my Reek checkout. It is git-ignored so it won't exist on a fresh checkout. It contains several sample Ruby files which I parked there to work on Reek issues. I can imagine adding `tmp/` as an ignored directory in `.reek.yml`, but wouldn't want Reek to fail on a fresh checked-out repository.

In general, it's mainly an issue when different people working on the same project have different local artifacts lying around in their repositories that they want to ignore but may not necessarily exist.

Another example is the `vendor/` directory. Notice that it is ignored in our `.rubocop.yml` config because it exists in CI. People might similarly run just `reek` in CI, but not want it to scan `vendor/`.

troessner: But the downside is that you might have some configuration in your config file that you think is doing sth when in fact is doing nothing because you made a typo in the path you listed. We could fix this by adding a "strict" mode or similar in the future.

I do see your point though and will remobmve those checks.

troessner:
Addressed all comments - merge? :)

PR #1426, opened 2018-10-10
[WIP] Allow users to include other configuration files.

troessner:
Fixes part 2 of #1411

### DIFF
--- a/features/configuration_files/include_other_configuration_files.feature
+++ b/features/configuration_files/include_other_configuration_files.feature
@@ -1,0 +1,55 @@
+Feature: Include other configuration files
+  In order to give users more flexibility when it comes to configuration files
+  we want to make it possible to include other configuration files into the main
+  configuration file.
+
+  Scenario: Include other configuration files
+    Given a file named "smelly.rb" with:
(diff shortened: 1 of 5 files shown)


troessner:
@mvz should we allow deep nesting of inclusion? wdyt?

features/configuration_files/include_other_configuration_files.feature
```diff
@@ -0,0 +1,55 @@
...
+      [3]:UnusedParameters: Smelly#x has unused parameter 'y'
+    """
+
+  Scenario: Included file has additional include directives
+
+  Scenario: Other configuration file overwrites configuration in the main configuration
+
+  Scenario: Other configuration file does not exist
+
+  Scenario: Other configuration file contains a syntax error
```
mvz: Maybe we can handle the error cases in specs instead?

mvz:
> should we allow deep nesting of inclusion? wdyt?

Yes, I think we should probably allow it, unless it makes things more complicated.

troessner:
Stale, closing for now.

PR #1431, opened 2018-10-18
Use squiggly heredocs instead of strip_heredoc.

troessner:
I'd also like to suggest to rename *all* heredoc delimiters we have to HERE so it's consistent everywhere (in a separate PR though). Right now it's EOF, EOS, SRC, FOO and god knows what else ;)


mvz:
* Yes to squiggly heredocs if supported by all supported Rubies
* No to having all delimiters be `HERE`. Let's use a format-descriptive delimiter instead as suggested by the `Naming/HeredocDelimiterNaming` cop (currently disabled by our config), and the discussion in https://github.com/troessner/reek/pull/1259#discussion_r139307715.

spec/reek/report/code_climate/code_climate_report_spec.rb
```diff
@@ -26,7 +26,7 @@
     let(:source) { 'def simple(a) a[3] end' }
 
     it 'prints smells as json' do
-      expected = <<-EOS.strip_heredoc.delete("\n")
+      expected = <<~HERE.delete("\n")
```
mvz: I really wonder why we escape all the double quotes below here?

mvz:
RuboCop may be able to autocorrect all the strip_heredocs if you change the config?

mvz:
I meant what @Drenmi said :slightly_smiling_face: .

troessner:
Makes sense, updated ;=

PR #1452, opened 2019-02-18
Enable regex for directory directives

pbernery:
The code was already ready for that but escaped the potential regexp declared in the reek configuration, actually disabling regexp.

This commit fixes this by removing the call to escape.

This allows this kind of declaration in the configuration file:

\# .reek.yml
```yaml
---
[...]
  "engines/.*/test/.*":
    InstanceVariableAssumption:
      enabled: false
    UncommunicativeVariableName:
      enabled: false
```

If this PR is accepted, I'll update the documentation to show this _feature_.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,12 +309,18 @@ detectors:
 # You can configure smells on a per-directory base.
 # E.g. the classic Rails case: controllers smell of NestedIterators (see /docs/Nested-Iterators.md) and
 # helpers smell of UtilityFunction (see docs/Utility-Function.md)
-# Note that we only allow configuration on a directory level, not a file level, so all paths have to point to directories.
+#
+# Note that we only allow configuration on a directory level, not a file level,
+# so all paths have to point to directories.
+# A Dir.glob pattern can be used.
 directories:
   "web_app/app/controllers":
     NestedIterators:
       enabled: false
-  "web_app/app/helpers":
+  "web_app/app/helpers**:
+    UtilityFunction:
+      enabled: false
+  "web_app/lib/**/test/**":
     UtilityFunction:
       enabled: false
 
--- a/lib/reek/configuration/directory_directives.rb
+++ b/lib/reek/configuration/directory_directives.rb
@@ -74,6 +74,7 @@ module Reek
                      gsub(%r{/\*\*$}, '<<to_eol_wildcards>>').
                      gsub('**', '<<to_wildcards>>').
                      gsub('*', '[^/]*').
+                     gsub('.', '\.').
                      gsub('<<to_eol_wildcards>>', '.*').
                      gsub('<<to_wildcards>>', '.*')
                  else
--- a/lib/reek/configuration/directory_directives.rb
+++ b/lib/reek/configuration/directory_directives.rb
@@ -53,7 +53,7 @@ module Reek
       # @quality :reek:FeatureEnvy
       def best_match_for(source_base_dir)
         keys.
-          select { |pathname| source_base_dir.to_s.match(/#{Regexp.escape(pathname.to_s)}/) }.
+          select { |pathname| source_base_dir.to_s.match(/#{pathname.to_s}/) }.
           max_by { |pathname| pathname.to_s.length }
       end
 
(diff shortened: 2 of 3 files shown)


pbernery:
Note that with this solution, if there is this directive "engines/.*/test/.*", it only trigger for subfolders of  "engines/.*/test/", not for files in "engines/.*/test/", meaning the directive may have to be duplicated.

I believe this part of the code should use a Dir.glob like syntax, which is commonly used in Ruby configuration files. The example of this PR would look like this in that case:

```yaml
---
[...]
  "engines/**/test/**":
[...]
```

What do you think?

troessner:
>> I believe this part of the code should use a Dir.glob like syntax, which is commonly used in Ruby configuration files. The example of this PR would look like this in that case:

I like the idea. @mvz wdyt?

mvz:
Yes, using glob-like syntax would definitely improve this :+1:.

We are currently matching by substring, but it seems that is an accident. It is also undocumented :man_shrugging:, so we can safely make this improvement.

troessner:
Sounds good.  @pbernery are you up for updating your PR accordingly?

pbernery:
Sure, I'll have a look into that.

pbernery:
I've spent some time trying to implement something with `Dir.glob`.

There will be an incompatibility with the existing configuration: currently, a directory directive like "foo/bar/baz" sets a configuration for this directory and all files under it. 
With `Dir.glob`, "foo/bar/alpha" will match exactly this directory but not its subdirectories. To include subdirectories, the configuration should be "foo/bar/alpha/**".

Do you think we can ask user to update their configuration if this feature is integrated?
I also thought about adding a new configuration key, which would be something like "file_directives": using `Dir.glob`, we can let user specify rules for file patterns instead of just directories. For instance, something like "test/**/*_test.rb".

troessner:
>> Do you think we can ask user to update their configuration if this feature is integrated?

That's not really an option but we can do a new major release for this given that the feature seems quite useful.

mvz:
We can simply assume trailing `**/*` for current values, I think.

Also, don't we just blanket stop descending into directories once they're found to be ignored?

pbernery:
> I think I'd rather not have a new key but instead have a major release, a short migration guide (can be a paragraph and an example, nothing fancy) and some updated docs.

I like this idea too.

> We can simply assume trailing `**/*` for current values, I think.

Yes, and no. Yes if you don't want to support exact match, but this is not an expected behaviour when having configuration based on `Dir.glob`, this could lead to some incomprehension from users I believe.

> Also, don't we just blanket stop descending into directories once they're found to be ignored?

I am not sure I can do that easily with an implementation based on `Dir.glob` as complex patterns may be given.

mvz:
I would prefer to implicitly add `**/*` at the end over having a breaking change at this time.

The most important part is fleshing out how this feature should behave with some specs, and how that differs from the current behavior. There is currently a bit of logic to find the 'best' match, for example, and we should think about how that will behave in any new system.

To sum up, I like the idea of allowing globbing, but we need to think a bit more about how that should work exactly.

pbernery:
Alright.

So I can start with a first implementation with implicit `**/*` and we can then discuss from that. This is actually what I did at first, I should have something rapidly.

I'll try to push something this week.

pbernery:
Sorry for the delay, here is a proposition of a configuration that is compatible with the current one, and that supports `Dir.glob` like patterns.

troessner:
This looks good to me :)
@mvz thoughts?
@pbernery let's add some documentation (to this PR) and then make this ready to merge unless @mvz has comments.

mvz:
I think this is definitely the way to go at this time :+1:.

A couple of things I think we should look at:

* Any `.` and perhaps some other characters will need to be escaped.
* `**` should indeed expand to `.*`, but `*` should probably become `[^/]*` or similar.

pbernery:
I updated the documentation and added a gsub statement to escape `.`.

@mvz, there is already a rule to replace `*` by `[^/]*` if I am not mistaken. Or did I miss something?

pbernery:
I noticed the CI didn't pass because of an alignment issue. I fixed that.

Do you think this PR is good to merge or is there something missing?

troessner:
I'd be good with merge-squashing it like it is. @mvz you good as well?

troessner:
Merged! Thank you @pbernery, I'll release it right away.

troessner:
And released it is as 5.4.0 ;)

PR #1464, opened 2019-03-21
Travis: Drop outdated setting "sudo: false"

olleolleolle:
This setting now does nothing, see https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration


troessner:
Nice catch, merging, thanks!

Issue #1469, opened 2019-03-29
Parsing Current Ruby and Syntax Error messages


mvz:
This is a follow-up on #1460.

I'd like to discuss the following points:
* Should Reek parse with Parser's 'current ruby'?
* Should we simplify the Syntax Error message so it doesn't always look like Reek crashed?

My opinion is a tentative yes to both questions:
* The first change will avoid Reek having to play catch-up with Parser's development: If a new version of Ruby is supported by Parser, Reek will automatically support it when using the latest version of Parser. Syntax errors on older Rubies make sense because the code should be valid syntax anyway and Reek should probably not be the first thing you run when you change a file.
* The second change is mostly for aesthetic reasons but will hopefully also make instructions for actually checking the syntax stand out more.

troessner:
That's an easy one: I'd give a strong yes in both cases :)

chastell:
Me too, yes on both counts! :+1: :+1:

etagwerker:
Yes to both. 👍

mvz:
This was implemented in #1474.

Issue #1460, opened 2019-03-21
Support running reek on open-ended range [1..]


olleolleolle:
Open-ended range fails.

## Reproduce

Have a file with an open range.

```ruby
p [:a, :b, :c, :d, :e][2..]
```

Use 2.6.2.

```
$ ruby -v
ruby 2.6.2p47 (2019-03-13 revision 67232) [x86_64-darwin18]
```

Try the program with ruby:

```
$ ruby o.rb
[:c, :d, :e]
```

Run `reek` to inspect this file, and see failure on `unexpected token tRBRACK`:

```
$ reek o.rb
    Inspecting 1 file(s):
        !!!
        Source 'o.rb' cannot be processed by Reek due to a syntax error in the source file.

        This is a problem that is outside of Reek's scope and should be fixed by you, the
        user, in order for Reek being able to continue.

        Things you can try:
        - Check the syntax of the problematic file
        - If the file is not in fact a Ruby file, exclude it in your .reek.yml file

        Exception message:

        #<Parser::SyntaxError: unexpected token tRBRACK>

        Original backtrace:

        /Users/olle/.rvm/gems/ruby-2.6.2/gems/parser-2.6.2.0/lib/parser/diagnostic/engine.rb:73:in `process'
	/Users/olle/.rvm/gems/ruby-2.6.2/gems/parser-2.6.2.0/lib/parser/base.rb:262:in `on_error'
	/Users/olle/.rvm/rubies/ruby-2.6.2/lib/ruby/2.6.0/racc/parser.rb:259:in `_racc_do_parse_c'
	/Users/olle/.rvm/rubies/ruby-2.6.2/lib/ruby/2.6.0/racc/parser.rb:259:in `do_parse'
	/Users/olle/.rvm/gems/ruby-2.6.2/gems/parser-2.6.2.0/lib/parser/base.rb:167:in `parse'
	/Users/olle/.rvm/gems/ruby-2.6.2/gems/parser-2.6.2.0/lib/parser/base.rb:184:in `parse_with_comments'
	/Users/olle/.rvm/gems/ruby-2.6.2/gems/reek-5.3.1/lib/reek/source/source_code.rb:116:in `parse'
	/Users/olle/.rvm/gems/ruby-2.6.2/gems/reek-5.3.1/lib/reek/source/source_code.rb:52:in `syntax_tree'
[...]

PR #1474, opened 2019-04-28
Use current ruby for parsing

mvz:
Resolves part of #1469.

* Use `Parser::CurrentRuby` for parsing
* Silence parser's verbose warnings

### DIFF
--- a/lib/reek/source/source_code.rb
+++ b/lib/reek/source/source_code.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 require_relative '../cli/silencer'
-Reek::CLI::Silencer.without_warnings { require 'parser/ruby26' }
+# Silence Parser's warnings about Ruby micro version differences
+Reek::CLI::Silencer.silently { require 'parser/current' }
 require_relative '../tree_dresser'
 require_relative '../ast/node'
 require_relative '../ast/builder'
@@ -53,7 +54,7 @@ module Reek
       end
 
       def self.default_parser
-        Parser::Ruby26.new(AST::Builder.new).tap do |parser|
+        Parser::CurrentRuby.new(AST::Builder.new).tap do |parser|
           diagnostics = parser.diagnostics
           diagnostics.all_errors_are_fatal = true
           diagnostics.ignore_warnings      = true


troessner:
Wait, isnt this a huge change? This means that when I am running Reek under 2.3 on a 2.6 source the 2.3 parser will be used, doesnt it?

mvz:
> Wait, isnt this a huge change? This means that when I am running Reek under 2.3 on a 2.6 source the 2.3 parser will be used, doesnt it?

I'm not sure it's that big a change in practice, but yes, that's what would happen. But why would you run Reek under 2.3 on 2.6 source? Also, we currently have the opposite problem: If I run Reek under 2.3 on a 2.3 source, it will use the 2.6 parser.

troessner:
This pull request confuses me quite a bit. What problem are we solving?

> But why would you run Reek under 2.3 on 2.6 source?

I actually do this all the time. I have a lot of ruby projects, most of them running with different ruby versions. Being able to quickly run Reek across multiple projects without having to switch to a different Ruby version manually is quite valuable to me. I also imagine a lot of false bug reports coming our way from people being very confused about this.

> Also, we currently have the opposite problem: If I run Reek under 2.3 on a 2.3 source, it will use the 2.6 parser.

How is that a problem? The Ruby 2.6 parser should handle 2.3 just fine, shouldn't it?

mvz:
> What problem are we solving?

This comes from #1469, which was inspired by #1460 and the discussion there. Basically, it would avoid Reek breaking on newer Ruby syntax whenever a new Ruby version comes out.

Maybe we need to revisit the discussion in #1469 since it seems it was unclear what I was proposing.

> Being able to quickly run Reek across multiple projects without having to switch to a different Ruby version manually is quite valuable to me.

That makes sense. Do you do that while your current version is less than 2.6?

troessner:
> That makes sense. Do you do that while your current version is less than 2.6?

Yes, mostly its 2.3 right now.

etagwerker:
I think this PR would be useful for `metric_fu`, I'm trying to solve an issue related to it over here: https://github.com/metricfu/metric_fu/pull/308/files#issuecomment-531776242

What is the worst that can happen? 

> Wait, isnt this a huge change? This means that when I am running Reek under 2.3 on a 2.6 source the 2.3 parser will be used, doesn't it?

You would get a bunch of syntax errors if the codebase is using Ruby 2.6 features, right? 

I believe the best thing about this PR is that it would solve issues like this one: https://github.com/metricfu/metric_fu/issues/305

troessner:
Hey folks,

sorry for letting this pull request just be for so long. I'm open to change the default behaviour, merge this one and release a new minor version. @mvz you're still up for it?

mvz:
@troessner yes I'm still up for it. Should I rebase to trigger a new build?

troessner:
Yes, that wouod be nice @mvz

mvz:
Done!

troessner:
Merged! New release? I think we should also mention this somewhere prominently in our docs.

mvz:
I'll prepare a new release.

mvz:
See #1497. Please take a look at the (small) change to the README.

PR #1475, opened 2019-05-01
Update RuboCop dependency

mvz:
Updates RuboCop to the 0.68.x series. With RuboCop 0.68.0, rubocop-performance was extracted to a separate gem, so this is added as a separate dependency and loaded in the RuboCop configuration.


troessner:
I don't think we need a proper review for PRs like that, feel free to merge yourself once CI is green ;)

PR #1489, opened 2019-08-25
Simplify SmellWarning class

mvz:
* Pass smell type instead of full class to `SmellWarning#initialize`
* Simplify spec setup by skipping detector creation

### DIFF
--- a/lib/reek/smell_warning.rb
+++ b/lib/reek/smell_warning.rb
@@ -15,22 +15,21 @@ module Reek
     extend Forwardable
 
     # @public
-    attr_reader :context, :lines, :message, :parameters, :smell_detector, :source
-    def_delegators :smell_detector, :smell_type
+    attr_reader :context, :lines, :message, :parameters, :smell_type, :source
 
     # @note When using Reek's public API, you should not create SmellWarning
     #   objects yourself. This is why the initializer is not part of the
     #   public API.
     #
     # @quality :reek:LongParameterList { max_params: 6 }
-    def initialize(smell_detector, context: '', lines:, message:,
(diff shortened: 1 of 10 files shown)


lib/reek/smell_warning.rb
```diff
@@ -15,22 +15,21 @@ class SmellWarning
...
-    def_delegators :smell_detector, :smell_type
+    attr_reader :context, :lines, :message, :parameters, :smell_type, :source
 
     # @note When using Reek's public API, you should not create SmellWarning
     #   objects yourself. This is why the initializer is not part of the
     #   public API.
     #
     # @quality :reek:LongParameterList { max_params: 6 }
-    def initialize(smell_detector, context: '', lines:, message:,
+    def initialize(smell_type, context: '', lines:, message:,
```
troessner: Let's use this opportunity to add proper parameter documentation :)

mvz: Done!

mvz:
Hm. Can we not squash merge? I really hate it.

troessner:
You mean not squash anymore at all or are you referring to this specific case? If it's the latter, how about leaving a comment for the reviewer (or label it) that we shouldnt squash on merge?

mvz:
I mean in general not use GitHub's squash merge option.

mvz:
A short list of why I don't like it:
* It loses fine-grained git blame information
* It makes it harder to see that local branches have been merged
* Its method of generating a commit message is very simplistic

troessner:
Ok, no worries, I'm ok with not using it anymore :)

PR #1501, opened 2019-12-29
Support Ruby 2.7

mvz:
* Add Ruby 2.7 to the build matrix
* Clean up passing of hash for kwargs
* Avoid passing keyword arguments to raise


mvz:
There are some warnings left over from FactoryBot's use of `delegate`. These will be fixed in some future version of ActiveSupport.

PR #1510, opened 2020-02-01
Refactor data clump detector

mvz:
Because each detector now detects one context we don't need the extra classes anymore.


troessner:
Very nice!

PR #1513, opened 2020-02-01
Raise error when legacy code comment format is used

### LINKED ISSUE #1341: Detect old-style reek comments
As noticed in #1322, Reek currently still accepts old-style comments with three colons, because the regex used will simply ignore the third colon and everything that follows, essentially blanked-disabling the given smell. This means users will not normally be aware that they need to update their code comments.

Options to fix:

1. Detect this particular usage and complain about it loudly. We probably should have done this already when we deprecated the extra colon.
2. Make the regex less lenient and just ignore the entire code comment if it doesn't use the new style. The reek failures should point out what needs updating, but this could be confusing.

Option 2 could be implemented by:
* Requiring the comment to come at the end of a line.
* Requiring the comment to be followed by anything but a colon.

mvz:
Fixes #1341


troessner:
Failed on CI with our own quality spec ;)

PR #1516, opened 2020-02-17
docs: solnic.eu -> solnic.codes

CGA1123:
solnic.eu no longer points to the intended source.

old link is nsfw fyi.


troessner:
Thanks!

Issue #1369, opened 2018-06-25
Remove deprecated ModuleContext#node_instance_methods


mvz:
This method has been deprecated since Reek 3.8. It's time.

PR #1518, opened 2020-02-28
Remove deprecated node_instance_methods method

mvz:
#1369.


mvz:
Remove code, coverage drops !?

mvz:
Ah, there were some leftover methods in this commit. I have pushed an updated commit. The coverage is still down a tiny bit since the total number of lines is lower but the set of uncovered lines remains the same. Funny how that works.

mvz:
There are some differences between the deprecated method and its replacement, so this has to be handled a bit more carefully.

PR #1532, opened 2020-04-14
Fix example of reek configuration in docs

rasaffie:
Example of configuration in README was broken in [this commit](https://github.com/troessner/reek/commit/421990de37c74e6afee2e431c9eff5ffbb618e4e#diff-04c6e90faac2675aa89e2176d2eec7d8L339), erasing directory configuration.


troessner:
Merged, thanks!

PR #1534, opened 2020-04-18
Remove FactoryBot and ActiveSupport dependencies

mvz:
This replaces our FactoryBot factories with simple methods. This allows us to drop dependency on FactoryBot, and, finally, after one minor change, on ActiveSupport.

The reason to do this now is that FactoryBot does not work on Ruby 2.8.0-dev, due to the existing warnings becoming errors.

The reason to do this at all rather than wait for FactoryBot to catch up, is that I've long disliked the development dependency on ActiveSupport, which causes methods to be available on core classes that may not be available to Reek at runtime. This can cause Reek to break without our tests catching it. By removing ActiveSupport as a dependency, we take away one source of such problems.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,10 +5,10 @@ gemspec
 ruby RUBY_VERSION
 
 group :development do
+  gem 'activesupport',         '~> 6.0'
   gem 'aruba',                 '~> 1.0'
   gem 'codeclimate-engine-rb', '~> 0.4.0'
   gem 'cucumber',              '~> 3.0'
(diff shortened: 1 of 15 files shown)


troessner:
Very nice!

troessner:
But CI and 2.4 is red ;)

mvz:
Well that's one bug found I guess.

Drenmi:
It's beautiful.

❤️🧡💛💚💙💜

Issue #1540, opened 2020-05-22
Relative paths in .reek.yml directories section do not match to absolute paths


maximpertsov:
When I run reek via overcommit, it does not honor custom configurations in the `directories` section of my `.reek.yml` file. After debugging the issue a bit, I found that the problems stems from the fact that overcommit passes in absolute directory paths, which do not resolve to the relative paths in `directories`.

For example, here is a part of my `.reek.yml` configuration:
```yaml
directories:
  "app/controllers/api/v1":
    UncommunicativeModuleName:
      enabled: false
```

Reek works correctly when I run `bundle exec reek`. However, when I run the command via overcommit, it reports linting errors that should have been disabled through my `directories` configuration.

However, reek via overcommit works correctly if I change my `directories` configuration to use absolute directory paths like so:
```yaml
directories:
  "<PWD>/app/controllers/api/v1":
    UncommunicativeModuleName:
      enabled: false
```

maximpertsov:
Here is my proposal to address the issue: https://github.com/troessner/reek/pull/1541, unless the current behavior is intended.

PR #1541, opened 2020-05-22
Match absolute paths of directories in .reek.yml

maximpertsov:
Proposed solution to address https://github.com/troessner/reek/issues/1540


mvz:
Thanks, @maximpertsov, that looks like a good solution.

@troessner I think you know this part of the code best. Any thoughts?

troessner:
I like it! Good to merge from my side, let's go ;)

troessner:
Thanks @maximpertsov !

troessner:
And released it is! 

> Pushed reek 6.0.1 to rubygems.org

PR #1544, opened 2020-06-01
Set link metadata in gemspec

### LINKED ISSUE #1536: Rubygems.org "source code" link gives 404
https://rubygems.org/gems/reek

Click on "Source Code"

mvz:
It is no longer possible to set these via the rubygems.org interface. Instead, these should now be set in the gemspec. Wiki and mailing list are omitted since the former is disabled and the latter hasn't seen
activity since 2012.

See https://blog.rubygems.org/2019/03/08/and-then-there-was-one-metadata-links.html for more info on the metadata UI removal.

Fixes #1536.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -57,6 +57,7 @@ Lint/BooleanSymbol:
 # Spec blocks can be any size
 Metrics/BlockLength:
   Exclude:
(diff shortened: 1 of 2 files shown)


troessner:
Traaaaaaavisssssssss :)

PR #1554, opened 2020-09-30
Update .travis.yml

gogainda:


mvz:
This PR doesn't make much sense :confused:

mvz:
@gogainda after looking at your other recent pull requests, I see what you're trying to do here. But open source maintainers shoudn't have to go hunting for this. Since you're filing a lot of these, consider making a standard text to put in these pull requests to explain the purpose.

gogainda:
Usually I close such pr when I see that build is failing. In that case tests are passing but code cov is different for truffleruby so it failed

mvz:
You're still alerting all the maintainers with your test. You can also run the tests on your own machine.

PR #1567, opened 2020-12-04
Do not run RuboCop in the GitHub Action test job

mvz:
RuboCop is run in a separate job so don't duplicate this effort.


mvz:
An alternative solution is to adjust the `ci` task itself. @troessner thoughts?

troessner:
> An alternative solution is to adjust the ci task itself.

I think the solution above is nice already, merging :)

PR #1574, opened 2021-01-03
In a refinement, assign smells to the refined module or class

### LINKED ISSUE #1563: Fix UtilityFunction false positive with Kernel refinement
## Overview

Seeing the following warning:

``` bash
lib/versionaire/cast.rb -- 1 warning:
  [7]:UtilityFunction: Versionaire::Cast#Version doesn't depend on instance state (maybe move it to another class?) [https://github.com/troessner/reek/blob/v6.0.2/docs/Utility-Function.md]
```

...for this refinement:

``` ruby
module Versionaire
  # Refines Kernel in order to provide a top-level Version conversion function.
  module Cast
    refine Kernel do
      def Version object
        Versionaire::Version object
      end
    end
  end
end
```

Potential Usage:

``` ruby
using Versionaire::Cast

Version "1.2.3"  # => #<struct Versionaire::Version major=1, minor=2, patch=3>
```

## Desired Behavior

It would be nice to not flag this as a violation in this case. This is similar in nature to what was [reported earlier](https://github.com/troessner/reek/issues/1438).

## Notes

I do a lot with refinements so, for context, this might help:

- [Refinements](https://www.alchemists.io/projects/refinements) - A collection of enhancements to Ruby core classes.
- [Versionaire](https://www.alchemists.io/projects/versionaire) - The Ruby `Version` type, shown above, that I'd like to access at the Kernel level much like `Integer`, `String`, `Array`, `Hash`, etc.

mvz:
Fixes #1563.


lib/reek/context_builder.rb
```diff
@@ -263,9 +264,16 @@ def process_super(exp, _parent)
     #
     # Counts non-empty blocks as one statement.
     #
+    # A refinement block is handled differently and causes a RefinementContext
+    # to be opened.
+    #
     def process_block(exp, _parent)
       increase_statement_count_by(exp.block)
-      process(exp)
+      if exp.call.name == :refine
```
troessner, line 272: I kind of dislike how this method now went from very simple to not so simple anymore but I don't have a better idea so ... 🤷

troessner (approved):

PR #1584, opened 2021-02-04
Fix duplicate 6.0.3 header (should be 6.0.2) in CHANGELOG.md

smlsml:
Simple typo, there are two 6.0.3 version headers in the file, clearly the earlier one was supposed to be 6.0.2


mvz:
Nice catch, thanks!

smlsml:
Thanks for the quick merge! The first of many, maybe :)

Issue #1538, opened 2020-05-15
False positive of UtilityFunction for class methods


VladYermakov:
Hi there

I've observed a false positive for UtilityFunction when class methods are defined under `class_methods` block of `ActiveSupport::Concern`

For instance

```ruby
module SomeModule
  extend ActiveSupport::Concern

  class_methods do
    def some_method(arg1, arg2)
      return 1 if arg1.something?
      return 2 if arg2.something_else?

      3
    end
  end
end
```

So here I'm getting `UtilityFunction` warning, but it makes no sense, and by definition it wouldn't warn this method was defined using `self.`

Thanks in advance

mateusluizfb:
Reproduced in version `'reek', '~> 6.0'` and the warning still occurs.

Can I open a PR to fix the issue?

mvz:
@mateusluizfb yes, please. Let me know if you need any pointers.

mateusluizfb:
@mvz PR here https://github.com/troessner/reek/pull/1596 :)

PR #1596, opened 2021-03-29
Do not raise Utilityfunction for methods defined in a class_method block

mateusluizfb:
Addresses https://github.com/troessner/reek/issues/1538

This PR implements a way to not raise `UtilityFunction` warning for methods defined in `class_methods` block

### DIFF
--- a/lib/reek/context/method_context.rb
+++ b/lib/reek/context/method_context.rb
@@ -85,6 +85,17 @@ module Reek
         ''
       end
 
+      def class_method_block?
+        return false unless @parent_exp
+
+        result = @parent_exp.each_node(:send).select do |node|
+          node_args = *node
+          node_args.include?(:class_methods)
+        end
+
+        !result.empty?
+      end
+
       private
 
       attr_reader :parent_exp
--- a/spec/reek/smell_detectors/utility_function_spec.rb
+++ b/spec/reek/smell_detectors/utility_function_spec.rb
@@ -177,6 +177,27 @@ RSpec.describe Reek::SmellDetectors::UtilityFunction do
     end
   end
 
+  context 'when examining class_methods block' do
+    it 'reports on the refined class' do
+      src = <<-RUBY
+        module Alpha
+          extend ActiveSupport::Concern
+
+          class_methods do
+            def bravo(arg1, arg2)
+              return 1 if arg1.something?
+              return 2 if arg2.something_else?
+
+              3
+            end
+          end
+        end
+      RUBY
+
+      expect(src).not_to reek_of(:UtilityFunction, context: 'Alpha#bravo')
+    end
+  end
+
   describe 'method visibility' do
     it 'reports private methods' do
       src = <<-RUBY
--- a/lib/reek/smell_detectors/utility_function.rb
+++ b/lib/reek/smell_detectors/utility_function.rb
@@ -57,8 +57,7 @@ module Reek
       # @return [Array<SmellWarning>]
       #
       def sniff
-        return [] if context.singleton_method? || context.module_function?
-        return [] if context.references_self?
+        return [] if context.singleton_method? || static_method_context?
         return [] if num_helper_methods.zero?
         return [] if ignore_method?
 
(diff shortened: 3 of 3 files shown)


lib/reek/context/method_context.rb
```diff
@@ -85,6 +85,17 @@ def full_comment
         ''
       end
 
+      def class_method_block?
+        return false unless @parent_exp
+
+        result = @parent_exp.each_node(:send).select do |node|
+          node_args = *node
+          node_args.include?(:class_methods)
+        end
```
mateusluizfb, line 94: Is there a way to `pattern match` the `s(:send, nil, class_methods)` directly?

troessner:
I am kind of hesitant to merge something that is specific to Rails but since it's seemingly used a lot...idk. Wdyt @mvz ?

mvz:
Yes, I'm also wondering how we can support this in a more generic way. Also, I think there's in issue that the proposed solution will suppress UtilityFunction in the whole module instead of just inside the `class_method` block. I need to look into this a little bit more over the weekend.

spec/reek/smell_detectors/utility_function_spec.rb
```diff
@@ -177,6 +177,27 @@ def bravo(charlie)
     end
   end
 
+  context 'when examining class_methods block' do
+    it 'reports on the refined class' do
```
mvz, line 181: This description seems wrong.

lib/reek/context/method_context.rb
mvz: You can check `node.name == :class_methods`

spec/reek/smell_detectors/utility_function_spec.rb
```diff
@@ -177,6 +177,27 @@ def bravo(charlie)
...
+
+              3
+            end
+          end
+        end
+      RUBY
+
+      expect(src).not_to reek_of(:UtilityFunction, context: 'Alpha#bravo')
+    end
+  end
```
mvz, line 199: Can you also add a test case that shows that methods defined outside the `class_methods` block still trigger a UtilityFunction warning?

mvz:
To generalize this, we could add a configuration option to list which methods introduce a block that defines class methods. The default option would be empty, but users can set it to `[:class_methods]` if Reek is used in a Rails project.

troessner:
Excellent idea @mvz!

spec/reek/smell_detectors/utility_function_spec.rb
mateusluizfb: Opps, ctrl + C problems

spec/reek/smell_detectors/utility_function_spec.rb
mateusluizfb: Yup

mateusluizfb:
@mvz nice idea, is there something in the project that I could reuse or base myself? Like a custom Rails configuration module or a class that does something similar

chopperg (commented):
spec/reek/smell_detectors/utility_function_spec.rb

troessner:
@mvz seems like we forgot this one, I can see that you assigned this to yourself, any chance you pick it up again?

mateusluizfb:
Feel free to ping me again as well

spec/reek/smell_detectors/utility_function_spec.rb
```diff
@@ -177,6 +177,53 @@ def bravo(charlie)
...
+          def charlie(arg1, arg2)
+            return 4 if arg1.something?
+            return 5 if arg2.something_else?
+
+            6
+          end
+        end
+      RUBY
+
+      expect(src).not_to reek_of(:UtilityFunction, context: 'Alpha#charlie')
```
claudiob, line 223: @mateusluizfb Thanks for this PR!

I have a question about this line. `charlie` is defined _outside_ of the `class_methods do` block.
So I would expect it to raise a `UtilityFunction` warning. 

Can you help me understand why this expectation is a `.not_to reek_of`? Thanks.

Issue #12, opened 2009-04-23
Large Module


kevinrutherford:
The name Large Class is a hangover from Martin Fowler's original Java smells; in Ruby it should probably be renamed to Large Module.

PR #1599, opened 2021-04-27
Add build-base to Dockerfile

frnico33:
There was an issue when was trying to build the docker image. This fix adds the developer tools to the image.

`#12 3.786 You have to install development tools first.`

Complete error:

[Docker build error.txt](https://github.com/troessner/reek/files/6384623/Docker.build.error.txt)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ ENV code_dir /code
 ENV app_dir /usr/src/app
 ENV user app
 
-RUN apk --update add git
+RUN apk --update add git build-base
 ADD . ${app_dir}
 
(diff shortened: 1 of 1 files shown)


troessner:
Oh, interesting, wasn't aware of this. @mvz wdyt?

mvz (approved):
I can reproduce the issue, and also confirm that this change fixes it.

It seems this problem occurs because we now need to compile native extensions for psych, which we didn't have to when the Dockerfile was created

frnico33:
Thanks for the quick review! @troessner @mvz

PR #1602, opened 2021-04-30
Add GitHub action to automerge minor and patch updates

mvz:


troessner:
Very nice!

mvz:
Hm .. this may not actually work.

PR #1619, opened 2021-09-25
Drop support for Ruby 2.5

mvz:
With the release of JRuby 9.3.0.0, we can drop compatibility with Ruby 2.5.

- Require Ruby 2.6 or higher in the gemspec
- Update RuboCop configuration and fix new offenses
- Update set of Rubies in CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ demo.rb -- 2 warnings:
 
 ## Supported Ruby versions
 
-Reek is officially supported for CRuby 2.4 to 2.7 and for JRuby 9.2.
+Reek is officially supported for CRuby 2.6 through 3.1 and for JRuby 9.3.
 Other Ruby implementations (like Rubinius) are not officially supported but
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -94,7 +94,7 @@ demo.rb -- 2 warnings:
 
 ## Supported Ruby versions
 
-Reek is officially supported for CRuby 2.4 to 2.7 and for JRuby 9.2.
+Reek is officially supported for CRuby 2.6 to 2.7 and for JRuby 9.3.
```
troessner, line 97: Shouldnt we mention 3.0 as well?

chastell (approved):
The current workflow also test JRuby 9.2 – but I’m for dropping both CRuby 2.5 and JRuby 9.2 if they’re intertwined. 👍

README.md
mvz: Fixed.

PR #1657, opened 2022-06-15
TooManyMethods only checks instance methods

whiteleaf7:
https://github.com/troessner/reek/blob/master/lib/reek/smell_detectors/too_many_methods.rb#L39
> TODO: Only checks instance methods!

I resolved the todo above.

### DIFF
--- a/docs/Too-Many-Methods.md
+++ b/docs/Too-Many-Methods.md
@@ -30,8 +30,8 @@ test.rb -- 1 warning:
 ```
 ## Current Support in Reek
 
-Reek counts all the methods it can find in a class &mdash; instance *and* class
-methods. So given `max_methods` from above is 4, this:
+Reek counts only instance methods it can find in a class.
+So given `max_methods` from above is 4, this:
 
 ```Ruby
 class Smelly
@@ -45,7 +45,7 @@ class Smelly
 end
 ```
 
-would cause Reek to emit the same warning as in the example above.
+would not cause Reek to emit the warning in the example above.
 
 ## Configuration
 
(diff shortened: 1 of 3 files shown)


troessner:
Thanks for the pull request! One thing though, could you please add a spec that tests the new behaviour explicitly rather than implicitly?

whiteleaf7:
@troessner 
[0bd6fca](https://github.com/troessner/reek/pull/1657/commits/0bd6fca95a63eaf0427c14bbff96074a7f09b43c)
I have modified the test, could you please check the test?

docs/Too-Many-Methods.md
```diff
@@ -45,7 +45,7 @@ class Smelly
 end
 ```
 
-would cause Reek to emit the same warning as in the example above.
+would not cause Reek to emit the warning in the example above.
```
troessner, line 48: This now doesnt make sense anymore and is very confusing to new users. You're basically saying what Reek now is NOT doing when documentation should tell you what it IS doing ;)
It would be great if you could invert the whole paragraph here.

spec/reek/smell_detectors/too_many_methods_spec.rb
```diff
@@ -51,4 +51,21 @@ def echo; end
...
+    src = <<-RUBY
+      class Alfa
+        def bravo; end
+        def charlie; end
+        def delta; end
+        def self.echo; end
+
+        class << self
+          def foxtrot; end
+        end
```
troessner, line 65: Hmmm, why use 2 different notations for class methods here?

troessner (changes requested):

troessner:
@whiteleaf7 any chance you'll pick this up again?

troessner:
Closing this one since it has become stale, feel free to re-open @whiteleaf7

PR #1669, opened 2022-09-12
Check all Ruby blocks in documentation specs

mvz:
We were missing most Ruby blocks in this check. This change will show that a lot of the Ruby blocks in the documentation cannot be executed.

We need to decide whether to skip this check altogether (we don't seem to use the `# =>` syntax a lot), or to fix the code documentation so this works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ UtilityFunction:
 
 For a basic overview, run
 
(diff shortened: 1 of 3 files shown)


mvz:
These specs have now been removed.

PR #1684, opened 2022-12-24
Allow failing safely when trying to recursively find a configuration …

gonubana:
…file

instead of triggering 'Permission denied @ dir_initialize - (Errno::EACCES)' by attempting to list an ancestor directory for which we don't have sufficient permissions.

```ruby
# Will never raise 'Errno::EACCES'.
Pathname.pwd.ascend { |d| p d }

# Can raise 'Errno::EACCES' as nothing guarantees that we have enough
# permissions to list every ancestor directory of our working directort.
Pathname.pwd.ascend { |d| p d.children }
```


troessner:
Looks good to me - @mvz wdyt?

mvz:
Yes, the code looks good and I've managed to reproduce the problem locally as well.

troessner:
Just released 6.1.3 with the fix, thanks @gonubana !

PR #1703, opened 2023-02-14
Add missing dependency rexml

tricknotes:
rexml has been removed from default gems since Ruby 3.0. https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

To use rexml, it should be added as runtime dependency. https://github.com/troessner/reek/blob/v6.1.4/lib/reek/report/xml_report.rb#L13

Since rexml is depended from rubocop and kramdown that are only development dependencies, all tests are passed on CI.


mvz:
Thanks @tricknotes! Can you please add a version limitation for this? I think "~> 3.1" should be fine.

tricknotes:
@mvz 
Thanks. It is a good idea.
I updated this PR with version limitation.

PR #1710, opened 2023-04-07
Fix badges.

troessner:


mvz:
This pull request is obsolete since the badges have been updated in #1711.

PR #1735, opened 2023-10-08
Rename Code Climate related classes

mvz:
The Code Climate classes were living under the Report namespace but did not all implement report classes. With this change, they are moved to the CodeClimate namespace instead. They are also moved to the `lib/reek/code_climate` directory to put their file names in line with their namespace name.

The spec files are moved as well, fixing an outstanding RSpec/SpecFilePathFormat offense.


troessner:
Nice, I just wanted to look in the Rspec related rubocop failure that popped up in our other PRs but I see you fixed this here as well, merging!

PR #1736, opened 2023-10-09
Fix cucumber_opts warning

JuanVqz:
I was reading the [Developing Reek Contributing](https://github.com/JuanVqz/reek?tab=readme-ov-file#developing-reek--contributing)
to understand how `reek` works, then run the `rake console` and saw the `warning` message.

```ruby
➜  reek git:(master) be rake console
WARNING: consider using an array rather than a space-delimited string with cucumber_opts to avoid undesired behavior.
```

after making it an array the `warning` was gone, let's see if that works for you.

### DIFF
--- a/tasks/test.rake
+++ b/tasks/test.rake
@@ -18,7 +18,7 @@ namespace 'test' do
   end
 
   Cucumber::Rake::Task.new(:features) do |t|
-    t.cucumber_opts = 'features --format progress --color'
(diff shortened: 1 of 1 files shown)


JuanVqz:
@troessner rebased!

troessner (approved):

mvz:
Thanks, @JuanVqz!

PR #1737, opened 2023-10-10
Fix Instance Variable Assumption false positive

### LINKED ISSUE #1492: False positive of Instance Variable Assumption with Struct initialize overload
Hi, I found an issue about [Instance Variable Assumption][instance-variable-assumption] that occurs when a class have a `Struct` with its own `initialize` method in its block.

[instance-variable-assumption]: https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md

Example code:

```ruby
class Foo
  Bar = Struct.new(:a) do
    def initialize(a)
      super
      p 'bar created'
    end
  end

  def initialize
    @foo = :foo
  end

  def foo?
    @foo == :foo
  end
end
```

When I run `reek` against this file, Reek reports a warning of `Instance Variable Assumption`, even though `@foo` in `Foo` class is initialized correctly.

```
[1]:InstanceVariableAssumption: Foo assumes too much for instance variable '@foo' [https://github.com/troessner/reek/blob/v5.4.0/docs/Instance-Variable-Assumption.md]
```

It looks like the cause is `Bar` which has a block containing an `initialize` method.
I confirmed that Reek did not warn if:

- `Bar` does not have an `initialize` method.
- `Bar` is defined by normal class syntax (`class Bar ... end`).
- `Bar` is defined after the `initialize` method of `Foo`.

Similar issue: #1137

JuanVqz:
Can you please have a look?

I hope this fixes #1492

What this basically does is look up through all the initializers if it matches the `ivasgn` to consider it instance variable assumption.

### DIFF
--- a/lib/reek/smell_detectors/instance_variable_assumption.rb
+++ b/lib/reek/smell_detectors/instance_variable_assumption.rb
@@ -20,7 +20,7 @@ module Reek
       # @return [Array<SmellWarning>]
       #
       def sniff
-        assumptions = (variables_from_context - variables_from_initialize).uniq
+        assumptions = (variables_from_context - variables_from_initializers).uniq
 
         assumptions.map do |assumption|
           build_smell_warning(assumption)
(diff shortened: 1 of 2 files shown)


troessner:
Hey @JuanVqz ,

that does look nice! @mvz any comments?

troessner (approved):

mvz (approved):
Looks good to me too. Let's merge this!

JuanVqz:
@troessner and @mvz thank you for having a look, I rebased it :+1:

mvz:
Thanks @JuanVqz!

PR #1740, opened 2023-10-11
Fix repeated conditional when local variables in diff methods

### LINKED ISSUE #1455: Repeated Conditional incorrectly reports in case there are local variable checks in different methods
In case we have code like

```
class RepeatedConditionals
  def repeat_1
    switch = true # or some other code, e.g. `Service.call(...)`
    puts "Repeat 1!" if switch
  end

  def repeat_2
    switch = false # or some other code, e.g. `Service.call(...)`
    puts "Repeat 2!" if switch
  end

  def repeat_3
    switch = 123 # or some other code, e.g. `Service.call(...)`
    puts "Repeat 3!" if switch
  end
end
```

reek incorrectly reports Repeated Conditional

It shouldn't do this if all of the conditionals are local variables.

JuanVqz:
There might be a lot of things broken I'm opening the PR just to have feedback, I hope that is fine.

Fixes #1455

### DIFF
--- a/lib/reek/smell_detectors/repeated_conditional.rb
+++ b/lib/reek/smell_detectors/repeated_conditional.rb
@@ -73,14 +73,27 @@ module Reek
       #
       # @quality :reek:TooManyStatements { max_statements: 9 }
       def conditional_counts
-        result = Hash.new { |hash, key| hash[key] = [] }
-        collector = proc do |node|
-          next unless (condition = node.condition)
-          next if condition == BLOCK_GIVEN_CONDITION
+        result = {}
+        collector = proc do |method|
+          [:if, :case].map do |stmt|
+            method.each_node(stmt) do |node|
+              next unless (condition = node.condition)
+              next if condition == BLOCK_GIVEN_CONDITION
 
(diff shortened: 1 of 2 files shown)


lib/reek/smell_detectors/repeated_conditional.rb
```diff
@@ -74,13 +74,18 @@ def max_identical_ifs
...
-          next unless (condition = node.condition)
-          next if condition == BLOCK_GIVEN_CONDITION
+        collector = proc do |method|
+          [:if, :case].map do |stmt|
+            method.each_node(stmt) do |node|
+              next unless (condition = node.condition)
+              next if condition == BLOCK_GIVEN_CONDITION
 
-          result[condition].push(condition.line)
+              result[method.name].push(condition.line)
```
JuanVqz, line 83: This is kind of incomplete. I was thinking of doing something like

```ruby
matches = {}
matches[method.name] = result[condition].push(condition.line)

{
  {:bravo => [4]},
  {:delta => [9]},
  {:echo => [12]},
}
```

Then count on each method if there are more that 2 or the allowed default

Issue #1734, opened 2023-10-06
Possible bug or improvement for error undefined method `keys'


fbuys:
Current behaviour
When a user:
- Defines a directory with a detector in the .reek.yml
- But fails to provide a configuration for the detector
It:
- Fails
- And produces a cryptic message that reads:
```
configuration_converter.rb:66:in `convertible_attributes': undefined method `keys' for nil:NilClass (NoMethodError)
```

Expected behaviour
When a user:
- Defines a directory with a detector in the .reek.yml
- But fails to provide a configuration for the detector
It:
- Does not fail
- Warns the user that they might be missing configuration for a specific directory + detector.
- Continues and produces reek result for the correctly defined detectors.

See: `lib/reek/configuration/configuration_converter.rb` (method convertible_attributes)
See: `lib/reek/configuration/configuration_converter.rb` (method strings_to_regexes_for_detectors)
See: `lib/reek/configuration/configuration_converter.rb` (method strings_to_regexes_for_directories)

Repo: 
- Clone: https://github.com/fbuys/reek_issue_1734
- bundle install
- run `bundle exec reek .`
- see error
- see `.reek.yml` config

mvz:
This has been fixed in #1749.

PR #1741, opened 2023-10-12
ISSUE-1734 Add missing configuration warning

fbuys:
This PR adds a more descriptive error message when users add a detector without configuration. Before this change a NoMethodError exception was raised with an unclear error message. Now we catch this exception and provide a more helpful error message.

A better future option might be add [dry-schema](https://dry-rb.org/gems/dry-schema/1.13/). 
The [current yaml schema validator](https://github.com/sunaku/kwalify) is outdated and does not seem to have good documentation.

See: https://github.com/troessner/reek/issues/1734

### DIFF
--- a/lib/reek/configuration/configuration_converter.rb
+++ b/lib/reek/configuration/configuration_converter.rb
@@ -80,6 +80,8 @@ module Reek
                 to_regex item
               end
             end
+          rescue NoMethodError
+            warn_about_missing_configuration(detector) if detectors[detector].nil?
           end
         end
       end
@@ -101,9 +103,24 @@ module Reek
                   to_regex item
                 end
               end
+            rescue NoMethodError
+              warn_about_missing_configuration(detector) if configuration.nil?
             end
           end
         end
+      end
+
+      def warn_about_missing_configuration(detector)
+        msg = <<~ERROR
+          ###############################
+
+          Please review the configuration file (e.g. .reek.yml in your project directory).
+          It looks like the configuration for #{detector} is missing.
+          You can find the configuration options documentation over here: https://github.com/troessner/reek#configuration-options.
+
+          ###############################
+        ERROR
+        warn msg
       end
     end
   end
--- a/lib/reek/configuration/default_directive.rb
+++ b/lib/reek/configuration/default_directive.rb
@@ -21,8 +21,10 @@ module Reek
       # @return [self]
       def add(detectors_configuration)
         detectors_configuration.each do |name, configuration|
-          detector = key_to_smell_detector(name)
-          self[detector] = (self[detector] || {}).merge configuration
+          if configuration
+            detector = key_to_smell_detector(name)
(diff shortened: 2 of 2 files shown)


lib/reek/configuration/configuration_converter.rb
```diff
@@ -80,6 +80,8 @@ def strings_to_regexes_for_detectors
                 to_regex item
               end
             end
+          rescue NoMethodError
+            warn_about_missing_configuration(detector) if detectors[detector].nil?
```
mvz, line 84: I think it's better to detect that `detector[detector]` is `nil` at the beginning of this block and handle that specific case there.

lib/reek/configuration/configuration_converter.rb
```diff
@@ -101,10 +103,25 @@ def strings_to_regexes_for_directories
                   to_regex item
                 end
               end
+            rescue NoMethodError
+              warn_about_missing_configuration(detector) if configuration.nil?
```
mvz, line 107: I think it's better to detect that `configuration` is `nil` at the beginning of this block and handle that specific case there.

lib/reek/configuration/configuration_converter.rb
```diff
@@ -101,10 +103,25 @@ def strings_to_regexes_for_directories
...
             end
           end
         end
       end
+
+      def warn_about_missing_configuration(detector)
```
mvz, line 113: The way error handling works in Reek is that we raise specific errors throughout the code and then do the warning in `lib/reek/cli/application.rb`.

We do this, among other reasons, because it allows us to test the error handling code without having warnings appear all over the place during the spec run.

In this case, the error should probably be a `Reek::Errors::ConfigFileError`. It will be handled here: https://github.com/troessner/reek/blob/b1fa495b4f0fa617552a71790e2fed56db667f77/lib/reek/cli/application.rb#L48

lib/reek/configuration/default_directive.rb
```diff
@@ -21,8 +21,10 @@ module DefaultDirective
       # @return [self]
       def add(detectors_configuration)
         detectors_configuration.each do |name, configuration|
-          detector = key_to_smell_detector(name)
-          self[detector] = (self[detector] || {}).merge configuration
+          if configuration
```
mvz, line 24: Empty configs should be filtered out earlier on. In fact, if you move the handling of the error, Reek will not run with nil configurations so this condition will no longer be needed.

mvz (commented):
This is very useful but the error handling should be more in line with the rest of Reek.

Once that's done it should be easy to have some tests for this. This will make the specs pass again.

mvz:
An alternative solution would be to move to `dry-schema` immediately.

troessner:
Agree with all of @mvz comments so far :)

fbuys:
Making this a WIP PR, waiting on https://github.com/troessner/reek/pull/1749

PR #1745, opened 2023-10-16
Fix editor highlight

JuanVqz:
Hey, I've noticed my editor is not rendering the code blocks as expected because the first letter is upper case in the code blocks.
I wonder if that could be the case for somebody else.

Peek 2023-10-16 16-45

- [x] Update Ruby code blocks
- [x] Update Yaml code blocks
- [x] Update Bash code block


JuanVqz:
Just noticed it has broken some tests if you are interested in the PR I can fix them otherwise, you can just close it. thanks!

JuanVqz:
I'm really glad to have touched those code blocks, I didn't know we had a test that uses our `.md` files as specs 😃. 

https://github.com/JuanVqz/reek/blob/fe0ccaf1718c8b22f7749d09ebed1264f3b7cbca/spec/quality/documentation_spec.rb#L4

certainly, I don't need to explain to you what is happening there because you wrote it but in general, changing the code block from `Ruby` to `ruby` makes them executable code and if the block doesn't find the `# =>` it will break because doesn't have anything to assert.

Example 1:
Example. Subclassed-From-Core-Class.md
```
(rdbg) code
"List = Class.new(Array)\n\nl = List.new\nl << 1\nl << 2\nputs l.reverse.class # => Array"
(ruby)           spec_code = code.gsub(/(^.+) # ?=> (.+$)/, 'assert_equal(\2, \1)')
"List = Class.new(Array)\n\nl = List.new\nl << 1\nl << 2\nassert_equal(Array, puts l.reverse.class)"
(ruby) spec_code
"List = Class.new(Array)\n\nl = List.new\nl << 1\nl << 2\nassert_equal(Array, puts l.reverse.class)"
```

troessner:
Merged, thanks @JuanVqz !

troessner:
Errrrr, can you please squash it to one commit @JuanVqz ? Then I'm happy to merge it.

JuanVqz:
@troessner and @mvz squashed, thanks!

Issue #1505, opened 2020-01-10
False positive for MonduleInitialize


jimeh:
I've come across two instances of a false positive ModuleInitialize warnings when using anonymous class definitions.

First the RSpec related one:

```ruby
module Alfa
  RSpec.describe Bravo do
    let(:klass) do
      Class.new do
        def initialize; end
      end
    end
  end
end
```

And secondly a DSL-based one:

```ruby
module Alfa
  bravo Class.new do
    def initialize; end
  end
end
```

The DSL variant is probably not very likely to be done in a real-world scenario, but technically the `initialize` method is not defined on the module that Reek warns about. The RSpec variant is something I've come across in the real world however.

If the anonymous class is assigned to a constant it does not fail however, which makes sense cause it's no longer an anonymous class:

```ruby
module Alfa
  Bravo = Class.new do
    def initialize; end
  end
end
```

A slightly different variant of this was reported in #1137 and fixed in #1287, but it specifically requires the anonymous class definition to be within a method definition on the module. Hence both of these are not reported:

```ruby
module Alfa
  def self.bravo
    Class.new do
      def initialize; end
    end
  end
end
```

```ruby
module Alfa
  def bravo
    Class.new do
      def initialize; end
    end
  end
end
```

Jwashton:
In a very similar case, I also get a false positive for this rule when defining a [Data](https://docs.ruby-lang.org/en/3.2/Data.html) class with an initializer.

```ruby
module Example
  Thing = Data.define(:value) do
    def initialize(value: 'some default')
      super(value:)
    end
  end
end
```

PR #1746, opened 2023-10-18
Fix Module with initialize method

JuanVqz:
Fixes https://github.com/troessner/reek/issues/1505

### DIFF
--- a/lib/reek/smell_detectors/module_initialize.rb
+++ b/lib/reek/smell_detectors/module_initialize.rb
@@ -21,14 +21,16 @@ module Reek
       # @return [Array<SmellWarning>]
       #
       def sniff
-        context.defined_instance_methods.each do |node|
-          if node.name == :initialize
-            return smell_warning(
-              lines:   [source_line],
-              message: 'has initialize method')
-          end
-        end
-        []
+        return [] if does_not_have_initializer_method?
+
+        smell_warning(lines: [source_line], message: 'has initialize method')
+      end
+
+      private
+
+      # :reek:FeatureEnvy
+      def does_not_have_initializer_method?
+        context.exp.direct_children.none? { |child| child.type == :def && child.name == :initialize }
       end
     end
   end
--- a/spec/reek/smell_detectors/module_initialize_spec.rb
+++ b/spec/reek/smell_detectors/module_initialize_spec.rb
@@ -86,4 +86,44 @@ RSpec.describe Reek::SmellDetectors::ModuleInitialize do
 
     expect(src).not_to reek_of(:ModuleInitialize)
   end
+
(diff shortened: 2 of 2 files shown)


lib/reek/smell_detectors/module_initialize.rb
```diff
@@ -21,14 +21,18 @@ def self.contexts # :nodoc:
...
       def sniff
-        context.defined_instance_methods.each do |node|
-          if node.name == :initialize
-            return smell_warning(
-              lines:   [source_line],
-              message: 'has initialize method')
-          end
-        end
-        []
+        return [] if does_not_have_initializer_method?
```
JuanVqz, line 24: @mvz @troessner I'd like to know what you think about the naming. thanks!

troessner: I am fine with the naming :)

spec/reek/smell_detectors/module_initialize_spec.rb
```diff
@@ -86,4 +86,44 @@ def initialize; end
...
+            end
+          end
+        end
+      end
+    RUBY
+
+    expect(src).not_to reek_of(:ModuleInitialize)
+  end
+
+  it 'reports nothing for a DSL-based' do
```
troessner, line 106: Seems like that description is incomplete?

troessner (changes requested):

troessner:
1 small comment from my side, other than that, this looks great. Wdyt @mvz ?

spec/reek/smell_detectors/module_initialize_spec.rb
JuanVqz: What sounds correct to you?

```ruby
  it 'reports nothing for a DSL-based method' do
```
```ruby
  it 'reports nothing for a DSL-based method initializing a class' do
```

lib/reek/smell_detectors/module_initialize.rb
```diff
@@ -21,14 +21,16 @@ def self.contexts # :nodoc:
       # @return [Array<SmellWarning>]
       #
       def sniff
-        context.defined_instance_methods.each do |node|
```
mvz, line 24: Apparently, the cause of #1505 is that `ModuleContext#defined_instance_methods` includes methods from dynamically defined nested classes. Instead of dropping down to bare sexp manipulation, I think this method should be adjusted instead.

mvz (changes requested):

lib/reek/smell_detectors/module_initialize.rb
JuanVqz: Do you have any clue where can I start?

PR #1747, opened 2023-10-23
Add `Tempfile.create` to ignored iterators list

### LINKED ISSUE #1490: NestedIterator should ignore _currently open_ like methods (like CSV and tempfile).
CSV (from standard library) uses a lot of methods that yield a `csv` file (that is opened and closed at the end of the block), `Tempfile.create` do the same:

Sample code:

~~~ ruby
CSV.generate headers: ['metric', 'value'], write_headers: true do |csv|
  metric_types.map do |id, value_type|
    csv << [id, SAMPLE_VALUE[value_type]]
  end
end

Tempfile.create('foo', '/home/temp') do |f|
   array.each do |element|
     f.puts element
  end
end
~~~

I'm not sure if I can list all the method names, but, I would use `open`, `create` and `generate`.

On CSV, there is also the `new` method that receives a block, but I'm not sure if it can be ignored (I can't think on an example where it is used as an iterator, but I'm not sure).

Currently `tap` is the only ignored method, what you thing about it?

mateusdeap:
Fixes #1490

I've added `Tempfile.create` to the list of ignored iterators and added a spec for it.

I'm submitting the PR to also get some feedback on the implementation. The specs all pass but it fails on the manual method dispatch code smell.

I tried searching for alternative ways to implement this without using `respond_to?` but didn't really find anything satisfactory, and if I don't test for that beforehand, many specs fail on exceptions because the `exp` object lacks the `receiver` method. Feedback on how to fix this would be most welcome!

Otherwise, I hope this is useful!

### DIFF
--- a/lib/reek/smell_detectors/nested_iterators.rb
+++ b/lib/reek/smell_detectors/nested_iterators.rb
@@ -26,7 +26,7 @@ module Reek
       # The name of the config field that sets the names of any
       # methods for which nesting should not be considered
       IGNORE_ITERATORS_KEY = 'ignore_iterators'
-      DEFAULT_IGNORE_ITERATORS = ['tap'].freeze
+      DEFAULT_IGNORE_ITERATORS = ['tap', 'Tempfile.create'].freeze
 
       def self.default_config
         super.merge(
@@ -128,8 +128,9 @@ module Reek
 
       # @quality :reek:FeatureEnvy
       def ignored_iterator?(exp)
-        ignore_iterators.any? { |pattern| /#{pattern}/ =~ exp.call.name } ||
-          exp.without_block_arguments?
+        ignore_iterators.any? do |pattern|
+          /#{pattern}/ =~ exp.call.format_to_ruby
+        end || exp.without_block_arguments?
       end
     end
   end
--- a/docs/defaults.reek.yml
+++ b/docs/defaults.reek.yml
@@ -57,6 +57,7 @@ detectors:
     max_allowed_nesting: 1
     ignore_iterators:
     - tap
+    - Tempfile.create
   NilCheck:
     enabled: true
     exclude: []
--- a/spec/reek/smell_detectors/nested_iterators_spec.rb
+++ b/spec/reek/smell_detectors/nested_iterators_spec.rb
@@ -66,6 +66,18 @@ RSpec.describe Reek::SmellDetectors::NestedIterators do
     expect(src).not_to reek_of(:NestedIterators)
   end
 
+  it 'does not report nested iterators for Tempfile#create' do
+    src = <<-RUBY
+      def alfa(bravo)
+        Tempfile.create('bravo', '.') do |charlie|
+          charlie.each { |delta| delta }
+        end
+      end
+    RUBY
+
+    expect(src).not_to reek_of(:NestedIterators)
+  end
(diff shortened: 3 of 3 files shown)


troessner (approved):

troessner:
Looks good to me! @mvz ?

mvz:
I'll have a look this evening.

mvz:
> For some reason, ManualDispatch complains that "Reek::SmellDetectors::NestedIterators#ignored_iterator? manually dispatches method call". I don't see that, so maybe that's a bug in ManualDispatch?

Ah, that detector simply finds calls to `#respond_to?`, so I guess that makes sense.

lib/reek/smell_detectors/nested_iterators.rb
```diff
@@ -128,8 +128,15 @@ def max_allowed_nesting
...
-        ignore_iterators.any? { |pattern| /#{pattern}/ =~ exp.call.name } ||
-          exp.without_block_arguments?
+        ignore_iterators.any? do |pattern|
+          call = exp.call
+          called_name = call.name
+          /#{pattern}/ =~ if call.respond_to? :receiver
+                            "#{call.receiver&.name}.#{called_name}"
+                          else
+                            called_name
+                          end
```
mvz, line 138: The following seems to work:

```ruby
/#{pattern}/ =~ call.format_to_ruby
```

mateusdeap: Interesting. I never heard of that method. Will edit the PR

mvz: It's a method defined on AST nodes by Reek itself.

mateusdeap: Applied changes

mateusdeap:
> @troessner if I read https://github.com/troessner/reek/issues/1490#issuecomment-525278433 correctly, you wanted to allow the option of ignoring Tempfile.create but not have that method be added to the default list of ignored methods?

@mvz As for this, now that I'm rereading it, I'm not sure. I interpreted it at the time as being "A PR that whitelists Tempfile.create". But if it's an option, than I suppose it would be something like a `whitelist` section in the reek configuration?

mvz:
>  But if it's an option, than I suppose it would be something like a whitelist section in the reek configuration?

Yes, it would be the `ignore_iterators` section just like here: https://github.com/troessner/reek/pull/1747/files#diff-473a12b36b25df9c36979c53cb67f2dce02afc614dca0c971b96154a7474d483R60, except in the user's own `.reek.yml` file.

mateusdeap:
I see. In that case, I'm not sure, honestly.

Now that I think about it, maybe I just leave the addition of the `Tempfile.create` item in the default reek.yml and that would be good enough?

Because I still get the impression the idea was that `Tempfile.create` (and some other methods, I now notice) are getting flagged with this smell even though they actually aren't cases of nested iterators. Basically all situations where the method yields a file that is open and closed within the loop.

Now I'm not so sure about the PR anymore lol

troessner:
Puhh, that's 4 years ago and I haven't really any context anymore regarding this. Feel free to ignore my comment ;)

mvz (approved):
Alright then, let's merge this!

PR #1750, opened 2023-11-12
Add downloads badge.

troessner:
Unfortunately [this nice little service](https://github.com/bogdanRada/ruby-gem-downloads-badge) is now defunct, so let's at least add a static badge. And yeah, it's a shameless plug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


mvz (approved):
Sure, let's go for it!

PR #1766, opened 2024-01-26
Update Class-Variable.md

BrianPurgert:
updated links, the article link is gone also added a Stack overflow explanation.


troessner:
Thanks for this!

PR #1768, opened 2024-02-03
Fix supported Ruby versions in the readme

jbockler:
I just stumbled across the Gem, but was surprised that Ruby 3.2 and 3.3 were not among the supported versions. I then found out from the changelog that they are.
Let's update the readme so that no one else is put off. 🙂


mvz (approved):

troessner:
Thanks @jbockler!

Issue #1627, opened 2021-12-16
Errors when building YARD index


Ji-You:
Could anyone please help me troubleshoot this issue?
For context, I am trying to run `yard gems` in my project directory for solargraph. (https://solargraph.org/guides/yard)
However, when it tries to generate documentation for `reek` gem, it outputs following error:
```
[info]: Building yardoc index for gem: reek-6.0.6
[error]: The file `./docs/yard_plugin.rb' could not be loaded:
cannot load such file -- ./docs/yard_plugin.rb
```
Your help would be most appreciated.

troessner:
What would be the easiest way for us to reproduce this?

Ji-You:
1. Enable auto-generation of docs `yard config --gem-install-yri`
2. Install reek `gem install reek`

Expected:
- Documentation is auto-generated for `reek` gem

Actual:
- It errors with following message:
```
$ gem install reek
Fetching: reek-6.0.6.gem (100%)
Successfully installed reek-6.0.6
Building YARD (yri) index for reek-6.0.6...
[error]: The file `./docs/yard_plugin.rb' could not be loaded:
cannot load such file -- ./docs/yard_plugin.rb
```

anicholson-sq:
I had this issue too – I did this to workaround it:

```shell
$ pushd $(bundle show reek)
$ mkdir -p docs
$ touch docs/yard_plugin.rb
$ popd
```

and then yard was able to process `reek` with no issues…

Ji-You:
@anicholson-sq 's workaround solved my problem. Thank you. I'm closing this issue now.

PR #1784, opened 2024-07-02
Fix Reek installation issue when Yard is installed

typhoon2099:
As per #1627, Yard can't process Reek because `docs/yarn_plugin.rb` is missing. It's not included in the Gem, so this PR adds it to correct this error.

This was reported upstream in
https://github.com/lsegal/yard/issues/1462, but it seems that it either wasn't, or people were happy enough with the workaround (touching the missing file in the Gem's installed location). The workaround doesn't work when using `asdf`/`mise` with [default
Gems](https://mise.jdx.dev/lang/ruby.html#default-gems). If Yard's automatic documentation is enabled (`yard config --gem-install-yri`) then the Ruby build fails.

### DIFF
--- a/reek.gemspec
+++ b/reek.gemspec
@@ -17,6 +17,7 @@ Gem::Specification.new do |spec|
   spec.files = Dir.chdir(File.expand_path(__dir__)) do
     `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features|samples|docs|tasks)/}) }
   end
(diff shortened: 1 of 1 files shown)


mvz:
Thanks @typhoon2099! Pity that the reporter closed #1627 since touching random files in gem installations is hardly a fix :disappointed:.

mvz:
I can confirm that this fixes the problem.

mvz:
Thanks, @typhoon2099!

PR #1793, opened 2024-10-28
Add `--format github` option

r7kamura:
How about adding `reek --format github` CLI option, like `rubocop --format github` or `brakeman --format github`?

This would make it very convenient for users in GitHub Actions, as they would be able to see violations on pull request diff page and GitHub Checks page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -556,13 +556,14 @@ e.g., `codeclimate analyze -e duplication`
 
 ## Output formats
 
-Reek supports 5 output formats:
+Reek supports 6 output formats:
 
 * plain text (default)
-* HTML (`--format html`)
-* YAML (`--format yaml`, see also [YAML Reports](docs/YAML-Reports.md))
-* JSON (`--format json`)
-* XML  (`--format xml`)
+* HTML   (`--format html`)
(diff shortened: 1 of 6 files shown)


r7kamura:
In GitHub Actions, when a message of the following form is written to standard output, GitHub will recognize it as a single warning annotation.

```
::warning file={name},line={line},endLine={endLine},title={title}::{message}
```

Detailed documentation can be found at:

- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions

The presence of this annotation will automatically annotate the diff page of the associated pull request; in the case of reek, this means that when the offending code is added, the details of the violation will be annotated there.

Here is an example of this in action:

- https://github.com/r7kamura/reek_github_format_example/pull/1/files
- https://github.com/r7kamura/reek_github_format_example/actions/runs/11551416639

Screenshots:

image

image

troessner:
Pretty cool, didn't know about this as well, PR looks good to me! LGTM @mvz ?

mvz:
:shipit:

PR #1803, opened 2025-02-21
Remove unused documention quality specs

mvz:
In practice, no assertions were run in this spec file: Code blocks were either filtered out because their language was not recognized as Ruby, or the Ruby code didn't contain the required marker.

With some work, three code blocks could be tested in the Subclassed-From-Core-Class documentation file, but I think this is more trouble than it's worth.

With this change, we can now downcase all language markers, and remove the kramdown dependency.


troessner (approved):

troessner:
Sorry, missed the review on this - lgtm?

PR #1840, opened 2026-08-14
Upgrade to simplecov 1.x

mvz:
In addition to bumping the dependency, this updates the simplecov setup to use simplecov's new configuration methods in a SimpleCov.configure block, and puts SimpleCov.start in the spec helper, which is the recommended and supported location.

It also places all configuration together in the configure block, dropping the custom SimpleCov.at_exit.

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -1,19 +1,14 @@
 # frozen_string_literal: true
 
-SimpleCov.start do
-  track_files 'lib/**/*.rb'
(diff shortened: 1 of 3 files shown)


mvz:
SimpleCov 1.x requres Ruby 3.2, so we first need to bump the minimum Ruby version.

mvz:
Waiting for #1842.

PR #1841, opened 2026-08-14
Test with JRuby 10.0 and 10.1 in CI

mvz:
This also sorts the JRuby versions so each comes right after the CRuby version whose compatibility it targets.


mvz:
There seem to be some performance issues here.

PR #1842, opened 2026-08-16
Support Ruby 3.2 and up

mvz:
- Require Ruby 3.2 or higher in the gemspec
- Bump RuboCop target ruby version and autocorrect new offenses
- Drop Ruby 3.1 and JRuby 9.4 from the CI configuration


mvz:
I originally wanted to first include JRuby 10.0 and 10.1 in the build matrix, but they're even slower than 9.4 (see #1841), so I'm dropping that for now.

PR #1844, opened 2026-08-28
Fix DuplicateMethodCall false positives

### LINKED ISSUE #1813: Two ways of false positives for DuplicateMethodCall
Hello reekers and thanks for the effort to support a world with better Ruby code.

In my opinion I found two kinds of false positives for the DuplicateMethodCall smell. Consider the following method - which also contains nested iterators and could be split such that the duplicate method calls are not found, but that is not the point. Reek reports duplicate method calls for `att[:file_name]`.

```ruby
def dedup_attachment_names
  attachments.group_by { |att| att[:file_name] }.each_value do |atts|
    atts.each_with_index do |att, idx|
      next if idx.zero?

      att[:file_name] += "_#{idx + 1}"
    end
  end
end
```
First, renaming one of the `att` block arguments makes the duplicate method call going away. My point is, it is a mere syntactical duplication but not a semantic one. It would not be possible to hold the lookup in a separate variable and have it reused in both places.

This is the weaker of my false positives and I'd say as tools like reek have to make tradeoffs its better to get a hint here which makes me reconsider what I wrote and if I'm sticking with it, use a magic comment to disable the finding. Nevertheless I thought, it might be worth to bring that to your attention.

The second false positive, for which I think this should not be flagged, is, that its not actually a duplicate method call. The first is a lookup. The second is a write using `+=`. I can see the syntactical reason for that, but semantically, to my knowledge, there is no way to avoid that. I see the need for duplicate method calls being flagged with method chaining - `foo.bar.baz` and `foo.bar.baz += 42` should flag `foo.bar` as duplicate - but I think `bar.baz` and `bar.baz += 42`  should not be considered duplicate method calls.

oiahoon:
## Summary

- distinguish identical-looking calls that refer to different block parameter bindings
- keep ordinary calls separate from compound-assignment targets while still reporting duplicated receiver prefixes
- add focused regression coverage and synchronize the sample report

## Verification

- `bundle exec rspec spec/reek/smell_detectors/duplicate_method_call_spec.rb`
- `bundle exec cucumber features/samples.feature`
- `bundle exec rake`

Closes #1813

### DIFF
--- a/features/samples.feature
+++ b/features/samples.feature
@@ -63,7 +63,7 @@ Feature: Basic smell detection
       UncommunicativeVariableName: Inline::C#module_name has the variable name 'x'
       UncommunicativeVariableName: Inline::C#parse_signature has the variable name 'x'
       UtilityFunction: Inline::C#strip_comments doesn't depend on instance state (maybe move it to another class?)
-    optparse.rb -- 126 warnings:
+    optparse.rb -- 122 warnings:
       Attribute: OptionParser#banner is a writable attribute
       Attribute: OptionParser#default_argv is a writable attribute
       Attribute: OptionParser#program_name is a writable attribute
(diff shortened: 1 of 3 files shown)


mvz:
Thanks @oiahoon. I'll take a look at the code in detail this weekend.

